PR #47, opened 2024-02-16
Print error if `ping_test` fails to run

dmitry-markin:
`ping_test` fails with older glibc. This PR is not really a fix, just info for a user on the root cause of the error (otherwise the test fails with error about socket connection failure). May be we should compile `ping_test` from sources?

### DIFF
--- a/tests/conformance/golang/tcp_ping_identify/mod.rs
+++ b/tests/conformance/golang/tcp_ping_identify/mod.rs
@@ -39,8 +39,12 @@ async fn go_libp2p_dials() {
     let mut file_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     file_path.push("tests/conformance/golang/tcp_ping_identify/ping_test");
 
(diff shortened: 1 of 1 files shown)


altonen (approved):
LGTM but we could also just disable that test, it's not bringing much value in its current form

At some point a better conformance test harness should be implemented for Golang so that it would be easier to write those tests

But let's go with this for now

dmitry-markin:
Another option would be to provide statically linked executable. But, generally, I don't think it's a good idea to provide binaries with the repo.

altonen:
I agree, it was a dumb idea and I can't even remember why litep2p is shipping a binary. That will have to be fixed

dmitry-markin:
May be it's not too late to rewrite the history and exclude the blob.

altonen:
I'd lean towards just removing the entire Golang conformance test code for now and creating an issue to bring it back at some point and do it well

PR #49, opened 2024-02-19
protocol/request-response: Fix memory leak

altonen:
Cancellation handles did not get removed after the request was handled successfully and would instead stay in the cancellation handle hashmap until the node was shut down.

### DIFF
--- a/src/protocol/request_response/mod.rs
+++ b/src/protocol/request_response/mod.rs
@@ -896,6 +896,8 @@ impl RequestResponseProtocol {
                             "failed to handle substream event",
                         );
                     }
+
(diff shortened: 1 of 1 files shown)


dmitry-markin (approved):

dmitry-markin:
Was this one responsible for 8GB memory usage by litep2p?

altonen:
Sadly not, this was leaking about 8 bytes of memory per outbound request but I found this leak while I was hunting for the bigger leak. That memory is still hiding somewhere

dmitry-markin:
> But decided to make PR instead of pushing directly master because there's now two of us working on this repository

Yes, thanks. That's how the expertise on the codebase is built.

PR #50, opened 2024-03-01
Refactor example code

altonen:
Just a small clean up

### DIFF
--- a/examples/echo_notification.rs
+++ b/examples/echo_notification.rs
@@ -20,6 +20,8 @@
 
 //! This example demonstrates a simple echo server using the notification protocol
 //! in which client connects to server and sends a message to server every 3 seconds
+//!
+//! Run: `cargo run --example echo_notification`
 
 use litep2p::{
     config::ConfigBuilder,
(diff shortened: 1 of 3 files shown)


examples/echo_notification.rs
```diff
@@ -38,19 +40,22 @@ use std::time::Duration;
...
-                NotificationEvent::NotificationStreamOpened { peer, .. } => break peer,
+                NotificationEvent::NotificationStreamOpened { .. } => break,
                 _ => {},
             }
         }
-    };
+    }
 
     // after the substream is open, send notification to server and print the response to stdout
     loop {
```
dmitry-markin, line 61: Could you explain what `.freeze()` is doing with `notification` in `server_event_loop()` below?

examples/echo_notification.rs
```diff
@@ -38,19 +40,22 @@ use std::time::Duration;
...
         tokio::select! {
             _ = litep2p.next_event() => {}
             event = handle.next() => match event.unwrap() {
-                NotificationEvent::NotificationStreamOpened { peer, .. } => break peer,
+                NotificationEvent::NotificationStreamOpened { .. } => break,
                 _ => {},
             }
         }
-    };
+    }
```
dmitry-markin, line 58: Just to make sure, that we are using `vec![1, 3, 3, 7]` both for handshake and for actual message doesn't mean that we must send the handshake manually?

dmitry-markin (approved):

examples/echo_notification.rs
altonen: Currently there is a bit of mismatch in the API whereby it returns `BytesMut` but accepts `Vec<u8>`. Internally litep2p uses `BytesMut` for almost everything and it'd be nice if the APIs were consistent in that regard but it hasn't been a high priority.

examples/echo_notification.rs
altonen: Yeah, it's just lack of imagination that both the notification that it sends and the handshake it uses are the same. But handshake is sent automatically by `NotificationProtocol`

PR #51, opened 2024-03-01
Refactor WebRTC code

altonen:
This PR refactors the WebRTC code into something that actually works. The `WebRtcConnection` code is split into `OpeningWebRtcConnection` and `WebRtcConnection`. The former only deals with the Noise handshake whereas the latter implements a typical connection event loop, dealing with inbound/outbound substreams and data exchange. The PR also implements a new substream type for WebRTC which is implemented using `tokio::sync::mpsc::{Sender, Receiver}` since `str0m` doesn't expose a substream type. The `multistream-select` code used by WebRTC is refactored and its test coverage is increased.

This PR also updates `str0m` to the latest version but during testing it was discovered that there are issues with `ChannelId`/SCTP stream id reuse and those have to be reported to upstream and fixed. They're are not blockers for merging this PR though because the fixes should be contained to `str0m` and for litep2p it would only be a version bump if the fixes are accepted.

The WebRTC implementation (with pending fixes to `str0m`) has been tested with smoldot and the code works. IPFS Ping, Kademlia and Identify all work, opening and negotiating notification protocols work and block requests work. There are, however, two issues that I was able to find:

1) The first "notification" smoldot reads from the substream is not actually a notification but the substream handshake which results in "protocol-error":
```
[network] protocol-error; peer_id=12D3KooWNGCW63xu9tGh1WxWEazHs6UgMdTjrLSos89NrVSCh9bz, error=BadBlockAnnounce(DecodeBlockAnnounceError(Verify))
[network] protocol-error; peer_id=12D3KooWNGCW63xu9tGh1WxWEazHs6UgMdTjrLSos89NrVSCh9bz, error=BadGrandpaNotification(DecodeGrandpaNotificationError(Verify))
```

It looks like smoldot is still able to receive block announcements and GRANDPA notifications so I'm not entirely what the issue here is.

2) There is another issues with justifications but I don't see how it could be related to WebRTC
```
[sync-service-westend2] finality-proof-verify-error; error=FinalityVerify(UnknownTargetBlock { block_number: 19772928, block_hash: [22, 163, 175, 94, 114, 75, 142, 174, 83, 146, 11, 251, 184, 119, 212, 76, 34, 85, 144, 47, 83, 251, 83, 86, 152, 141, 204, 104, 253, 236, 21, 48] }), sender=12D3KooWNGCW63xu9tGh1WxWEazHs6UgMdTjrLSos89NrVSCh9bz
[sync-service-westend2] Error while verifying justification: Justification targets a block (#19772928) that isn't in the chain.
```

This happens on the first try, gossip is closed, then it's opened again and it's able to start warp sync.


altonen:
@dmitry-markin 

Since Versi was busy with Kusama tests, I spoke with Andre and we decided that I should look into WebRTC in the meantime, even though it was technically your task. This is a big PR so I'm not expecting a comprehensive review but it'll (kind of) make it work so it should be relatively straightforward to fix any further issues.

dmitry-markin:
> This PR also updates `str0m` to the latest version but during testing it was discovered that there are issues with `ChannelId`/SCTP stream id reuse and those have to be reported to upstream and fixed. They're are not blockers for merging this PR though because the fixes should be contained to `str0m` and for litep2p it would only be a version bump if the fixes are accepted.

What this means for us? I.e., is this causing any issues for substrate/lightp2p, or we can happily live without these issues fixed in `str0m`?

altonen:
It means that outbound substreams don't work very well which is a blocker insofar as releasing WebRTC goes. Code needed to fix those issues are quite trivial so I'll open a pull request to `str0m` tomorrow/next week and hopefully they'll accept it.

dmitry-markin (approved):
Definitely not a thorough review as this is all new stuff for me, and I'm far from understanding all the details. Having said that, looks good.

src/transport/webrtc/config.rs
```diff
@@ -27,4 +27,20 @@ use multiaddr::Multiaddr;
...
+    /// Connection datagram buffer size.
+    ///
+    /// How many datagrams can the buffer between `WebRtcTransport` and a connection handler hold.
+    pub datagram_buffer_size: usize,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            listen_addresses: vec!["/ip4/127.0.0.1/udp/8888/webrtc-direct"
```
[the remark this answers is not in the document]
altonen: Oversight, should be `/ip4/0.0.0.0/udp/0/webrtc-direct`

src/multistream_select/dialer_select.rs
```diff
@@ -269,29 +270,22 @@ impl DialerState {
...
-
-        // TODO: add fallback names
-
-        header.append(&mut proto_bytes.into());
+        let message = encode_multistream_message(
+            std::iter::once(protocol.clone())
+                .chain(fallback_names.clone())
+                .filter_map(|protocol| Protocol::try_from(protocol.as_ref()).ok())
+                .map(|protocol| Message::Protocol(protocol)),
+        )?
```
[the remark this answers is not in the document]
altonen: I'll have to check this but this works with smoldot, both for listener and dialer.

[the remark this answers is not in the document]
altonen: You're right. It looks like an incorrect way of handling the handshake and litep2p [would reject](https://github.com/altonen/litep2p/blob/d3b1be62904a98b19d03c0bf213658f9f441cf71/src/multistream_select/listener_select.rs#L259) this message as well but smoldot both sends and accepts `Message::Protocols`. I'm kind of tempted to keep it as is since it works but I'll check if smoldot would work with `Message::Protocol`

src/transport/webrtc/config.rs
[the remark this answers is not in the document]
altonen: This is the default implementation for `Config`. If the default is no good, you can override, like we do in [Polkadot SDK](https://github.com/paritytech/polkadot-sdk/blob/98379bd4914cc942bd9a5c8ac9717e890b5faf4e/substrate/client/network/src/litep2p/mod.rs#L355-L365)

dmitry-markin:
@lexnv https://github.com/paritytech/litep2p/pull/74 made resolving conflicts with master trivial, and this is now ready for review.

tests/webrtc.rs
```diff
@@ -0,0 +1,80 @@
...
+#[tokio::test]
+#[ignore]
+async fn webrtc_test() {
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init();
+
+    let (ping_config, mut ping_event_stream) = ping::Config::default();
+    let (notif_config, mut notif_event_stream) = ConfigBuilder::new(ProtocolName::from(
+        // "/c0c89622f83f6f3c6b94bc307fec1652a3aa58ac88a564c34706633f44cbb3d1/block-announces/1",
```
lexnv, line 40: I think we can safely remove this

```suggestion
```

tests/webrtc.rs
```diff
@@ -0,0 +1,80 @@
...
+    tracing::info!("listen address: {address:?}");
+
+    loop {
+        tokio::select! {
+            event = litep2p.next_event() => {
+                tracing::debug!("litep2p event received: {event:?}");
+            }
+            event = ping_event_stream.next() => {
+                if std::matches!(event, None) {
+                    tracing::error!("ping event stream termintated");
```
lexnv, line 70:
```suggestion
                    tracing::error!("ping event stream terminated");
```

tests/webrtc.rs
```diff
@@ -0,0 +1,80 @@
...
+    ))
+    .with_max_size(5 * 1024 * 1024)
+    .with_handshake(vec![1, 2, 3, 4])
+    .with_auto_accept_inbound(true)
+    .build();
+
+    let config = Litep2pConfigBuilder::new()
+        .with_keypair(Keypair::generate())
+        .with_webrtc(Config {
+            listen_addresses: vec!["/ip4/192.168.1.170/udp/8888/webrtc-direct".parse().unwrap()],
```
lexnv, line 51: The block-announces protocol uses `e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e` which is the genesis of Westend, however I cannot connect to this addr.
Is there another way to test this?

tests/webrtc.rs
```diff
@@ -0,0 +1,80 @@
...
+    config::ConfigBuilder as Litep2pConfigBuilder,
+    crypto::ed25519::Keypair,
+    protocol::{libp2p::ping, notification::ConfigBuilder},
+    transport::webrtc::config::Config,
+    types::protocol::ProtocolName,
+    Litep2p,
+};
+
+#[tokio::test]
+#[ignore]
```
lexnv, line 32: I think the test is ignored because the IP address is not reachable anymore

src/transport/webrtc/opening.rs
```diff
@@ -0,0 +1,473 @@
...
+const LOG_TARGET: &str = "litep2p::webrtc::connection";
+
+/// Create Noise prologue.
+fn noise_prologue(local_fingerprint: Vec<u8>, remote_fingerprint: Vec<u8>) -> Vec<u8> {
+    const PREFIX: &[u8] = b"libp2p-webrtc-noise:";
+    let mut prologue =
+        Vec::with_capacity(PREFIX.len() + local_fingerprint.len() + remote_fingerprint.len());
+    prologue.extend_from_slice(PREFIX);
+    prologue.extend_from_slice(&remote_fingerprint);
+    prologue.extend_from_slice(&local_fingerprint);
```
lexnv, line 51: IIUC, this is <prefix><fingerprint A><fingerprint B>, depending on whatever this is the handshake responder or the initator.
(extracted from: https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md#connection-security)

Do we need to handle the case where we are the responder? Ie <prefix><local><remote> instead of <prefix><remote><local>

src/transport/webrtc/substream.rs
```diff
@@ -18,132 +18,447 @@
...
+        if let Some(flags) = message.flags {
+            if flags == Flag::Fin as i32 {
+                let _ = self.tx.send(Event::RecvClosed).await?;
+            }
+
+            if flags & 1 == Flag::StopSending as i32 {
+                *self.state.lock() = State::SendClosed;
+            }
+
+            if flags & 2 == Flag::ResetStream as i32 {
```
lexnv, line 128: dq: We don't care about `Flag::FinAck` because we are closing the connection either way? (https://github.com/libp2p/specs/blob/master/webrtc/README.md#shared-concepts)

tests/webrtc.rs
```diff
@@ -0,0 +1,80 @@
+// Copyright 2023 litep2p developers
```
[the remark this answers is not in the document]
dmitry-markin: We can gradually append this line to modified files. I think we should also keep the original copyright that basically refers to Aaro, as litep2p started as his side project.

tests/webrtc.rs
dmitry-markin: The listen address should be updated to 127.0.0.1. Or does the WebRTC spec impose the limitations of connecting to loopback addresses?

Then, I guess, for this test to succeed another Westend node with litep2p backend is needed configured to dial this listen address.

tests/webrtc.rs
dmitry-markin: It probably is intended to be run manually with a companion node.

src/transport/webrtc/opening.rs
dmitry-markin: My guess is it's based on the role of smoldot in the WebRTC noise handshake. It looks like we don't support litep2p to litep2p WebRTC connections.

tests/webrtc.rs
dmitry-markin: Taking into account the noise prologue comment, it looks like smoldot is needed as a second node for this test to succeed.

src/transport/webrtc/substream.rs
dmitry-markin: I wonder why we apply binary `&`, even though flags seem to be exclusive and not a bitmask. Also, this way `FinAck` is handled by conditions `flags & 1 == StopSending` and `flags & 2 == ResetStream`, so this is probably the reason we don't have a separate condition for `FinAck`...

dmitry-markin: I would merge as is and refactored this in a follow-up PR.

tests/webrtc.rs
dmitry-markin: May be just add the chain names and keep the commented-out genesis hash? the test is intended for manual invocation, so the protocol names are expected to be manually modified.

src/transport/webrtc/opening.rs
altonen: This PR only implemented WebRTC Direct (browser <-> server), not server <-> server.

tests/webrtc.rs
dmitry-markin: Thanks @altonen for the explanation!

tests/webrtc.rs
dmitry-markin: I have no idea what chain the commented-out genesis hash belongs to. It's neither polkadot/kusama/westend/rococo/wococo. Let's delete it.

tests/webrtc.rs
```diff
@@ -0,0 +1,79 @@
...
+#[tokio::test]
+#[ignore]
+async fn webrtc_test() {
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init();
+
+    let (ping_config, mut ping_event_stream) = ping::Config::default();
+    let (notif_config, mut notif_event_stream) = ConfigBuilder::new(ProtocolName::from(
+        "/e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e/block-announces/1",
```
dmitry-markin, line 40:
```suggestion
        // Westend block-announces protocol name.
        "/e143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e/block-announces/1",
```

PR #52, opened 2024-03-07
[kademlia] Use `ClosestBucketsIter` from libp2p to get closest peers

dmitry-markin:
Find peers in the routing table close to target using `ClosestBucketsIter` from libp2p.

Resolves https://github.com/altonen/litep2p/issues/38.

### DIFF
--- a/src/protocol/libp2p/kademlia/bucket.rs
+++ b/src/protocol/libp2p/kademlia/bucket.rs
@@ -101,12 +101,10 @@ impl KBucket {
 
     /// Get iterator over the k-bucket, sorting the k-bucket entries in increasing order
     /// by distance.
-    pub fn closest_iter<'a, K: Clone>(
-        &'a mut self,
-        target: Key<K>,
-    ) -> impl Iterator<Item = &'a KademliaPeer> {
-        self.nodes.sort_by(|a, b| target.distance(&a.key).cmp(&target.distance(&b.key)));
-        self.nodes.iter().filter(|peer| !peer.addresses.is_empty())
+    pub fn closest_iter<K: Clone>(&self, target: &Key<K>) -> impl Iterator<Item = KademliaPeer> {
+        let mut nodes = self.nodes.clone();
+        nodes.sort_by(|a, b| target.distance(&a.key).cmp(&target.distance(&b.key)));
+        nodes.into_iter().filter(|peer| !peer.addresses.is_empty())
     }
 }
 
@@ -129,7 +127,7 @@ mod tests {
             .collect::<Vec<_>>();
 
         let target = Key::from(PeerId::random());
-        let mut iter = bucket.closest_iter(target.clone());
+        let mut iter = bucket.closest_iter(&target);
         let mut prev = None;
 
         while let Some(node) = iter.next() {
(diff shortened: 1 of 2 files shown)


src/protocol/libp2p/kademlia/routing_table.rs
```diff
@@ -178,46 +178,106 @@ impl RoutingTable {
...
+    }
 
-            nodes.sort_by(|a, b| target.distance(&a.key).cmp(&target.distance(&b.key)));
+    fn next_out(&self, i: BucketIndex) -> Option<BucketIndex> {
+        (i.get() + 1..NUM_BUCKETS).find_map(|i| {
+            if !self.distance.0.bit(i) {
+                Some(BucketIndex(i))
+            } else {
+                None
+            }
```
altonen, line 248: `then_some()` for both of these?

src/protocol/libp2p/kademlia/bucket.rs
```diff
@@ -101,12 +101,10 @@ impl KBucket {
 
     /// Get iterator over the k-bucket, sorting the k-bucket entries in increasing order
     /// by distance.
-    pub fn closest_iter<'a, K: Clone>(
-        &'a mut self,
-        target: Key<K>,
-    ) -> impl Iterator<Item = &'a KademliaPeer> {
-        self.nodes.sort_by(|a, b| target.distance(&a.key).cmp(&target.distance(&b.key)));
-        self.nodes.iter().filter(|peer| !peer.addresses.is_empty())
+    pub fn closest_iter<K: Clone>(&self, target: &Key<K>) -> impl Iterator<Item = KademliaPeer> {
```
altonen, line 104: Is `K: Clone` still needed?

altonen (approved):
Looks reasonable. Probably wise to run a burn-in for this which I'll do alongside other Versi tests

src/protocol/libp2p/kademlia/bucket.rs
dmitry-markin: Yes, as this is the requirement of `Key`.

src/protocol/libp2p/kademlia/bucket.rs
```diff
@@ -101,12 +101,10 @@ impl KBucket {
...
+        let mut nodes = self.nodes.clone();
```
dmitry-markin, line 105: I'm not super happy about this `clone()` — a better way would be to only clone nodes that we finally yield from `RoutingTable::closest()`. But without this clone we'll need to manually iterate over buckets and peers within them, and the trick with nesting `ClosestBucketsIter` and `KBucket::closest_iter` and flattening them won't work due to borrowing of `self` in closures in `RoutingTable::closest()`.

src/protocol/libp2p/kademlia/routing_table.rs
dmitry-markin: Done

dmitry-markin:
> Looks reasonable. Probably wise to run a burn-in for this which I'll do alongside other Versi tests

Would you like to have this merged for burn-ins from master?

altonen:
> > Looks reasonable. Probably wise to run a burn-in for this which I'll do alongside other Versi tests
> 
> Would you like to have this merged for burn-ins from master?

Yes let's merge it and if it shows issues, we can revert it

src/protocol/libp2p/kademlia/bucket.rs
altonen: Maybe create an issue if it's worth revisiting at some point in the future? While I agree with you that it could be better, this PR is already a big improvement over the previous implementation so I'm fine with merging this as is

PR #54, opened 2024-03-07
Fix handling of pending actions

altonen:
Pending actions stored to `pending_dials` were handled incorrectly, causing the code to overwrite previous pending actions that would lead to queries getting stuck, and in `GET_VALUE`'s case, leaking noticeable amount of memory.

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -95,17 +95,16 @@ struct PeerContext {
 }
 
 impl PeerContext {
+    /// Create new [`PeerContext`].
     pub fn new() -> Self {
         Self {
             pending_actions: HashMap::new(),
         }
     }
 
-    /// Create new [`PeerContext`] with pending [`PeerAction`].
-    pub fn with_pending_action(substream_id: SubstreamId, action: PeerAction) -> Self {
-        Self {
-            pending_actions: HashMap::from_iter([(substream_id, action)]),
-        }
+    /// Add pending action for peer.
+    pub fn add_pending_action(&mut self, substream_id: SubstreamId, action: PeerAction) {
+        self.pending_actions.insert(substream_id, action);
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -188,11 +187,23 @@ impl Kademlia {
 
         match self.peers.entry(peer) {
             Entry::Vacant(entry) => {
-                // TODO: verify that peer limit is respected
-                match self.pending_dials.remove(&peer) {
-                    Some(action) => match self.service.open_substream(peer) {
+                if let KBucketEntry::Occupied(entry) = self.routing_table.entry(Key::from(peer)) {
+                    entry.connection = ConnectionType::Connected;
+                }
```
dmitry-markin, line 192: Are we not trying to insert the peer into the routing table on purpose? Doesn't make sense for outbound connections, but can't this be called when remote nodes discover us?

dmitry-markin (approved):

dmitry-markin:
Is it now _that one_ leak?

altonen:
Sadly not. With Jan's help I was able to locate the other two. One is in `notification::Connection` which needs a small rewrite and it looks like the other one is in yamux which I'm still trying to figure out.

src/protocol/libp2p/kademlia/mod.rs
altonen: I'm not sure I understand your comment. If the peer already exists in `peers`, we don't do anything because but if it doens't, we set its state to `Connected`. Am I missing something?

dmitry-markin: I just thought we can learn the addresses of connecting peers (if the other peer initiated the connection) and insert them into the routing table. I'm not sure if this is a good idea, would bring any value on top of discovering peers via random walks, and wouldn't introduce a DoS vector.

altonen: I think there is a certain "risk" in adding addresses reported via `TransportEvent::ConnectionEstablished` to routing table because we don't really know if those addresses are reachable and if they're not, then we advertise bad addresses to anyone who might get this peer in a `FIND_NODE` response.

dmitry-markin: I agree, that was not the best idea.

PR #55, opened 2024-03-10
Fix memory leaks

altonen:
The big memory leak was related to outbound backpressure. Notification connection handler would block on `Substream::send_framed()` and accumulated a lot of memory in the notification channel which apparently didn't get released appropriately, possibly due to connection handler not getting closed. The connection handler is rewritten in a way where it processes both inbound and outbound notifications without blocking on either one of them so if the node is trying to send a large notification/many notifications, it won't block the entire connection event loop and instead gradually writes the notification(s) to the substream. Size of the channel that's used to send notifications synchronously is adjusted to match the one in Polkadot SDK.

`NotificationHandle::send_sync_notification()` is modified to match the implementation currently found in Polkadot SDK whereby it closes the connection if the channel is clogged. I don't think this is a particularly good idea in the first place and once the protocols in Polkadot SDK have been written in such a way that they can deal with backpressure, e.g., by prioritizing certain notifications over others, this behavior can (and probably should) be removed.

Third change is the largest and that is to (temporarily) move yamux into litep2p. The flamegraphs indicated that yamux was consuming a lot of memory and while bumping it to `0.11.1`, where the patch released fixed an issue with GC, helped, it was still leaking memory. The follow-up for the GC was fixed in `0.12.1` but `0.12.0` introduced significant API breakage, probably requiring a rewrite of the TCP and WebSocket connection handlers, which I'm not interested in doing at this time. I backported the memory leak fix from `0.12.1` to `0.11.1` and transferred the source code of yamux to litep2p and it improved things to a point where litep2p doesn't leak memory anymore. Moving yamux into litep2p is not optimal but considering that `0.12.0` removed the APIs that litep2p is relying on, thus requiring non-trivial rewrites of the connection handlers, and that litep2p is leaking too much memory on `0.11.1`, it seems like the best approach for now. Yamux source code doesn't need a review and can be skipped.

Lastly, add timeouts for IPFS Identify, IPFS Ping, and request-response protocols. It may be possible for the read/write to block indefinitely which could cause the future to stay active which in turn would keep the connection open as each substream holds a permit, preventing the connection from getting closed. Now that each substream operation has a timeout, the connection won't be kept open erroneously, allowing resources to be released.

This is big PR but I've run a burn-in for these changes in Versi so it should be safe to merge this.

### DIFF
--- a/src/protocol/libp2p/ping/mod.rs
+++ b/src/protocol/libp2p/ping/mod.rs
@@ -124,15 +124,23 @@ impl Ping {
         tracing::trace!(target: LOG_TARGET, ?peer, "handle outbound substream");
 
         self.pending_outbound.push(Box::pin(async move {
-            // TODO: generate random payload and verify it
-            let _ = substream.send_framed(vec![0u8; 32].into()).await?;
-            let now = Instant::now();
-            let _ = substream.next().await.ok_or(Error::SubstreamError(
-                SubstreamError::ReadFailure(Some(substream_id)),
-            ))?;
-            let _ = substream.close().await;
+            let future = async move {
+                // TODO: generate random payload and verify it
+                let _ = substream.send_framed(vec![0u8; 32].into()).await?;
+                let now = Instant::now();
+                let _ = substream.next().await.ok_or(Error::SubstreamError(
+                    SubstreamError::ReadFailure(Some(substream_id)),
+                ))?;
+                let _ = substream.close().await;
 
-            Ok((peer, now.elapsed()))
+                Ok(now.elapsed())
+            };
+
+            match tokio::time::timeout(Duration::from_secs(10), future).await {
+                Err(_) => return Err(Error::Timeout),
+                Ok(Err(error)) => return Err(error),
+                Ok(Ok(elapsed)) => Ok((peer, elapsed)),
+            }
         }));
     }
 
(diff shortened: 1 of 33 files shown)


src/protocol/libp2p/ping/mod.rs
```diff
@@ -141,14 +149,22 @@ impl Ping {
...
-            let _ = substream.next().await.map(|_| ());
-
-            Ok(())
+            let future = async move {
+                let payload = substream
+                    .next()
+                    .await
+                    .ok_or(Error::SubstreamError(SubstreamError::ReadFailure(None)))??;
+                substream.send_framed(payload.freeze()).await?;
+                let _ = substream.next().await.map(|_| ());
```
dmitry-markin, line 158: What are we reading here?

src/protocol/notification/connection.rs
```diff
@@ -126,64 +136,137 @@ impl Connection {
...
+}
+
+/// Connection events.
+pub enum ConnectionEvent {
+    /// Close connection.
+    ///
+    /// If `NotificationProtocol` requested [`Connection`] to be closed, it doesn't need to be
+    /// notified. If, on the other hand, connection closes because it encountered an error or one
+    /// of the substreams was closed, `NotificationProtocol` must be informed so it can inform the
+    /// user. If the connection should be closed because [`Connection`] encountered an error
```
dmitry-markin, line 172:
```suggestion
    /// user.
```

dmitry-markin (approved):

src/protocol/libp2p/ping/mod.rs
altonen: Nothing, it's just some old code back when I was debugging the ping protocol and [it didn't work without it](https://i.makeagif.com/media/11-28-2016/XaxcSr.mp4). Probably can be removed.

PR #56, opened 2024-03-13
[ci] Add workflow

alvicsam:
cc https://github.com/paritytech/ci_cd/issues/960


alvicsam:
Sure, added. The jobs have the same CI image (`image: "paritytech/ci-unified:bullseye-1.75.0-2024-01-22-v20240222"`) in order to not repeat it several times we can add it to repo environment variables

dmitry-markin:
@alvicsam it seems a loopback interface is not enabled in the container. All the "real" networking tests fail with
```
called `Result::unwrap()` on an `Err` value: Os { code: 99, kind: AddrNotAvailable, message: "Cannot assign requested address" }
```
Can you have a look, please?

dmitry-markin:
> The jobs have the same CI image (`image: "paritytech/ci-unified:bullseye-1.75.0-2024-01-22-v20240222"`) in order to not repeat it several times we can add it to repo environment variables (but I need permissions for that)

That would be nice. What kind of permissions do you need? What would be the best way forward to keep the image in sync with polkadot-sdk repo tests (so that at least the rust toolchain is the same)?

alvicsam:
> Can you have a look, please?

Sure, I'll take a look

> What would be the best way forward to keep the image in sync with polkadot-sdk repo tests (so that at least the rust toolchain is the same)?

This is a really good question. The problem is actually bigger because there will be more and more repos that will depend on the same CI image that is used in polkadot-sdk repo. I will create an issue in our tracker with thoughts how we can achieve that. 

As for permissions I'll ask the security team to provide them.

alvicsam:
@dmitry-markin it seems I need your help with the tests
Tests in container fail with the error you sent (even if the container runs in the privileged mode). Without container it hangs on `test protocol::request_response::large_response_quic has been running for over 60 seconds`

dmitry-markin:
> @dmitry-markin it seems I need your help with the tests Tests in container fail with the error you sent (even if the container runs in the privileged mode). Without container it hangs on `test protocol::request_response::large_response_quic has been running for over 60 seconds`

My guess is that the issue can be due to loopback inteface not being present / being down. What was the output of `ip a` in the commit [`ci image var, check lo`](https://github.com/paritytech/litep2p/pull/56/commits/cb3dea2d518be5023e13fffc274f68988c4c0a1a)? If the `lo` interface was down, it should be brought up with `ip link set lo up` (probably appropriate privileges are required), and if it's not — there is something severely off with the networking in the container.

If `lo` is present and is up, we'll need to investigate further.

dmitry-markin:
> I could run a simple http server on the lo interface so it's available.

Thanks, I'll think how to investigate the problem.

lexnv:
I've created https://github.com/paritytech/litep2p/pull/63 to test out a few ideas.

The first commits validate that the IPv6 is not enabled, by using `127.0.0.1` to bind instead of `::1`.

After reading more about this, I think the fix should be present in: https://github.com/paritytech/litep2p/pull/63/commits/8c6f88e39297fb5f648dfcff5ae70bb0960f21b2, which adds `--sysctl net.ipv6.conf.all.disable_ipv6=0`. I don't believe offhand that we need to run the container in privileged mode.

I'll wait a bit to see if the testing is passing.

As always, thanks a lot @alvicsam for helping us here! 🙏

dmitry-markin:
> I've created #63 to test out a few ideas.

This have solved the issue with binding to loopback addresses! :tada: Thanks @lexnv!

Yet, we have some other failures:
1. `conformance::golang::tcp_ping_identify::go_libp2p_dials` — as I wrote somewhere this one should be ditched as it uses a pre-built go binary, and as per Aaro, doesn't provide much value
2. `protocol::request_response::large_response_*`
    - `large_response_quic` `FAILED`
    - `large_response_websocket` `FAILED`
    - `large_response_tcp` hangs
    All three works locally.

altonen:
Not sure what kind of machines the CI uses but is it possible that the request timeouts when it's trying to create the 15MiB response? I remember generating large number of random bytes was somewhat CPU intensive when I ran some tests locally so if the test would instead generate 15MiB vector of zeros *before* sending the request ([swapping the order of these lines](https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/tests/protocol/request_response.rs#L2586-L2590)), would that make it pass?

dmitry-markin:
Pushed to a wrong branch, continuing in https://github.com/paritytech/litep2p/pull/63.

dmitry-markin:
> Not sure what kind of machines the CI uses but is it possible that the request timeouts when it's trying to create the 15MiB response? I remember generating large number of random bytes was somewhat CPU intensive when I ran some tests locally so if the test would instead generate 15MiB vector of zeros _before_ sending the request ([swapping the order of these lines](https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/tests/protocol/request_response.rs#L2586-L2590)), would that make it pass?

It worked, and I also removed golang conformance tests, so now we should have all tests passing.

lexnv (approved):

lexnv:
I'm happy to merge this whenever you think is ready! 🙏

dmitry-markin (approved):

dmitry-markin:
> I'm happy to merge this whenever you think is ready! 🙏

CC @alvicsam

PR #57, opened 2024-03-13
clippy: Fix cargo clippy

lexnv:
This PR fixes:
- clippy lints
- rust doc missing links
- remove unused dependency


dmitry-markin (approved):
Everything seems good, but did you run the tests just to make sure everything is OK? Don't know, may be even check with a real node, because the changes are quite bulky.

Issue #59, opened 2024-03-18
multistream: Dialer accepts multiple /multistream/1.0.0 headers


lexnv:
The dialer of the multistream protocol can accept multiple times the `multistream` V1 header:
https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/src/multistream_select/dialer_select.rs#L195-L197

This may lead to peers keeping connections alive due to state mismatch. 

My understanding of the [libp2p spec](https://github.com/libp2p/specs/blob/master/connections/README.md#multistream-select) is that the multistream header should be send once, as the first message. Then, peers should negotiate the protocol names they want to serve.

The listener code is correctly awaiting the header once:
https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/src/multistream_select/listener_select.rs#L149
 

The issue is also present in the [rust-libp2p repo](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/misc/multistream-select/src/dialer_select.rs#L178-L181), which is where I suspect our code took inspiration from.

Let's return `ProtocolError::InvalidMessage` if the header is sent multiple times.

cc @paritytech/networking

Issue #60, opened 2024-03-18
multistream: Lazy negotiated protocol cannot be closed before receiving messages


lexnv:
The lazy negotiated protocol is constructed when `V1Lazy` is specified (notes: https://github.com/libp2p/rust-libp2p/pull/1212).

https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/src/multistream_select/dialer_select.rs#L157-L165

The lazy negotiated protocol then expects the header (/multistream/1.0.0) and a single protocol name from the socket, otherwise the negotiation fails:

https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/src/multistream_select/negotiated.rs#L182-L194

However, the `Negotiated` cannot be closed before the negotiation message is received from the other end:

https://github.com/paritytech/litep2p/blob/cc043b66f9c2c20349359150f522b83f260ec587/src/multistream_select/negotiated.rs#L321-L332

This is because the `poll` method will await the protocol negotiation.

This behavior is different from the [libp2p component](https://github.com/libp2p/rust-libp2p/blob/3644879956b6ab93b8d23553a33e8fdb838f576f/misc/multistream-select/src/negotiated.rs#L307-L312) that the litep2p replaced.

I'm wondering why the lazy protocol should ensure that the `/prococol/...` has been received before closing the connection?

cc @paritytech/networking

PR #61, opened 2024-03-18
multistream: Dialer deny multiple /multistream/1.0.0 headers

lexnv:
This PR ensures that the `/multistream/1.0.0` header can only be received once during a protocol negotiation.
This is fixed for both the V1 and V1Lazy variations.

While at it, have added a few tests to check:
- protocol negotiation
- protocol negotiation failure due to unsupported protocols
- lazy protocol negotiation awaits protocol message before closing (to be fixed by: https://github.com/paritytech/litep2p/issues/60)
- low level protocol listener that writes the expected messages
- low level protocol listener that submits multiple headers, both for V1 and V1 lazy

Closes: https://github.com/paritytech/litep2p/issues/59

cc @paritytech/networking


dmitry-markin:
@lexnv We have a pending PR not yet merged that fixes WebRTC: https://github.com/paritytech/litep2p/pull/51.

I'd double check that we don't have conflicting changes, probably merging that one first.

lexnv:
That makes sense! Thanks for the info 🙏 

I think would proly be good to have the CI up and running before we merge any other PRs (I've also put https://github.com/paritytech/litep2p/pull/57 on hold until we get more confidence in our testing)

After having a quick look over the WebRTC PR, it seems that the changes will not conflict:
- as the main change is done to `DialerState` and adding tests for `DialerState` (prob need to double check names)
- and it does not touch the negotiating part of the code

Yep, I think would be good to get that PR in first to ensure we don't break anything with these changes

src/multistream_select/dialer_select.rs
```diff
@@ -346,3 +410,358 @@ impl DialerState {
...
+    use crate::multistream_select::listener_select_proto;
+    use std::time::Duration;
+    use tokio::net::{TcpListener, TcpStream};
+
+    #[tokio::test]
+    async fn select_proto_basic() {
+        async fn run(version: Version) {
+            let (client_connection, server_connection) = futures_ringbuf::Endpoint::pair(100, 100);
+
+            let server: tokio::task::JoinHandle<Result<(), ()>> = tokio::spawn(async move {
```
dmitry-markin, line 426: nit: Is the type annotation required here and with other `tokio::spawn()` invokations?

src/multistream_select/dialer_select.rs
```diff
@@ -346,3 +410,358 @@ impl DialerState {
...
+            // In Libp2p the lazy negotation of protocols can be closed at any time,
+            // even if the negotiation is not yet done.
+
+            // However, for the Litep2p the negotation must conclude before closing the
+            // lazy negotation of protocol. We'll wait for the close until the
+            // server has produced a message, in this test that means forever.
+            io.close().await.unwrap();
+        });
+
+        assert!(tokio::time::timeout(Duration::from_secs(10), client).await.is_err());
```
dmitry-markin, line 556: nit: May be add a remark that we should change this to `.is_ok()` and update the comment above after merging https://github.com/paritytech/litep2p/pull/62?

dmitry-markin (approved):
LGTM! Nice use of `futures_ringbuf` in tests, didn't know about it.

src/multistream_select/dialer_select.rs
lexnv: Unfortunately, I think so because we only return `Ok` from those futures, and the compiler cannot deduce the `Err` type

dmitry-markin: If we never return `Err`, why do we need to return `Result`?

lexnv: Thanks for spotting this! Don't remember why I've added the `Ok()`, probably wanted to use the `?` operator. I've removed the type annotation 🙏

PR #62, opened 2024-03-18
multistream/negotiated: Propagate poll_close on unreceived protocol messages

lexnv:
This PR ensures that lazily negotiated protocols can be closed before the specific protocol message has been received.

Closes: https://github.com/paritytech/litep2p/issues/60

cc @paritytech/networking

### DIFF
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -415,6 +415,8 @@ impl DialerState {
 mod tests {
     use super::*;
     use crate::multistream_select::listener_select_proto;
+    use std::time::Duration;
+    use tokio::net::{TcpListener, TcpStream};
 
     #[tokio::test]
     async fn select_proto_basic() {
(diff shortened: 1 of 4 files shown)


src/multistream_select/dialer_select.rs
```diff
@@ -544,16 +544,12 @@ mod tests {
                 dialer_select_proto(client_connection, protos, Version::V1Lazy).await.unwrap();
             assert_eq!(proto, "/proto1");
 
-            // In Libp2p the lazy negotation of protocols can be closed at any time,
+            // The lazy negotation of protocols can be closed at any time,
             // even if the negotiation is not yet done.
-
-            // However, for the Litep2p the negotation must conclude before closing the
-            // lazy negotation of protocol. We'll wait for the close until the
-            // server has produced a message, in this test that means forever.
```
dmitry-markin, line 552: I'm curious why it was implemented like this in the first place. Aren't we in a Chesterton's fence situation here?

dmitry-markin (approved):

src/multistream_select/dialer_select.rs
lexnv: Yep indeed, I wonder if this affects something in substrate. Proly would be good to have it running in versi for a while. I'll wait a bit for that and group this PR with the clippy PR (and any others) for testing

lexnv: cc @altonen would love to get your feedback on this if you remember why the change was needed 🙏

altonen: I honestly don't even remember touching the `multistream-select` code. Did you check the behavior in this test against the latest `multistream-select` that libp2p uses or some other version? I wonder if it could be related to this: https://github.com/paritytech/substrate/pull/14691

lexnv:
No longer needed, this has been included in the latest v0.9.1 release

PR #63, opened 2024-03-18
tests: Add sysctl ipv6 and extra CI steps

lexnv:
Changes the binding in response to 
- https://github.com/docker/for-linux/issues/1374
This is achieved via `--sysctl net.ipv6.conf.all.disable_ipv6=0`

CI steps that are running now:
- cargo fmt
- cargo machete for unused dependencies (has detected one)
- cargo check
- cargo doc (with extra checks)
- cargo clippy
- cargo test

While at it, I've also fixed cargo doc tests.

I've added `reorder_imports` (similar to Substrate) to group imports from the same crate.

At the time being, the clippy step is allowed to fail to not block this PR further and a proper fix should land with: https://github.com/paritytech/litep2p/pull/57.

Builds on https://github.com/paritytech/litep2p/pull/56

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,7 @@ jobs:
   test:
     name: "test"
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     container:
       image: ${{ vars.CI_IMAGE }}
       # options: --privileged
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,7 @@ jobs:
           cache-on-failure: true
           cache-all-crates: true
       - name: Check documentation
-        run: RUSTDOCFLAGS="--deny rustdoc::broken_intra_doc_links" cargo doc --workspace --no-deps --document-private-items
+        run: RUSTDOCFLAGS="-D warnings -D rustdoc::broken_intra_doc_links" cargo doc --workspace --no-deps --document-private-items
 
       - if: "failure()"
         uses: "andymckay/cancel-action@271cfbfa11ca9222f7be99a47e8f929574549e0a" # v0.4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,14 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
 
+      - name: Install Rust nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: rustfmt
+
       - name: Rust Cache
         uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,13 +21,6 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
 
-      - name: Install Rust nightly toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          components: rustfmt
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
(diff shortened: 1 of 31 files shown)


.github/workflows/ci.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     container:
```
alvicsam, line 40:
```suggestion
    timeout-minutes: 20
    container:
```
Can we also add time-out here? Because it seems that the job stuck and default timeout is 360 minutes

lexnv:
I think we could keep the `sysctl` fix, I'll revert my binding changes, good point thanks!

lexnv:
@dmitry-markin Let me know what you think of it, since I've change quite a few things.

CI steps that are running now:
- cargo fmt
- cargo machete for unused dependencies (has detected one)
- cargo check
- cargo doc (with extra checks)
- cargo clippy
- cargo test

While at it, I've also fixed cargo doc tests.

I've added `reorder_imports` (similar to Substrate) to group imports from the same crate.
Since these are quite opinionated, would love to get your input on them, we could otherwise revert the parts you think are unnecessary / too restrictive 🙏 

At the time being, the clippy step is allowed to fail to not block this PR further and a proper fix should land with: https://github.com/paritytech/litep2p/pull/57.

.github/workflows/ci.yml
```diff
@@ -12,98 +12,144 @@ concurrency:
...
       - name: Rust Cache
-        uses: Swatinem/rust-cache@v2.5.0
+        uses: Swatinem/rust-cache@23bce251a8cd2ffc3c1075eaa2367cf899916d84 # v2.7.3
         with:
           cache-on-failure: true
           cache-all-crates: true
-      - name: doc
-        run: cargo doc
+      - name: Check documentation
+        run: RUSTDOCFLAGS="-D warnings -D rustdoc::broken_intra_doc_links" cargo doc --workspace --no-deps --document-private-items
```
dmitry-markin, line 104: Haven't checked,  `--document-private-items` does not generate an error if a private element is not documented, does it?

.github/workflows/ci.yml
```diff
@@ -12,98 +12,144 @@ concurrency:
...
 
-      - name: check
-        run: cargo check
+      - name: Cargo fmt
+        run: cargo +nightly fmt --all -- --check
 
-  test:
-    name: "test"
+      - if: "failure()"
+        uses: "andymckay/cancel-action@271cfbfa11ca9222f7be99a47e8f929574549e0a" # v0.4
```
dmitry-markin, line 42: Does this cancel `Cargo fmt` check only, or all the CI checks? I would make checks to be independent.

dmitry-markin (approved):

dmitry-markin:
> I've added `reorder_imports` (similar to Substrate) to group imports from the same crate.
> Since these are quite opinionated, would love to get your input on them, we could otherwise revert the parts you think are unnecessary / too restrictive 🙏

Makes sense to me. Overall, you changes seem to be a big improvement, thank you!

One thing I don't understand / don't know how to check are all that github CI actions like:
```
uses: "andymckay/cancel-action@271cfbfa11ca9222f7be99a47e8f929574549e0a"
```
Not sure, how do we select the audited/trusted ones? (But may be if they work that's all we need to know.)

.github/workflows/ci.yml
```diff
@@ -12,98 +12,144 @@ concurrency:
...
-      # - name: fmt
-      #   run: |
-      #     cargo fmt --all --check
+      - name: Install Rust nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: rustfmt
```
alvicsam, line 30:
```suggestion
```
This part is not needed since we run the action in [the `ci-unified`docker image](https://github.com/paritytech/scripts/tree/master/dockerfiles/ci-unified). Also the `actions-rs` org is archived and the actions are not maintained anymore.

.github/workflows/ci.yml
lexnv: This would cancel all running CI steps. We've been using it in subxt to reduce the CI minutes in case of failure.
However, I've removed this from litep2p since I believe we don't consume that many minutes 👍

PR #64, opened 2024-03-20
Add support for protocol/agent version

altonen:
Add ability to specify IPFS Identify protocol version and user agent. The implementation follows rust-libp2p, making protocol version mandatory and user agent is optional, which defaults to `litep2p/1.0.0`.

This needs a Polkadot SDK "companion" because it changes the public API. I can make the fix if I have push rights to the PR. The changes are pretty trivial though, just adding two new parameters to `IdentifyConfig::new()`.

PR #65, opened 2024-03-21
[ci] Run ci for PR from forks

alvicsam:
PR adds additional step to run all jobs from a container. The step is needed because when CI runs from forks repo variables are not accessible. 

cc https://github.com/paritytech/litep2p/pull/64#issuecomment-2011683117
cc https://github.com/paritytech/ci_cd/issues/964


dmitry-markin (approved):
Thank you!

PR #68, opened 2024-03-28
protocol/mdns: Use `SO_REUSEPORT` for the mDNS socket

altonen:
On macOS, failure to use `SO_REUSEPORT` would result in mDNS not being able to bind to the socket, causing `Litep2p::new()` to exit with `EADDRINUSE`.

### DIFF
--- a/src/protocol/mdns.rs
+++ b/src/protocol/mdns.rs
@@ -128,9 +128,8 @@ impl Mdns {
     ) -> crate::Result<Self> {
         let socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP))?;
         socket.set_reuse_address(true)?;
-        // TODO: fix
-        // #[cfg(unix)]
-        // socket.set_reuse_port(true)?;
+        #[cfg(unix)]
+        socket.set_reuse_port(true)?;
         socket.bind(
             &SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), IPV4_MULTICAST_PORT).into(),
         )?;


src/protocol/mdns.rs
```diff
@@ -128,9 +128,8 @@ impl Mdns {
     ) -> crate::Result<Self> {
         let socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(Protocol::UDP))?;
         socket.set_reuse_address(true)?;
-        // TODO: fix
-        // #[cfg(unix)]
-        // socket.set_reuse_port(true)?;
+        #[cfg(unix)]
+        socket.set_reuse_port(true)?;
```
lexnv, line 132: dq: What was the reason this was commented out in the past with a "todo: fix"? Did something break previously?

lexnv (approved):
LGTM! Just one question about the prior comment

dmitry-markin:
> On macOS, failure to use `SO_REUSEPORT` would result in mDNS not being able to bind to the socket, causing `Litep2p::new()` to exit with `EADDRINUSE`.

Is there an explanation for macOS behaving this way?

dmitry-markin (approved):

src/protocol/mdns.rs
altonen: AFAIR when I wrote the code, I was using an incorrect version of `socket2` or something which caused compilation to fail on Linux even though I had `#[cfg(unix)]`. Since the code worked without `port_reuse(true)`, I just commented the code out and was planning on revisiting it in the future but never did until now.

altonen:
> > On macOS, failure to use `SO_REUSEPORT` would result in mDNS not being able to bind to the socket, causing `Litep2p::new()` to exit with `EADDRINUSE`.
> 
> Is there an explanation for macOS behaving this way?

[This](https://stackoverflow.com/questions/51998042/macos-so-reuseaddr-so-reuseport-not-consistent-with-linux) is the best reference I could find after a quick search. Looks like there are subtle differences between the operating systems which makes mDNS work on Linux despite not enabling `SO_REUSEPORT`

PR #69, opened 2024-03-29
Expose `reuse_port` option for TCP and WebSocket transports

dmitry-markin:
Add `reuse_port` option to TCP & WebSocket configs. When set, this both:
1. Sets `SO_REUSEPORT` on sockets.
2. Binds sockets to the listen address port when making outbound connections.

I'm not sure if these two behaviors should be controlled by a single option, but this follows libp2p convention.

CC @altonen

### DIFF
--- a/src/transport/tcp/config.rs
+++ b/src/transport/tcp/config.rs
@@ -30,8 +30,16 @@ use crate::{
 pub struct Config {
     /// Listen address for the transport.
     ///
-    /// Default listen addres is `/ip6/::1/tcp`.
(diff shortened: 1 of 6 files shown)


dmitry-markin:
While at this PR I noticed that our `rustfmt.toml` doesn't include `match_block_trailing_comma = true`, which is present in poladot-sdk `rustfmt.toml`. I'm going to add it and reformat the repo in a follow-up PR if you don't mind @altonen @lexnv

altonen (approved):

PR #71, opened 2024-04-02
Bring `rustfmt.toml` in sync with polkadot-sdk

dmitry-markin:
Make `rustfmt.toml` identical to the polkadot-sdk's one.

Personally, I don't like some choices like `binop_separator = "Back"`, `match_arm_leading_pipes = "Preserve"`, and `hard_tabs = true`, but I don't think it makes sense having different code style settings in polkadot-sdk & litep2p.

CC @altonen

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,14 @@
 fn main() {
-    prost_build::compile_protos(
-        &[
-            "src/schema/keys.proto",
-            "src/schema/noise.proto",
-            "src/schema/webrtc.proto",
-            "src/protocol/libp2p/schema/identify.proto",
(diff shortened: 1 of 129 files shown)


dmitry-markin:
The first commit is the one with meaningful changes.

altonen:
I don't see why litep2p as a dependency of Polkadot SDK would need to follow its formatting rules. Having said that, I think it's more important that there is a config that people can agree on and that it doesn't have to be changed every 3 months.

dmitry-markin:
I merged this PR and only afterwards realized this is going to cause conflicts with existing @lexnv PRs. Probably should revert it...

PR #78, opened 2024-04-18
kad: Limit MemoryStore entries

lexnv:
This PR ensures that the memory store entries are not growing indefinitely, which may cause:
- a memory leak and process being terminated
- using outdated records

The following changes are introduced: 
- records are stored in the memory store iff space is available
  - a record can replace another record iff it outlives the currently stored record
- records are removed from the store if they expire when the record is inspected
- record value must not exceed a sensible number of bytes

cc @dmitry-markin

### DIFF
--- a/src/protocol/libp2p/kademlia/store.rs
+++ b/src/protocol/libp2p/kademlia/store.rs
@@ -121,3 +121,95 @@ impl Default for MemoryStoreConfig {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_memory_store() {
+        let mut store = MemoryStore::new();
+        let key = Key::from(vec![1, 2, 3]);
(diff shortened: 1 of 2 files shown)


src/protocol/libp2p/kademlia/store.rs
```diff
@@ -32,28 +32,184 @@ pub enum MemoryStoreEvent {}
...
+                        return;
+                    }
+                }
+
+                entry.insert(record);
+            }
+
+            Entry::Vacant(entry) => {
+                if len >= self.config.max_records {
+                    return;
```
dmitry-markin, line 94: May be try cleaning up free space here by removing expired records? Out of scope of this PR, as this would require storing records both sorted by their TTL & hashed by their key. But something to consider as a low-priority follow up.

dmitry-markin (approved):
LGTM!

PR #82, opened 2024-04-19
kad: Implement put_record_to and try_put_record_to (#77)

lexnv:
This PR implements the `put_record_to` and `try_put_record_to` to selectively pick peers to update their records.

The main use-case from substrate would be the following:
- A peer is discovered to have an outdated authority record (needs https://github.com/paritytech/litep2p/pull/76)
- Update the record with the latest authority record available (part of this PR)

This PR provided peers to the engine if the peers are part of the kBucket. The first step of the discovery in substrate motivates this assumption. We can probably do things a bit more optimally since we know the peers part of the kBucket were discovered previously (or currently connected):
- The query starts with a [FindNodeContext](https://github.com/paritytech/litep2p/blob/96e827b54f9f937c6d0489bef6a438b48cf50e58/src/protocol/libp2p/kademlia/query/find_node.rs#L37), which in this case will do a peer discovery as well
- We could implement a `PutNodeContext` which circumvents the need to discover the peers and just forwards a kad `PUT_VALUE` to those peers We'd have to double check that with libp2p as well (my brief looking over code points to this direction).

To unblock https://github.com/paritytech/polkadot-sdk/pull/3786 we can merge this and then come back with a better / optimal solution for this

Builds on top of: https://github.com/paritytech/litep2p/pull/76

cc @paritytech/networking

---------

### DIFF
--- a/src/protocol/libp2p/kademlia/query/find_many_nodes.rs
+++ b/src/protocol/libp2p/kademlia/query/find_many_nodes.rs
@@ -27,7 +27,8 @@ use crate::{
 };
 
 /// Context for multiple `FIND_NODE` queries.
-// TODO: implement https://github.com/paritytech/litep2p/issues/80.
+// TODO: implement finding nodes not present in the routing table,
+//       see https://github.com/paritytech/litep2p/issues/80.
 #[derive(Debug)]
 pub struct FindManyNodesContext {
     /// Query ID.
--- a/src/protocol/libp2p/kademlia/query/find_many_nodes.rs
+++ b/src/protocol/libp2p/kademlia/query/find_many_nodes.rs
@@ -1,0 +1,64 @@
+// Copyright 2023 litep2p developers
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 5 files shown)


lexnv:
Forgot to change branch of https://github.com/paritytech/litep2p/pull/77 to master 🙈

src/protocol/libp2p/kademlia/query/find_many_nodes.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::{
+    protocol::libp2p::kademlia::{
+        query::{QueryAction, QueryId},
+        types::KademliaPeer,
+    },
+    PeerId,
+};
+
+/// Context for multiple `FIND_NODE` queries.
+// TODO: implement https://github.com/paritytech/litep2p/issues/80.
```
dmitry-markin, line 30:
```suggestion
// TODO: implement finding nodes not present in the routing table,
//       see https://github.com/paritytech/litep2p/issues/80.
```

dmitry-markin (approved):

PR #88, opened 2024-04-24
kad/store: Set memory-store on an incoming record for PutRecordTo

lexnv:
Small PR to always update the local store on `PutRecordToPeers`.

This is introduced to facilitate the use-case from:
- https://github.com/paritytech/polkadot-sdk/pull/3786/files#diff-5c2ed434b83c3fc3e5b6e2d2bc1895ec870d39e6b77a0f44b3f012694b5a943aR426

Where the `update_local_store` flag is used to optimize kad MemoryStore updates. However, for us this represents a hashmap insert.

While at it, have also filtered out peers that are equal to our local peer ID. 

cc @dmitry-markin

### DIFF
--- a/src/protocol/libp2p/kademlia/handle.rs
+++ b/src/protocol/libp2p/kademlia/handle.rs
@@ -102,6 +102,9 @@ pub(crate) enum KademliaCommand {
 
         /// Use the following peers for the put request.
         peers: Vec<PeerId>,
+
+        /// Update local store.
+        update_local_store: bool,
     },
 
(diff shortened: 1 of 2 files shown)


dmitry-markin:
It looks like this is different to what [libp2p does](https://docs.rs/libp2p/0.51.3/libp2p/kad/struct.Kademlia.html#method.put_record_to). May be keep the operations consistent between libraries until we keep only one of them and handle this in the client code?

dmitry-markin (approved):
LGTM!

PR #89, opened 2024-04-24
ci: Enable dependabot

lexnv:
Now that we have testing in our CI, it would be beneficial to pull in the latest crate dependencies 

cc @dmitry-markin


dmitry-markin (approved):
Wasn't it already enabled by default?

PR #95, opened 2024-04-25
cargo: Update str0m from 0.4.1 to 0.5.0

lexnv:
This PR updates str0m from 0.4.1 to 0.5.0.

This manually implements the decoding of ice-usr/ice-pwd, however this PR should unblock us upstream: https://github.com/algesten/str0m/pull/505.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ simple-dns = "0.5.3"
 smallvec = "1.10.0"
 snow = { version = "0.9.3", features = ["ring-resolver"], default-features = false }
 socket2 = { version = "0.5.5", features = ["all"] }
-str0m = "0.4.1"
+str0m = "0.5.0"
 thiserror = "1.0.39"
 tokio-stream = "0.1.12"
 tokio-tungstenite = { version = "0.20.0", features = ["rustls-tls-native-roots"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -36,7 +36,7 @@ simple-dns = "0.5.3"
 smallvec = "1.10.0"
 snow = { version = "0.9.3", features = ["ring-resolver"], default-features = false }
 socket2 = { version = "0.5.5", features = ["all"] }
-str0m = "0.4.1"
+str0m = "0.5.0"
```
jasl, line 39: `0.5.1` just out, maybe you can upgrade to the latest one?

src/transport/webrtc/mod.rs
```diff
@@ -712,3 +728,9 @@ impl Stream for WebRtcTransport {
             .map_or(Poll::Pending, |event| Poll::Ready(Some(event)))
     }
 }
+
+// Check if the packet received is STUN.
```
dmitry-markin, line 732: Is there a spec to reference to explain the check below?

dmitry-markin (approved):

src/transport/webrtc/mod.rs
lexnv: Yep, I ve added a link to the docs: https://datatracker.ietf.org/doc/html/rfc5389#page-10

PR #96, opened 2024-04-29
kad: Expose all peer records of `GET_VALUE` query

lexnv:
This PR extends the `GetRecordSucess` variant of kad queries to provide all the query information.
Previously, only the last stored query record was returned to the user.

In the libp2p crate, an event is generated for every found record. However, litep2p responds only when the query finishes.
Records are extended to offer similar information back to the user while sending a minimum number of events.

- A new `RecordsType` is introduced to avoid confusion about records extracted from the local store and records found on the network
- We save the record backed by the most number of peers in our local store
- A panic is removed from the `kad/query` since we now operate with a multitude of records and potentially none
- Expired records are now ignored

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -52,7 +52,7 @@ pub use handle::{KademliaEvent, KademliaHandle, Quorum, RoutingTableUpdateMode};
 pub use query::QueryId;
 pub use record::{Key as RecordKey, PeerRecord, Record};
 
-use self::handle::RecordsType;
+pub use self::handle::RecordsType;
 
 /// Logging target for the file.
 const LOG_TARGET: &str = "litep2p::ipfs::kademlia";
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -870,7 +870,7 @@ mod tests {
         event_rx: Receiver<KademliaEvent>,
     }
 
-    fn _make_kademlia() -> (Kademlia, Context, TransportManager) {
+    fn make_kademlia() -> (Kademlia, Context, TransportManager) {
         let (manager, handle) = TransportManager::new(
             Keypair::generate(),
             HashSet::new(),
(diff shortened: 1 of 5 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -52,6 +52,8 @@ pub use handle::{KademliaEvent, KademliaHandle, Quorum, RoutingTableUpdateMode};
 pub use query::QueryId;
 pub use record::{Key as RecordKey, PeerRecord, Record};
 
+use self::handle::RecordsType;
```
alexggh, line 55: pub, so we can use this type to deconstruct it, or maybe we can try to keep the libp2p semantics for PeerRecord and we don't need this extra type:
```
/// A record either received by the given peer or retrieved from the local
/// record store.
#[derive(Debug, Clone, PartialEq, Eq)]
pub struct PeerRecord {
    /// The peer from whom the record was received. `None` if the record was
    /// retrieved from local storage.
    pub peer: Option<PeerId>,
    pub record: Record,
}
```

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -636,11 +638,32 @@ impl Kademlia {
...
+                let rec = records
+                    .iter()
+                    .map(|peer_record| &peer_record.record)
+                    .fold(HashMap::new(), |mut acc, rec| {
+                        *acc.entry(rec).or_insert(0) += 1;
+                        acc
+                    })
+                    .into_iter()
+                    .max_by_key(|(_, v)| *v)
+                    .map(|(k, _)| k);
```
dmitry-markin, line 654: Would be nice to have tests for correctly updating the local store and also discarding the expired records. At least as a follow-up planned if it's too much work to implement such tests.

dmitry-markin (approved):
LGTM!

lexnv:
Thanks @dmitry-markin for the review 🙏

I've added to tests to validate the kademlia behavior in this case:
- one test to see that the record backed by most peers is added to the local storage
- one test to ensure that expired records don't count towards the records frequency

PR #97, opened 2024-04-29
kad: Refactor `GetRecord` query and add tests

lexnv:
This PR refactors the `GetRecord` query to make it more robust.
- Avoid panics on unwraps and unimplemented logic
- Separate config immutable variables from main query logic
- Simplifies logic of `next_action` method 
- Private methods for internal logic

Builds on top of: https://github.com/paritytech/litep2p/pull/96

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/query/get_record.rs
+++ b/src/protocol/libp2p/kademlia/query/get_record.rs
@@ -136,9 +136,10 @@ impl GetRecordContext {
             return;
         };
 
-        // TODO: validate record
         if let Some(record) = record {
-            self.found_records.entry(record).or_default().push(peer.peer);
+            if !record.is_expired(std::time::Instant::now()) {
+                self.found_records.entry(record).or_default().push(peer.peer);
+            }
         }
 
         // add the queried peer to `queried` and all new peers which haven't been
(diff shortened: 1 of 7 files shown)


src/protocol/libp2p/kademlia/query/get_record.rs
```diff
@@ -36,22 +36,55 @@ use std::collections::{BTreeMap, HashMap, HashSet, VecDeque};
...
 #[derive(Debug)]
-pub struct GetRecordContext {
+pub struct GetRecordConfig {
     /// Local peer ID.
-    local_peer_id: PeerId,
+    pub local_peer_id: PeerId,
 
-    /// How many records have been successfully found.
+    /// How many records we already know about (ie extracted from storage).
     pub record_count: usize,
```
dmitry-markin, line 46: May be rename this to `known_records`? And add a comment that this can be either `0` or `1`.

src/protocol/libp2p/kademlia/query/get_record.rs
```diff
@@ -145,96 +165,342 @@ impl GetRecordContext {
...
+            replication_factor: 20,
+            ..default_config()
+        };
+        assert!(config.sufficient_records(20));
+        assert!(!config.sufficient_records(19));
+
+        // Quorum::All with 10 known records.
+        let config = GetRecordConfig {
+            quorum: Quorum::All,
+            record_count: 10,
```
dmitry-markin, line 307: nit: It feels a little bit odd passing 10 where only 1 or 0 can be in real life... Also below.

dmitry-markin (approved):
LGTM!

Issue #100, opened 2024-05-06
kad: Revisit `FIND_NODE` and increase test coverage


lexnv:
Refactor the `FIND_NODE` query to make the code more robust
- avoid panics / unwraps / unimplemented 
- add an immutable config for the main query logic
- simplify the `next_action` method to handle cases in a graceful way
- add tests to increase test coverage
  - check the fulfillment of parallelism queries
  - no candidates for further actions
  - responding peers that we didn't query
  - completion logic etc

Similar to https://github.com/paritytech/litep2p/pull/97

PR #114, opened 2024-05-21
kad: Refactor FindNode query, keep K best results and add tests

lexnv:
This PR refactors the FindNode query to make it more robust.
 - Avoid panics on unwraps and unimplemented logic
 - Separate config immutable variables from main query logic
 - Simplifies logic of next_action method
 - Private methods for internal logic

It also fixes a bug where only the last peer ID of our results was updated.
- we were registering the first 20 (repl factor) responses
- whenever we got a new response (better) we tried to update just the peerID_ of the last entry

This had a few implications:
- the first 19 entries are never updated with a better (closer) peer
- the distance of the last entry was never updated. And because of this, we might end up in a situation where the record is updated with a furthest distance

```rust
// Presume: 
last_entry: distance = 10, peer = A

// Registering a new response
new_response: distance = 2, peer = B
-> last entry: distance = 10, peer = B

// Registering a new response
new_response: distance = 7, peer = C
-> last entry: distance = 10, peer = C

// we should've kept B as best
```

### Testing Done
- Completes if there's no further candidate to query
- Fulfill number of parallel queries
- Completes on the number of responses
- Provides the closest response out of the batch
- Provides the closest response when the closest peer is indirectly found and K factor is already fulfilled

Closes https://github.com/paritytech/litep2p/issues/100

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/query/find_node.rs
+++ b/src/protocol/libp2p/kademlia/query/find_node.rs
@@ -246,6 +246,7 @@ impl<T: Clone + Into<Vec<u8>>> FindNodeContext<T> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::protocol::libp2p::kademlia::types::ConnectionType;
 
     fn default_config() -> FindNodeConfig<Vec<u8>> {
         FindNodeConfig {
@@ -257,6 +258,15 @@ mod tests {
         }
     }
 
+    fn peer_to_kad(peer: PeerId) -> KademliaPeer {
+        KademliaPeer {
+            peer,
+            key: Key::from(peer),
+            addresses: vec![],
+            connection: ConnectionType::Connected,
+        }
+    }
+
     #[test]
     fn completes_when_no_candidates() {
         let config = default_config();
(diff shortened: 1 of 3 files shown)


src/protocol/libp2p/kademlia/query/find_node.rs
```diff
@@ -117,120 +122,373 @@ impl<T: Clone + Into<Vec<u8>>> FindNodeContext<T> {
...
         //  a) the map doesn't have 20 responses
         //  b) it can replace some other peer that has a higher distance
-        let distance = self.target.distance(&peer.key);
+        let distance = self.config.target.distance(&peer.key);
 
         // always mark the peer as queried to prevent it getting queried again
         self.queried.insert(peer.peer);
 
-        // TODO: could this be written in another way?
-        // TODO: only insert nodes from whom a response was received
```
dmitry-markin, line 126: Is this TODO still relevant? It looks like nothing has changed in its respect.

dmitry-markin (approved):
Nice PR! And good to see we have `FIND_NODE` tested now.

src/protocol/libp2p/kademlia/query/find_node.rs
lexnv: I dont think this is addressed indeed, will briefly look over libp2p or create an issue if this may be too involved / not covered by our testing atm, thanks!

lexnv: While looking at this I realized a few things:
- we were registering the first 20 (repl factor) responses
- whenever we got a new response (better) we tried to update just the _peerID__ of the __last entry__

This has a few implications:
- the first 19 entries are never updated with a better (closer) peer
- the distance of the last entry was never updated. And because of this, we might end up in a situation where the record is updated with a furthest distance

```rust
// Presume: 
last_entry: distance = 10, peer = A

// Registering a new response
new_response: distance = 2, peer = B
-> last entry: distance = 10, peer = B

// Registering a new response
new_response: distance = 7, peer = C
-> last entry: distance = 10, peer = C

// we should've kept B as best
```

In this case, we should have updated other records and only keep the window of the best 20 peers.

To mitigate this, if we receive a response better than the last entry: 
- insert the response with distance and peer to our btreemap, if and only if we didn't track that distance before
- pop the last entry to keep only K entries

Have also added a test for this `keep_k_best_results`

src/protocol/libp2p/kademlia/query/find_node.rs
```diff
@@ -131,9 +131,18 @@ impl<T: Clone + Into<Vec<u8>>> FindNodeContext<T> {
...
-            if let Some(mut entry) = self.responses.last_entry() {
-                if entry.key() > &distance {
-                    entry.insert(peer);
+            // Find the furthest distance.
+            let furthest_distance =
+                self.responses.last_entry().map(|entry| *entry.key()).unwrap_or(distance);
+
+            // The response received from the peer is closer than the furthest response.
+            if distance < furthest_distance {
+                // Update the entries only if the distance is not already present.
```
dmitry-markin, line 140: What is the drawback of overwriting the entry with this distance? I.e., there should be exactly one peer with such distance as it uniquely identifies the hash, so it should be safe to overwrite without checking.

dmitry-markin (approved):

src/protocol/libp2p/kademlia/query/find_node.rs
lexnv: I think I followed the libp2p here. However, I believe we can overwrite the entries without checking here, and we'll have a "fresher" report in the end. Will do that, thanks!

Issue #103, opened 2024-05-06
identify: Open inbound substream for unknown peer


lexnv:
There seems to be a state mismatch in the identify protocol implementation.

The following warning appeared 44 times on a warp sync node and 8 times on a full sync node, both nodes were running for almost 2 weeks.

```
2024-05-06 03:16:41.456  WARN tokio-runtime-worker litep2p::ipfs::identify: inbound identify substream opened for peer who doesn't exist peer=PeerId("12D3KooWA5taCbB5CEcz1VS126XjcbwBYqm9GmsASX5mBZU8611v") protocol=/ipfs/id/1.0.0
```

The warning is coming from:

https://github.com/paritytech/litep2p/blob/42ed1c1463a8b11eac5344af2871ba752335b06e/src/protocol/libp2p/identify.rs#L255-L266

This may happen when:
1. `on_connection_established` fails to open a new substream, either because the peerID isn't connected or the connection is clogged
2. `on_connection_closed` is called before `on_inbound_substream`
3. `on_inbound_substream` is called before `on_connection_established`

cc @paritytech/networking

altonen:
Does this happen when you run the node with `--profile testnet`/`--release`? I remember when I was testing litep2p with higher load, the debug build really started to struggle. FWIW, I only saw these issues emerge when I switched from tokio to Substrate's task executor so I think there might be something there but I didn't see any issues after running polkadot in release mode

lexnv:
I started two nodes from a release build and left them running for ~2weeks:

```
./target/release/polkadot -d /home/ubuntu/workspace/kusama-db --chain kusama --pruning=1000 --sync=warp --network-backend litep2p --detailed-log-output

./target/release/polkadot -d /home/ubuntu/workspace/kusama-db-full --chain kusama --port 30355 --pruning=1000 --network-backend litep2p --detailed-log-output
```

I'm happy to announce that both are still producing blocks. Good point about the executor, it might be that my VM was overloaded. I'm running 2 nodes on the same machine and doing dev-work at the same time (compile substrate / subxt, running test nodes etc) which might put extra stress on the system

lexnv:
After performing more intensive testing on this, with a number of 500 in/out peers, the issue is reproducing quite often.

Will have a look at this soon to filter out any warnings and see more trace logs 

```yaml
[...]

PR #133, opened 2024-05-29
ping: Make ping fault tolerant wrt outbound substreams races

lexnv:
We have discovered a race between the TransportService and Protocol interface.

This issue was first seen on the identify protocol:
- https://github.com/paritytech/litep2p/issues/103

After looking a bit closely into it, indeed there might be a chance for the following to happen:
- T0: Transport service sends `TransportEvent::ConnectionEstablished` and `TransportEvent::SubstreamOpened` 
- T1: The peer disconnects
- T2: Ping protocol receives an `TransportEvent::ConnectionEstablished` event for a disconnected peer
- T3: `Ping::on_connection_established` fails to open a substream and `pending_open` is never populated
- T4: Ping protocol receives an `TransportEvent::SubstreamOpened` event which would have panicked litep2p

This PR replaces the `todo!` in case of the race with a warning, and we'll no longer panic.

It is still good to capture those as warnings for visibility, even if there's no implied side-effect on our code.

### DIFF
--- a/src/protocol/libp2p/ping/mod.rs
+++ b/src/protocol/libp2p/ping/mod.rs
@@ -197,7 +197,11 @@ impl Ping {
                                     self.on_outbound_substream(peer, substream_id, substream);
                                 }
                                 None => {
-                                    todo!("substream {substream_id:?} does not exist");
+                                    tracing::warn!(
+                                        target: LOG_TARGET,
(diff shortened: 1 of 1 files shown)


dmitry-markin (approved):
Nice catch! Definitely logging is better then panicking here.

PR #135, opened 2024-06-03
Support manual DHT record insertion

dmitry-markin:
Support manual DHT record insertion to comply with `libp2p` implementation.

### DIFF
--- a/src/protocol/libp2p/kademlia/store.rs
+++ b/src/protocol/libp2p/kademlia/store.rs
@@ -78,7 +78,7 @@ impl MemoryStore {
                 target: LOG_TARGET,
                 key = ?record.key,
                 publisher = ?record.publisher,
-                "discarding a DHT record that exceeds the configured limit",
+                "discarding a DHT record that exceeds the configured size limit",
             );
             return;
         }
--- a/src/protocol/libp2p/kademlia/store.rs
+++ b/src/protocol/libp2p/kademlia/store.rs
@@ -25,6 +25,9 @@ use crate::protocol::libp2p::kademlia::record::{Key, Record};
 
 use std::collections::{hash_map::Entry, HashMap};
 
+/// Logging target for the file.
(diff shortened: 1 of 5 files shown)


src/protocol/libp2p/kademlia/store.rs
```diff
@@ -25,6 +25,9 @@ use crate::protocol::libp2p::kademlia::record::{Key, Record};
 
 use std::collections::{hash_map::Entry, HashMap};
 
+/// Logging target for the file.
+const LOG_TARGET: &str = "litep2p::ipfs::store";
```
lexnv, line 29: dq: Are we using ipfs in other places? Does it make sens for us to use `kad` here instead?

tests/protocol/kademlia.rs
```diff
@@ -122,3 +125,312 @@ async fn put_value() {
...
+                panic!("record was not stored in 10 secs")
+            }
+            _ = litep2p1.next_event() => {}
+            _ = litep2p2.next_event() => {}
+            _ = kad_handle1.next() => {}
+            event = kad_handle2.next() => {
+                match event {
+                    Some(KademliaEvent::IncomingRecord { record }) => {
+                        // Check if the record was stored.
+                        let _ = kad_handle2
```
lexnv, line 176: nit; We could probably compare the records here

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: Yes, we are using `litep2p::ipfs::kademlia` in other places. Corrected.

tests/protocol/kademlia.rs
dmitry-markin: Thanks, done.

lexnv: Perfect, thanks!

lexnv (approved):
LGTM! Nice work here!

alexggh (approved):

PR #141, opened 2024-06-05
ci: Change ci machine for tests

lexnv:


alvicsam:
@lexnv Do you think that the tests are failing due to lack of the resources?

lexnv:
I'm not entirely sure, wanted to exclude disk space by using a different ubuntu image 

From https://github.com/paritytech/litep2p/actions/runs/9351594490/job/25737765035:

```
 Filesystem      Size  Used Avail Use% Mounted on
overlay          73G   65G  8.0G  90% /
```

It might be possible that `cargo tests` now uses more space than before 🤔

alvicsam:
Yes, this can be, when tests finish successfully the runner has only 3% of space left (~3GB). I'm trying to find out if it's possible to clean up some space when job is running in a container in [this PR](https://github.com/paritytech/litep2p/pull/137). If I can't find a solution then we could install our self-hosted github runners for this repo and try them.

lexnv:
Handled by: https://github.com/paritytech/litep2p/pull/148

PR #143, opened 2024-06-06
protocol/notif: Fix panic on missing peer state 

lexnv:
This PR aims to remove the possible panics from the `protocol/notification` that may happen on missing peer contexts.

cc @paritytech/networking


dmitry-markin (approved):
Looks good!

One thing to consider is whether it's a logic error that peer doesn't exist (internal state mismatch), or some possible thing during normal operation. In the first case I would promote logging to `error!` and do `debug_assert!` to fix the inconsistency if it happens in the wild. But in the second case I wouldn't do `debug_assert!` and also may be demote the message to `debug!`.

PR #146, opened 2024-06-10
transport: Make `TCP_NODELAY` configurable

altonen:
Allow configuring `TCP_NODELAY` through TCP/WebSocket configurations. Polkadot configures it to `true` btw: https://github.com/paritytech/polkadot-sdk/blob/b65313e81465dd730e48d4ce00deb76922618375/substrate/client/network/src/transport.rs#L58


src/transport/tcp/mod.rs
```diff
@@ -201,6 +202,7 @@ impl TcpTransport {
 
         match dial_addresses.local_dial_address(&remote_address.ip()) {
             Ok(Some(dial_address)) => {
+                socket.set_nodelay(nodelay)?;
```
dmitry-markin, line 205: Why we set `TCP_NODELAY` on a socket only if we bind to specific addr/port? Shouldn't it be set in all cases?

src/transport/websocket/mod.rs
```diff
@@ -248,6 +249,7 @@ impl WebSocketTransport {
 
         match dial_addresses.local_dial_address(&remote_address.ip()) {
             Ok(Some(dial_address)) => {
+                socket.set_nodelay(nodelay)?;
```
dmitry-markin, line 252: The same as with TCP: why only if we bind to specific endpoint?

dmitry-markin (approved):
Looks good! Thank you @altonen!

src/transport/tcp/mod.rs
altonen: Oversight on my part, good catch. Fixed

lexnv (approved):
LGTM!

lexnv:
@altonen Thanks again for contributing! Could you sign the commits to unblock merging? 🙏

altonen:
> @altonen Thanks again for contributing! Could you sign the commits to unblock merging? 🙏

I will try later today, probably requires a force push. I'm curious, why external contributors are required to sign commits?

Issue #70, opened 2024-03-29
transport: Extract common transport code between TCP and WebSocket


lexnv:
Make the transport layer a bit more module and allow the TCP transport code to be reused for WebSocket purposes. 

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/pull/147
- [ ] https://github.com/paritytech/litep2p/pull/151
``` 

cc https://github.com/paritytech/litep2p/pull/69 @paritytech/networking

PR #147, opened 2024-06-11
transport: Introduce common listener for tcp and websocket

lexnv:
This PR aims to remove duplicated code for socket listener between TCP and WebSocket transports.

Changes done:
- Extended code logic to facilitate both TCP and WebSockets
- Moved the listener code to a dedicated `common` module
- Provided fairness for polling the listener stream by introducing a rotating index
- Traits to convert between multiaddr <-> socket addr
- Removed unnecessary cloning of listen addresses 

Part of https://github.com/paritytech/litep2p/issues/70.


src/transport/common/listener.rs
```diff
@@ -100,17 +91,34 @@ impl DialAddresses {
...
+    /// Listener for WebSocket.
+    WebSocket,
+}
+
+impl SocketListener {
+    /// Create new [`SocketListener`]
     pub fn new(
         addresses: Vec<Multiaddr>,
         reuse_port: bool,
+        ty: SocketListenerType,
```
dmitry-markin, line 116: nit: may be use the full `type` naming? `ty` personally associates with "thank you", and not with type :D

src/transport/common/listener.rs
```diff
@@ -194,10 +202,15 @@ impl WebSocketListener {
...
+                let multi_addr = Multiaddr::empty()
                     .with(Protocol::from(address.ip()))
-                    .with(Protocol::Tcp(address.port()))
-                    .with(Protocol::Ws(std::borrow::Cow::Owned("/".to_string())))
+                    .with(Protocol::Tcp(address.port()));
+
+                match ty {
+                    SocketListenerType::Tcp => multi_addr,
+                    SocketListenerType::WebSocket =>
+                        multi_addr.with(Protocol::Ws(std::borrow::Cow::Owned("/".to_string()))),
```
dmitry-markin, line 212: nit: any reason to not use `Cow::Borrowed("/")`?

src/transport/common/listener.rs
```diff
@@ -293,21 +319,29 @@ impl WebSocketListener {
...
+                Poll::Ready(Err(error)) => {
+                    self.poll_index = (self.poll_index + 1) % len;
+                    return Poll::Ready(Some(Err(error)));
+                }
+                Poll::Ready(Ok((stream, address))) => {
+                    self.poll_index = (self.poll_index + 1) % len;
+                    return Poll::Ready(Some(Ok((stream, address))));
+                }
             }
         }
```
dmitry-markin, line 347: Why not advance the index if all the listeners have returned `Pending`?

dmitry-markin (approved):
Looks good!

I would, may be, go one step further and make `SocketListener` completely unaware of whether it deals with TCP or WebSocket. Basically, the main difference is the code path in `get_socket_address()`, and may be this can be made external to `SocketListener`. The PR already simplifies a lot and is good in it's current form, so we can do this in a follow-up if you consider this valuable/doable.

src/transport/common/listener.rs
lexnv: Have removed this from the public api. `type` must be escaped to not be confused by `type X = ..` by the compiler :D

src/transport/common/listener.rs
```diff
@@ -100,18 +91,91 @@ impl DialAddresses {
...
+}
+
+/// TCP helper to convert between `Multiaddr` and `SocketAddr`.
+pub struct TcpAddress;
+
+impl GetSocketAddr for TcpAddress {
+    fn multiaddr_to_socket_address(
+        address: &Multiaddr,
+    ) -> crate::Result<(AddressType, Option<PeerId>)> {
+        SocketListener::get_socket_address(address, SocketListenerType::Tcp)
```
dmitry-markin, line 140: What about moving the corresponding branch from `SocketListener::get_socket_address` here, and getting rid of `SocketListener::get_socket_address` completely, so `SocketListener` does not deal with the conversions of addresses at all? IMO, the logic of specific miltiaddress conversion doesn't belong to `SocketListener`.

lexnv: I see what you mean now, yep that makes total sense! Have moved it to a separate wrapper fn. Let me know if the trait approach from the API perspective looks good to you, otherwise we can fallback to using the inner `ListenerType` enum, thanks! 🙏

dmitry-markin: The trait approach is perfectly fine!

dmitry-markin (approved):
Everything looks good!

PR #149, opened 2024-06-12
transport: Fix erroneous handling of secondary connections

altonen:
Previously `TransportManager` didn't correctly handle the case where local node's dialing connection was opened as a secondary connection, leaving `dial_record` as `Some` which didn't get cleaned up after the connections were closed.

This resulted in subsequent dials to fail because `dial_record` was `Some`, which erroneously indicated that the remote was already being dialed, even though the dial had concluded and the connection was closed

---
Logs would show the following and the substream wouldn't get opened:
```
TRACE litep2p::notification: open substream peer=PeerId("12D3KooWFFoGy3aN6Mrvp73kZGjXqXu1SQ8W9vGzRbFu1Fu5N54d") protocol=/notif/1
DEBUG litep2p::transport-manager: peer is aready being dialed peer=PeerId("12D3KooWFFoGy3aN6Mrvp73kZGjXqXu1SQ8W9vGzRbFu1Fu5N54d") dial_record=Some(AddressRecord { score: 100, address: "/ip4/127.0.0.1/tcp/34765/p2p/12D3KooWFFoGy3aN6Mrvp73kZGjXqXu1SQ8W9vGzRbFu1Fu5N54d", connection_id: Some(ConnectionId(0)) })
TRACE litep2p::notification: started to dial peer peer=PeerId("12D3KooWFFoGy3aN6Mrvp73kZGjXqXu1SQ8W9vGzRbFu1Fu5N54d") protocol=/notif/1
TRACE litep2p::notification: open substream peer=PeerId("12D3KooWBL4ZFU5hehhs3JbmAYY7ZsAVp6ev1wfpUjT9SpA5hJUg") protocol=/notif/1
DEBUG litep2p::transport-manager: peer is aready being dialed peer=PeerId("12D3KooWBL4ZFU5hehhs3JbmAYY7ZsAVp6ev1wfpUjT9SpA5hJUg") dial_record=Some(AddressRecord { score: 100, address: "/ip4/127.0.0.1/tcp/41441/p2p/12D3KooWBL4ZFU5hehhs3JbmAYY7ZsAVp6ev1wfpUjT9SpA5hJUg", connection_id: Some(ConnectionId(0)) })
TRACE litep2p::notification: started to dial peer peer=PeerId("12D3KooWBL4ZFU5hehhs3JbmAYY7ZsAVp6ev1wfpUjT9SpA5hJUg") protocol=/notif/1
```

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -914,7 +914,10 @@ impl TransportManager {
         let mut peers = self.peers.write();
         match peers.get_mut(&peer) {
             Some(context) => match context.state {
-                PeerState::Connected { .. } => match context.secondary_connection {
+                PeerState::Connected {
+                    ref mut dial_record,
+                    ..
+                } => match context.secondary_connection {
                     Some(_) => {
                         tracing::debug!(
                             target: LOG_TARGET,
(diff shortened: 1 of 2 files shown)


src/transport/manager/mod.rs
```diff
@@ -939,22 +942,50 @@ impl TransportManager {
...
+                                ?peer,
+                                connection_id = ?endpoint.connection_id(),
+                                address = ?endpoint.address(),
+                                "secondary connection",
+                            );
+
+                            context.secondary_connection = Some(AddressRecord::new(
+                                &peer,
+                                endpoint.address().clone(),
+                                SCORE_DIAL_SUCCESS,
```
dmitry-markin, line 976: Is it actually an incoming connection in this case, as `dial_record` is `None`?
`SCORE_*DIAL*_SUCCESS` confuses me a little here.

tests/connection/mod.rs
```diff
@@ -1352,3 +1355,186 @@ async fn unspecified_listen_address_websocket() {
...
+            },
+            event = handle2.next() => match event.unwrap() {
+                NotificationEvent::NotificationStreamClosed { .. } => {
+                    peer2_close = true;
+                }
+                _ => {},
+            }
+        }
+    }
+    tokio::time::sleep(std::time::Duration::from_secs(15)).await;
```
dmitry-markin, line 1509: Can we wait for a peer disconnected event instead, or this would complicate the test? Just not to wait for 15 secs.

dmitry-markin (approved):
LGTM! Thank you!

tests/connection/mod.rs
altonen: I simplified the test to use `Litep2p::dial()` + polling for `Litep2pEvent::ConnectionClosed`. Worth nothing that this issue only happens with TCP & WebSocket transport but not with QUIC

src/transport/manager/mod.rs
altonen: It is. AFAIR there were at least one other `SCORE_SOMETHING_SUCCESS` but it got removed for reasons I can't remember. Its name should probably be changed to something else

dmitry-markin (approved):

src/transport/manager/mod.rs
altonen: I renamed both of them to `SCORE_CONNECT_*` for consistence, I can revert the other rename if you don't like it.

dmitry-markin: It's fine, thank you!

lexnv (approved):
LGTM! Nice catch!

PR #150, opened 2024-06-13
[dnm] ci: Simplify / reduce test times

lexnv:
Compares times with / without rust cache


lexnv:
tldr; removing the cache does not improve the time execution of our tests in the CI machine. From this one run, it seems that other steps are affected by increased time when rust-cache is not present. 

Closing this PR for now, we'll continue to investigate and try to simplify the CI a bit later

PR #151, opened 2024-06-13
transport/common: Share DNS lookups between TCP and WebSocket

lexnv:
Move the dns resolving to a dedicated common module and return specific errors on failures. 

Part of: https://github.com/paritytech/litep2p/issues/70

### DIFF
--- a/src/transport/common/listener.rs
+++ b/src/transport/common/listener.rs
@@ -59,7 +59,7 @@ pub enum AddressType {
 
 impl AddressType {
     /// Resolve the address to a concrete IP.
-    async fn lookup_ip(self) -> crate::Result<SocketAddr> {
+    pub async fn lookup_ip(self) -> crate::Result<SocketAddr> {
         let (url, port, is_dns6) = match self {
             // We already have the IP address.
             AddressType::Socket(address) => return Ok(address),
--- a/src/transport/common/listener.rs
+++ b/src/transport/common/listener.rs
@@ -27,6 +27,10 @@ use multiaddr::{Multiaddr, Protocol};
 use network_interface::{Addr, NetworkInterface, NetworkInterfaceConfig};
 use socket2::{Domain, Socket, Type};
 use tokio::net::{TcpListener as TokioTcpListener, TcpStream};
+use trust_dns_resolver::{
+    config::{ResolverConfig, ResolverOpts},
+    TokioAsyncResolver,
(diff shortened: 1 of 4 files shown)


src/transport/common/listener.rs
```diff
@@ -311,10 +367,27 @@ fn multiaddr_to_socket_address(
...
         },
-        Some(Protocol::Dns(address))
-        | Some(Protocol::Dns4(address))
-        | Some(Protocol::Dns6(address)) => match iter.next() {
-            Some(Protocol::Tcp(port)) => AddressType::Dns(address.to_string(), port),
+        Some(Protocol::Dns(address)) | Some(Protocol::Dns4(address)) => match iter.next() {
+            Some(Protocol::Tcp(port)) => AddressType::Dns {
+                address: address.into(),
+                port,
+                is_dns6: false,
```
dmitry-markin, line 374: As [multiaddr spec](`https://github.com/multiformats/multiaddr/blob/master/protocols.csv`) says, `Protocol::Dns` can resolve to both IPv4 & IPv6. I would use this PR as an opportunity to fix the existing behavior and allow IPv6 addresses with `Protocol::Dns`.

dmitry-markin (approved):
LGTM! :+1:

src/transport/common/listener.rs
lexnv: Nice catch! For sure, will fix it right away

src/transport/tcp/connection.rs
```diff
@@ -455,9 +459,21 @@ impl TcpConnection {
...
+                DnsType::Dns => Multiaddr::empty()
+                    .with(Protocol::Dns(Cow::Owned(address)))
+                    .with(Protocol::Tcp(port)),
+                DnsType::Dns4 => Multiaddr::empty()
+                    .with(Protocol::Dns4(Cow::Owned(address)))
+                    .with(Protocol::Tcp(port)),
+                DnsType::Dns6 => Multiaddr::empty()
+                    .with(Protocol::Dns6(Cow::Owned(address)))
+                    .with(Protocol::Tcp(port)),
+            },
```
lexnv, line 476: Here we had another miss-match, where regardless of the initial DNS version we'd accept both ipv4 or ipv6

dmitry-markin (approved):
Nice!

Issue #129, opened 2024-05-28
kad: Ensure Kademlia Record has TTL


lexnv:
If the record received by `kad::put_value` does not have a TTL, use the default 36h expiry.

https://github.com/paritytech/litep2p/blob/467a457c78785819e91ecdb7a5a926b7cd1c3d25/src/protocol/libp2p/kademlia/record.rs#L88-L89

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/pull/154
- [ ] https://github.com/paritytech/litep2p/pull/162
```

PR #154, opened 2024-06-14
kad: Set default ttl 36h for kad records

lexnv:
This PR introduces a default TTL for records.
- the default is set to 36h if the record ttl is unspecified
- this ensures the local memory store is not polluted with records without TTL

Part of: https://github.com/paritytech/litep2p/issues/129


dmitry-markin (approved):
LGTM! Good catch!

Issue #153, opened 2024-06-14
How to keep connection while waiting for Response


Ma233:
Hello, I am working on a project that using litep2p with its RequestResponse protocol.

I found that the connection will be closed if a response is not sent within 5 seconds.

I have tested both TcpTransport and QuicTransport, and they are all like this.

I am uncertain whether this relates to the issue https://github.com/paritytech/litep2p/issues/131.

Is there a correct way to keep an open connection until it respond a response? It could be manually closed if neccessary.

dmitry-markin:
Hi, just to double check, did you try configuring the request-response protocol with a custom timeout?
https://github.com/paritytech/litep2p/blob/ebb5f2dc89a65b7b4c67b592c9312f874be4c0fc/src/protocol/request_response/config.rs#L143

It's set by default to 5 seconds.

Ma233:
@dmitry-markin Hi, thanks for your reminder, I have now added this configuration. But there are still disconnections.

After setting the log level to trace, I found a "keep-alive timeout exceeded, downgrading connection" message. I believe this could be the reason.

Ma233:
I temporarily bypassed this timeout by introducing a ping protocol in the project.

dmitry-markin:
Yeah, I see, the connection is downgraded if no substream is opened within 5 seconds after creating the connection. This is strange, because sending a request over request-response protocol opens a substream.

Are you sure you send a request within 5 seconds of opening the connection? Do you know you can let litep2p establish a connection for you when sending a request using [`DialOptions::Dial`](https://github.com/paritytech/litep2p/blob/ebb5f2dc89a65b7b4c67b592c9312f874be4c0fc/src/protocol/request_response/handle.rs#L202)?

Ma233:
And here is the logs of request side:
```
2024-06-14T14:52:30.439860Z DEBUG server_handshake:FramedWrite::buffer{frame=Settings { flags: (0x0), initial_window_size: 1048576, max_frame_size: 16384, max_header_list_size: 16777216 }}: h2::codec::framed_write: send frame=Settings { flags: (0x0), initial_window_size: 1048576, max_frame_size: 16384, max_header_list_size: 16777216 }
2024-06-14T14:52:30.439913Z TRACE server_handshake:FramedWrite::buffer{frame=Settings { flags: (0x0), initial_window_size: 1048576, max_frame_size: 16384, max_header_list_size: 16777216 }}: h2::frame::settings: encoding SETTINGS; len=18
[...]

PR #155, opened 2024-06-15
transport: Allow manual setting of keep-alive timeout

Ma233:
Allow configure time duration of keep_alive_timeouts in `TransportService`.
For `RequestResponseProtocol`, use its timeout configuration.
For other protocols, use the original default 5 seconds.

Related issue: #153

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -30,7 +30,7 @@ use crate::{
     },
     transport::{
         manager::limits::ConnectionLimitsConfig, tcp::config::Config as TcpConfig,
-        MAX_PARALLEL_DIALS,
+        KEEP_ALIVE_TIMEOUT, MAX_PARALLEL_DIALS,
     },
     types::protocol::ProtocolName,
     PeerId,
@@ -156,7 +156,7 @@ impl ConfigBuilder {
             request_response_protocols: HashMap::new(),
             known_addresses: Vec::new(),
             connection_limits: ConnectionLimitsConfig::default(),
-            keep_alive_timeout: Duration::from_secs(5),
+            keep_alive_timeout: KEEP_ALIVE_TIMEOUT,
         }
     }
 
@@ -273,7 +273,7 @@ impl ConfigBuilder {
     }
 
     /// Set keep alive timeout for connections.
-    pub fn set_keep_alive_timeout(mut self, timeout: Duration) -> Self {
+    pub fn with_keep_alive_timeout(mut self, timeout: Duration) -> Self {
         self.keep_alive_timeout = timeout;
         self
     }
--- a/src/config.rs
+++ b/src/config.rs
@@ -45,7 +45,7 @@ use crate::transport::websocket::config::Config as WebSocketConfig;
 
 use multiaddr::Multiaddr;
 
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
 /// Connection role.
 #[derive(Debug, Copy, Clone)]
@@ -121,6 +121,9 @@ pub struct ConfigBuilder {
 
     /// Connection limits config.
     connection_limits: ConnectionLimitsConfig,
+
+    /// Close the connection if no substreams are open within this time frame.
+    keep_alive_timeout: Duration,
 }
 
 impl Default for ConfigBuilder {
@@ -153,6 +156,7 @@ impl ConfigBuilder {
             request_response_protocols: HashMap::new(),
             known_addresses: Vec::new(),
             connection_limits: ConnectionLimitsConfig::default(),
+            keep_alive_timeout: Duration::from_secs(5),
         }
     }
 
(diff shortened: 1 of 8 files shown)


lexnv:
Thanks for contributing! Generally, the approach looks good to me, I'll have another look over the PR soon! 

Its always good to have the ability to configure the parameters to make litep2p suitable for more usecases. 
I'm wondering, does this approach solve your issue from: https://github.com/paritytech/litep2p/issues/153? (ie being able to configure the keep-alive)

Ma233:
@lexnv Hi, yes it solved the issue perfectly. Thank you for your support.

dmitry-markin:
I would add one global `keep_alive_timeout` for all protocols to [`LiteP2pConfig`](https://github.com/paritytech/litep2p/blob/fd02c1feaf993ccd3e4ab04e9cd81beae5793c79/src/config.rs#L275) and a corresponding `with_keep_alive_timeout()` to [`ConfigBuilder`](https://github.com/paritytech/litep2p/blob/fd02c1feaf993ccd3e4ab04e9cd81beae5793c79/src/config.rs#L120), and pass this directly to `TransportManager` upon construction. This way it would allow configuring timeouts for all protocols, and eliminate the need for hardcoding `Duration::from_secs(5)` in many places.

Otherwise looks good, I think we should add this to litep2p.

jasl:
Kindly ping.
Do you think this PR is merge-able?

dmitry-markin:
> Kindly ping. Do you think this PR is merge-able?

Yes, but it should be done as described in my previous comment.

dmitry-markin (changes requested):
See https://github.com/paritytech/litep2p/pull/155#issuecomment-2191061683.

lexnv:
Thanks a lot for contributing again! 

I do like the suggestion from Dmitry, I think we can go that way here.

Please let us know if you have some capacity to implement it, otherwise we can tackle it and include it in our next release (hopefully in the upcoming week) 🙏 

// cc @Ma233 @jasl @dmitry-markin let me know if you'd prefer to jump in on this and help with the implementation. We are moving as fast as we safely can and there are a few pending things in flight at the moment, any help is greatly appreciated here 🙏

Ma233:
@dmitry-markin @lexnv Sorry for taking so long to reply to you. I have added a `keep_alive_timeout` option in `Litep2pConfig`.

Ma233:
And the `RequestResponseProtocol` does not use its timeout configuration now. It use the `keep_alive_timeout` option as other protocols.

src/config.rs
```diff
@@ -268,6 +272,12 @@ impl ConfigBuilder {
         self
     }
 
+    /// Set keep alive timeout for connections.
+    pub fn set_keep_alive_timeout(mut self, timeout: Duration) -> Self {
```
dmitry-markin, line 276:
```suggestion
    pub fn with_keep_alive_timeout(mut self, timeout: Duration) -> Self {
```

src/config.rs
```diff
@@ -153,6 +156,7 @@ impl ConfigBuilder {
             request_response_protocols: HashMap::new(),
             known_addresses: Vec::new(),
             connection_limits: ConnectionLimitsConfig::default(),
+            keep_alive_timeout: Duration::from_secs(5),
```
dmitry-markin, line 159: Could you put this constant into [`transport/mod.rs`](https://github.com/paritytech/litep2p/blob/aa18af8fd4d2209e849308df08b4c792e6a468db/src/transport/mod.rs#L44-L48) with other timeouts, and use it (including tests) instead of hardcoding `Duration::from_secs(5)`, please?

src/protocol/transport_service.rs
```diff
@@ -122,6 +122,9 @@ pub struct TransportService {
     /// Next substream ID.
     next_substream_id: Arc<AtomicUsize>,
 
+    /// Close the connection if no substreams are open within this time frame.
+    keep_alive: Duration,
```
dmitry-markin, line 126: nit: You probably tried to separate this from `keep_alive_timeouts`, but IMO it would be better to use the `_timeout` suffix here as well.

src/transport/manager/mod.rs
```diff
@@ -322,6 +323,7 @@ impl TransportManager {
         protocol: ProtocolName,
         fallback_names: Vec<ProtocolName>,
         codec: ProtocolCodec,
+        keep_alive: Duration,
```
dmitry-markin, line 326:
```suggestion
        keep_alive_timeout: Duration,
```

src/protocol/transport_service.rs
```diff
@@ -134,6 +137,7 @@ impl TransportService {
         fallback_names: Vec<ProtocolName>,
         next_substream_id: Arc<AtomicUsize>,
         transport_handle: TransportManagerHandle,
+        keep_alive: Duration,
```
dmitry-markin, line 140:
```suggestion
        keep_alive_timeout: Duration,
```

dmitry-markin (approved):

dmitry-markin:
@Ma233 could you not force push, as it complicates the review, please?

Ma233:
@dmitry-markin Hello, I have consolidated the changes into a single commit:
1. Add a KEEP_ALIVE_CONST and replace all hard coded time duration with it.
2. Change the field of `keep_alive_timeouts` to `pending_keep_alive_timeouts` to avoid confusion.

dmitry-markin:
Thanks, I will have a look. Please don't force-push to public branches, as I now have to go through all the changes again, instead of just checking the latest commit.

lexnv (approved):
LGTM! Thanks again for contributing here 🙏 

We'll do a release shortly, probably next week or so which would include this PR as well! 👍

PR #162, opened 2024-06-24
kademlia: Preserve publisher & expiration time in DHT records

dmitry-markin:
This PR fixes a bug with publisher & expiration time not being preserved in DHT records.

Resolves https://github.com/paritytech/litep2p/issues/129.


lexnv (approved):
LGTM! Nice catch 👍

Issue #171, opened 2024-07-08
kad: Add kademlia metrics


lexnv:
Expose the following info for reporting back metrics:
- kademlia total number of records store in the memory store
- kademlia total number of bytes stored in the memory store (sum of all record's value)
- kademlia entries per bucket
  - For simplicity, this can be a Vec<(u32, size)> similar to [libp2p](https://github.com/paritytech/polkadot-sdk/blob/9afb15a31f791973878c93f398f55e17a9935901/substrate/client/network/src/discovery.rs#L464-L474), where `u32` is the base log2 of the bucket lower bound and `usize` is the number of peers in that bucket

lexnv:
We could also add here any missing "pending" peers from the the kademlia queries:
https://github.com/paritytech/litep2p/blob/66aaf9dbdf60a3b3bbfda2334e2e1887af3c2315/src/protocol/libp2p/kademlia/query/find_node.rs#L119

PR #174, opened 2024-07-09
kad: Add query id to log messages

lexnv:
This tiny PR adds the query ID to some kademlia log messages from the `FIND_NODE` and `GET_RECORDS` commands. 

Used to provide better debug information and upgrade a few missing states from trace to warn. 

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -399,6 +399,7 @@ impl Kademlia {
                             target: LOG_TARGET,
                             ?peer,
                             ?target,
+                            ?query_id,
                             "handle `FIND_NODE` response",
                         );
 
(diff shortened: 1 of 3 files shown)


skunert (approved):
Looks good. However just to make sure, these warn messages should not appear during normal operation right? Just to make sure that they will not spam the logs.

lexnv:
> Looks good. However just to make sure, these warn messages should not appear during normal operation right? Just to make sure that they will not spam the logs.


Initially wanted to expose the info better to our triaging. However, this is not something critical atm and we could probably better add this as a metric in the future:
https://github.com/paritytech/litep2p/issues/171#issuecomment-2222884722

Have downgraded this to debug, thanks!

dmitry-markin (approved):

Issue #172, opened 2024-07-09
transport-manager: Unknown connection opened


lexnv:
Investigate the following errors: 

```bash
    "2024-07-06 04:53:54.449  WARN tokio-runtime-worker litep2p::transport-manager: unknown connection opened as secondary connection, discarding peer=PeerId(\"12D3KooWB9DVNyi6TQUhHeyq4LyBLVykybXUmcNmEUD9tcKEaYq1\") connection_id=ConnectionId(246958) address=\"/ip6/*/tcp/51326/ws/p2p/12D3KooWB9DVNyi6TQUhHeyq4LyBLVykybXUmcNmEUD9tcKEaYq1\" dial_record=AddressRecord { score: 100, address: \"/ip4/1*/tcp/31334/ws/p2p/12D3KooWB9DVNyi6TQUhHeyq4LyBLVykybXUmcNmEUD9tcKEaYq1\", connection_id: Some(ConnectionId(246946)) }",
    "2024-07-06 04:53:54.450  WARN tokio-runtime-worker litep2p::transport-manager: non-primary connection was closed but secondary connection doesn't exist peer=PeerId(\"12D3KooWB9DVNyi6TQUhHeyq4LyBLVykybXUmcNmEUD9tcKEaYq1\") connection_id=ConnectionId(246958)",
    "2024-07-06 04:53:54.450 ERROR tokio-runtime-worker litep2p::transport-manager: failed to handle closed connection error=InvalidState",
    "2024-07-06 04:54:04.058  WARN tokio-runtime-worker litep2p::transport-manager: non-primary connection was closed but secondary connection doesn't exist peer=PeerId(\"12D3KooWB9DVNyi6TQUhHeyq4LyBLVykybXUmcNmEUD9tcKEaYq1\") connection_id=ConnectionId(246953)",
    "2024-07-06 04:54:04.058 ERROR tokio-runtime-worker litep2p::transport-manager: failed to handle closed connection error=InvalidState",
```

Data collected from a long running litep2p substrate node in kusama

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/pull/176
- [ ] https://github.com/paritytech/litep2p/pull/179
```

altonen:
The issue could be that after receiving an unknown unknown connection, `TransportManager` wants to discard it: https://github.com/paritytech/litep2p/blob/a27d0072913be67cce904e2e8c465cc37ec5223a/src/transport/manager/mod.rs#L980-L987

[...]

PR #176, opened 2024-07-09
transport: Reject secondary connections with different connection IDs

lexnv:
This PR ensures that the transport manager rejects secondary connections.

When the secondary connection is rejected, the internal `dial_record` is preserved.

Added a test to validate this behavior, with multiple wrong endpoints.

Part of: https://github.com/paritytech/litep2p/issues/172
Will keep the issue opened for to further into it. 

cc @paritytech/networking

### DIFF
--- a/src/transport/manager/address.rs
+++ b/src/transport/manager/address.rs
@@ -34,7 +34,7 @@ pub struct AddressRecord {
     /// Address.
     address: Multiaddr,
 
-    /// Connection ID, if specifed.
+    /// Connection ID, if specified.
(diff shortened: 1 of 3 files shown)


skunert (approved):

lexnv:
Have added a repro case in this PR as well. More details in the high-level fix: https://github.com/paritytech/litep2p/pull/179, which builds on top of this one :D

dmitry-markin (approved):

lexnv:
Forced push the merge origin/master commit to fix up testing

Issue #180, opened 2024-07-18
transport: Change `AddressStore` to keep track of more dial addresses


lexnv:
The address store keeps track of the addresses of peers, sorted internally by connection scores.
Once an address is inserted into the store, its score can never change.

We are currently losing track of dialed addresses if we already have a dial inflight:
- `fn dial_addresses(address: Multiaddr)` helps us learn about a peer multiaddress
- we completely discard the address if the peer is already `Dialing` `Connected` or `Opening`

Instead, we could keep track of the address:
- insert it into the store, if not already present with score 0 (zero)
- when we receive a successful / failure event -> update the reputation of the address (this is not present atm)
- this should help us lower the reputation of addresses that fail to respond over multiple attempts

A few other thoughts to look over:

- Here the store returns a sorted by score list of addresses. We collect them to a hashmap, which when iterated doesn't preserve the order of elements.
https://github.com/paritytech/litep2p/blob/443f05c4d041db42d0c00026783fb34f094c9c2b/src/transport/manager/mod.rs#L453-L457

- The following should be changed as it is redundant

https://github.com/paritytech/litep2p/blob/443f05c4d041db42d0c00026783fb34f094c9c2b/src/transport/manager/handle.rs#L194-L196

https://github.com/paritytech/litep2p/blob/443f05c4d041db42d0c00026783fb34f094c9c2b/src/transport/manager/address.rs#L204-L209

- Limit the number of known connections (maybe 16 would be a good start)

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/pull/250
```

lexnv:
Closed by https://github.com/paritytech/litep2p/pull/250

PR #184, opened 2024-07-23
kad: Update routing table on kademlia established connections

lexnv:
In this PR:
- Established connections from the kademlia component are reported immediately to the routing table iff the mode is in `Automatic`.
- The kademlia on connection established address is reported to the routing table for healthier tracking of addresses. This improves a bit the discoverability of peers, especially in cases where the `RoutingTableUpdate::Manual` option is set (ie Substrate)
- Refactor the `Kbucket::entry` function to iterate only once through the nodes, instead of twice
- Remove unneeded routing table ConnectionType's `CannotConnect` and `CanConnect`. The terminology comes from the kademlia peer status in low-level commands. However, the routing table only needs to know if the peer is connected or disconnected (similar to libp2p).

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/bucket.rs
+++ b/src/protocol/libp2p/kademlia/bucket.rs
@@ -94,7 +94,7 @@ impl KBucket {
 
         for i in 0..self.nodes.len() {
             match self.nodes[i].connection {
-                ConnectionType::NotConnected | ConnectionType::CannotConnect => {
+                ConnectionType::NotConnected => {
                     return KBucketEntry::Vacant(&mut self.nodes[i]);
                 }
                 _ => continue,
--- a/src/protocol/libp2p/kademlia/bucket.rs
+++ b/src/protocol/libp2p/kademlia/bucket.rs
@@ -68,17 +68,19 @@ impl KBucket {
         }
     }
 
-    /// Get the index position of the key in the k-bucket.
-    fn get_index<K: Clone>(&self, key: &Key<K>) -> Option<usize> {
-        self.nodes.iter().position(|p| p.key.as_ref() == key.as_ref())
-    }
-
     /// Get entry into the bucket.
-    // TODO: this is horrible code
     pub fn entry<K: Clone>(&mut self, key: Key<K>) -> KBucketEntry<'_> {
-        // Return the entry if present.
-        if let Some(index) = self.get_index(&key) {
-            return KBucketEntry::Occupied(&mut self.nodes[index]);
+        let mut replace_candidate = None;
+        for (index, node) in self.nodes.iter().enumerate() {
+            // If the node is already present in the k-bucket, return it.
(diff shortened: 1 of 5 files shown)


dmitry-markin:
> * The kademlia on connection established address is reported to the routing table for healthier tracking of addresses. This improves a bit the discoverability of peers, especially in cases where the `RoutingTableUpdate::Manual` option is set (ie Substrate)

I was thinking about this at some point, but ended up deciding that it can make things worse for kademlia, as the endpoint addresses are not necessarily the reachable ones. I think the current implementation of reporting back the observed addresses by `Identify` protocol and adding to the routing table only ones that survive the "externality" check using `add_self_reported_address` is fine.

dmitry-markin (commented):
The entry lookup optimization in `bucket.rs` looks good, but I don't think we should add endpoint addresses as is to the routing table. At least not in the `Manual` update mode.

As for `Automatic` routing table update mode, I actually wonder how a new node joining DHT should end up in the other nodes' routing tables. Current kademlia implementation spreads routing table entries to other peers via `FIND_NODE` and `GET_VALUE` responses, but we have a chicken and egg problem with new nodes joining the network. Correct me if I'm wrong, but it looks like the current `Automatic` implementation is going to have issues with connectivity (the network will never learn the address of a new-joiner). And it looks like this PR fixes it (even though it might not work good enough, as the endpoint address can have an ephemeral port, or can suffer NAT translation not suitable for incoming TCP connections).

lexnv:
I see what you mean now, yep that makes sense! Thanks for the info 🙏 

Thinking out loud, we can probably close the gap and make things a bit more resilient with:
- https://github.com/paritytech/litep2p/issues/180
- https://github.com/paritytech/litep2p/issues/191

The first issue tracks addresses a bit more robustly from the transport manager perspective, without necessarily loosing track of dialed addresses, or potential addresses to dial in the future.

After reading your message and looking back over the issue of authority-discovery not finding external addresses, I realized that we keep a static list of "listen addrs" in the Idenitfy protocol. The second issue should close the gap and provide proper information back to the remote peer.

All this relies on the fact that we report peer addresses via `add_self_reported_address` that is called on the `Discovery::Identified` event. If we provide outdated, or very few addresses in the Identify response, then the other peer won't have good information to dial us in the future. And instead, the routing table of the remote peer will contain outdated (possibly) expired information about our addresses.

We are also populating the routing table on automatic below, and reporting the address to the transport manager:
https://github.com/paritytech/litep2p/blob/2d1a4b48bb7f0f768b825ce88d456fe741c922df/src/protocol/libp2p/kademlia/mod.rs#L367-L377

Making the transport manager a bit more robust in tracking addresses should help out.

Let me know if this sounds like a plan 🙏

dmitry-markin:
Very good findings. Reporting discovered external addresses in `Identify` messages should improve things significantly. I would start from implementing it.

PR #190, opened 2024-07-31
kad: Do not update memory store on incoming `GetRecordSuccess`

lexnv:
This PR ensures that the records discovered by the `GetRecord` query are not propagated to the memory store.
This makes the litep2p behavior similar to the libp2p.

Another approach could be to put the record into the memory store if the validation mode is `Automatic`. However, for networks with a large number of records, this will exhaust the limit of our memory store.

Thanks @alexggh for catching this 🙏 

Fix for:
- https://github.com/paritytech/litep2p/pull/190

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -991,7 +991,7 @@ mod tests {
         let action = QueryAction::GetRecordQueryDone { query_id, records };
         assert!(kademlia.on_query_action(action).await.is_ok());
 
-        // Check the local storage should not get updated updated.
+        // Check the local storage should not get updated.
         assert!(kademlia.store.get(&key).is_none());
     }
 }
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -656,31 +656,6 @@ impl Kademlia {
                 Ok(())
             }
             QueryAction::GetRecordQueryDone { query_id, records } => {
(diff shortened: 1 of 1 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -976,9 +951,8 @@ mod tests {
         let action = QueryAction::GetRecordQueryDone { query_id, records };
         assert!(kademlia.on_query_action(action).await.is_ok());
 
-        // Check the local storage was updated.
-        let record = kademlia.store.get(&key).unwrap();
-        assert_eq!(record.value, vec![0x1]);
+        // Check the local storage should not get updated updated.
```
dmitry-markin, line 954:
```suggestion
        // Check the local storage should not get updated.
```

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -1017,8 +991,7 @@ mod tests {
         let action = QueryAction::GetRecordQueryDone { query_id, records };
         assert!(kademlia.on_query_action(action).await.is_ok());
 
-        // Check the local storage was updated.
-        let record = kademlia.store.get(&key).unwrap();
-        assert_eq!(record.value, vec![0x2]);
+        // Check the local storage should not get updated updated.
```
dmitry-markin, line 994:
```suggestion
        // Check the local storage should not get updated.
```

dmitry-markin (approved):

alexggh (approved):

Issue #161, opened 2024-06-21
Make features optional


nazar-pc:
Right now due to str0m litep2p pulls OpenSSL, which is very undesirable due to how annoying it is to deal with.

Please consider making a set of optional features that selectively enable certain protocols. For example I'd like to disable WebRTC protocol because it is not used in Substrate anyway and makes me deal with OpenSSL with literally no benefit whatsoever. And annoyingly litep2p is a mandatory dependency of Substrate now, so I can't get rid of it easily either.

Something like libp2p-core and stuff would be much preferred over this monolith, but at very least having features that can be disabled would be very much appreciated.

dmitry-markin:
This can be done, of course. However, one thing to consider is that WebRTC is going to be enabled by default in substrate as soon as this transport is stable in litep2p, as this is a long awaited feature needed by https://github.com/smol-dot/smoldot.

nazar-pc:
RocksDB is also enabled by default, but it can be disabled by specifying `default-features = false`, so I hope the same could be done here as well. Most of Substrate seems to only need a few types, so support for specific transports can be made behind features.

PR #192, opened 2024-08-02
Make transports optional

dmitry-markin:
Put extra transports behind feature flags to make some dependencies optional.

While at it, also made sure no warnings are emitted on the recent rust version.

Resolves https://github.com/paritytech/litep2p/issues/161.

### DIFF
--- a/src/transport/manager/handle.rs
+++ b/src/transport/manager/handle.rs
@@ -121,16 +121,18 @@ impl TransportManagerHandle {
 
         match iter.next() {
             None => false,
-            Some(Protocol::Tcp(_)) => match (
-                iter.next(),
-                self.supported_transport.contains(&SupportedTransport::WebSocket),
-            ) {
-                (Some(Protocol::Ws(_)), true) => true,
-                (Some(Protocol::Wss(_)), true) => true,
-                (Some(Protocol::P2p(_)), _) =>
+            Some(Protocol::Tcp(_)) => match iter.next() {
+                Some(Protocol::P2p(_)) =>
                     self.supported_transport.contains(&SupportedTransport::Tcp),
+                #[cfg(feature = "websocket")]
+                Some(Protocol::Ws(_)) =>
+                    self.supported_transport.contains(&SupportedTransport::WebSocket),
+                #[cfg(feature = "websocket")]
+                Some(Protocol::Wss(_)) =>
+                    self.supported_transport.contains(&SupportedTransport::WebSocket),
                 _ => false,
             },
+            #[cfg(feature = "quic")]
             Some(Protocol::Udp(_)) => match (
                 iter.next(),
                 self.supported_transport.contains(&SupportedTransport::Quic),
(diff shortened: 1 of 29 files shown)


src/transport/manager/handle.rs
```diff
@@ -285,7 +287,7 @@ pub struct TransportHandle {
     pub protocols: HashMap<ProtocolName, ProtocolContext>,
     pub next_connection_id: Arc<AtomicUsize>,
     pub next_substream_id: Arc<AtomicUsize>,
-    pub protocol_names: Vec<ProtocolName>,
+    pub _protocol_names: Vec<ProtocolName>,
```
lexnv, line 290: dq: Do we plan to use this in the future?

tests/connection/mod.rs
```diff
@@ -21,30 +21,46 @@
...
     protocol::libp2p::ping::{Config as PingConfig, PingEvent},
-    transport::{
-        quic::config::Config as QuicConfig, tcp::config::Config as TcpConfig,
-        websocket::config::Config as WebSocketConfig,
-    },
+    transport::tcp::config::Config as TcpConfig,
     Litep2p, Litep2pEvent, PeerId,
 };
 
+#[cfg(feature = "websocket")]
```
lexnv, line 30: nit: It might be worth having this module structured as:
- `mod.rs` to bridge the `quic.rs` `websocket.rs` and `tcp.rs` depending on feature flags; and here we can have the common logic as well
- then have the tests moved to individual files so that the only `cfg` we state is in mod.rs


```
#mod.rs

#[cfg(feature = "quic")]
use quic.rs
...
```

We can also handle this later since I expect to run into some edge-cases along the way which may involve a closer refactor

tests/protocol/identify.rs
```diff
@@ -26,19 +26,33 @@ use litep2p::{
...
 enum Transport {
-    Quic(QuicConfig),
     Tcp(TcpConfig),
+    #[cfg(feature = "quic")]
+    Quic(QuicConfig),
+    #[cfg(feature = "websocket")]
     WebSocket(WebSocketConfig),
 }
 
+fn add_transport(config: ConfigBuilder, transport: Transport) -> ConfigBuilder {
```
lexnv, line 46: nit: By looking at this fn, I believe we can extract some common bits into tests/common.rs? Can be done as a followup :D

lexnv (approved):
LGTM! Nice work here 👍 

I like that we guard dependencies by feature flags. This should be quite important for users that only desire individual transports. It adds a bit of complexity for us in managing feature-flags, and we can look into the future at structuring our testing in a friendlier way for us, although put in balance I believe this is good 🙏

tests/connection/mod.rs
dmitry-markin: I was thinking about putting quick/websocket/webrtc tests into dedicated files, but it looks like the tests only differ in the transport used, and belong better to the connection/protocol test files. I.e., it would be easier to have these tests along each other in the same file. I will think if there is another alternative to having `cfg` spread over the files. Let me know if you have anything in mind

tests/protocol/identify.rs
dmitry-markin: Done. While doing this I ran into issue that `tests/common.rs` is considered to be a test on it's own, and rust generates dead code warnings. So I ended up putting it in `tests/common/mod.rs`.

src/transport/manager/handle.rs
dmitry-markin: I think we can safely remove it for now (done).

nazar-pc:
Thank you, very appreciated!

Issue #186, opened 2024-07-26
transport/manager: Add ability to accept/reject incoming connections before negotiation 


lexnv:
Refactor the transport manager and every transport to allow the transport manager to accept or reject incoming connections.

Currently, each transport is responsible for polling the underlying socket and all incoming connections are immediately accepted and negotiated:

https://github.com/paritytech/litep2p/blob/4c7cda693a953548e1aaa594dbe6972a5fac18fa/src/transport/tcp/mod.rs#L421-L429

https://github.com/paritytech/litep2p/blob/4c7cda693a953548e1aaa594dbe6972a5fac18fa/src/transport/tcp/mod.rs#L113-L126

It would be beneficial in terms of resources to move the accept/reject responsibility to the transport manager for inbound connections (similar how we handle outbound ones). We can reject the connection immediately rather than negotiating it to simply reject later.

One option would be to extend the `TransportEvent::ConnectionOpened` to support an `Endpoint` instead of a connection + address. Then we can emit this event from each transport to the transport manager. Then, the transport manager can check the limits and handle the negotiation. This may also require a refactor to the peer state logic.

Issue #17, opened 2023-09-25
Implement connection limits


altonen:
```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/pull/185
- [ ] https://github.com/paritytech/litep2p/pull/194
```

lexnv:
Monitoring a litep2p node with a libp2p node, it is clear that litep2p can sustain many more connections.

This data point supports the connection limits, as I believe it will free up some more resources (ie not having to deal with that many connections):


![Screenshot 2024-07-09 at 14 18 46](https://github.com/paritytech/litep2p/assets/60601340/5eb545aa-ff1b-4318-a08e-1252ab162261)

altonen:
The connections also allocate quite a bit of memory, like ~600KB of memory per connection. Those numbers are configurable but it'd be better to have proper connection limits which could be as simple as just checking the total number of connections in `TransportManager` before accepting an inbound connection/starting to dial remote peer.

The more advanced option would be to not poll TCP/WebSocket listeners at all if the maximum number of connections has been reached but last time I looked into it, it required some non-trivial API rework in the transport interface.

If resource consumption is an issue, you can decrease the sizes of Noise read/write buffers in the transport configs.

PR #194, opened 2024-08-06
transport: Add accept_pending/reject_pending for inbound connections and introduce inbound limits

lexnv:
This PR refactors the transport trait to allow the transport manager to decide if the incoming connection should be negotiated or rejected immediately. 

Before this PR, the transports (TCP, ws) eagerly accepted inbound connections and started negotiating.
This consumed resources if the node was already at the limits or connected.

To achieve this, each transport was refactored to pass the responsibility of accepting connections to the transport manager.

After the refactor, we are able to look at the connection limits before accepting a new inbound connection. 

### Notes

This PR aims to introduce the methods and only looks at the connection limits.
In the future (soonish), we need to tackle separately:
- state machine refactoring to accept only 1 inbound connection per peer context
- accept/reject inbounds after looking at peer context (ie if already connected there's no benefit)
- s2n-quic transport is out of date and outside the scope of this PR

### Next Steps
- Look at the performance impact of communicating between transport manager <-> individual transports
- If the impact is significant (although under load I expect the channel communication of 2 messages to be faster than eagerly negotiating), then we can look into simplifying the events. However, this gives us a robust view of each transport, and allows us not to negotiate a connection if the peer is already connected.

Closes: https://github.com/paritytech/litep2p/issues/186
Part of: https://github.com/paritytech/litep2p/issues/17

cc @paritytech/networking

### DIFF
--- a/src/transport/websocket/mod.rs
+++ b/src/transport/websocket/mod.rs
@@ -154,7 +154,7 @@ impl WebSocketTransport {
         let address = Multiaddr::empty()
             .with(Protocol::from(address.ip()))
             .with(Protocol::Tcp(address.port()))
-            .with(Protocol::Ws(std::borrow::Cow::Owned("/".to_string())));
+            .with(Protocol::Ws(std::borrow::Cow::Borrowed("/")));
 
         self.pending_connections.push(Box::pin(async move {
             match tokio::time::timeout(connection_open_timeout, async move {
@@ -450,7 +450,6 @@ impl Transport for WebSocketTransport {
     }
 
     fn reject_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
-        self.canceled.insert(connection_id);
         self.pending_open
             .remove(&connection_id)
             .map_or(Err(Error::ConnectionDoesntExist(connection_id)), |_| Ok(()))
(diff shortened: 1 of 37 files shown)


lexnv:
Something must have changed in the latest clippy version, will look into fixing that in a separate PR, then come back to this and rebase on origin/master

lexnv:
Screenshot 2024-08-07 at 11 14 15

In the left side of the gap, we have a litep2p node running without this PR, on the right side the litep2p is running with this PR. The CPU performance looks similar between nodes. (magenta is libp2p)

src/transport/websocket/mod.rs
```diff
@@ -385,6 +438,24 @@ impl Transport for WebSocketTransport {
...
+            .remove(&connection_id)
+            .ok_or(Error::ConnectionDoesntExist(connection_id))?;
+
+        self.on_inbound_connection(connection_id, pending.connection, pending.address);
+
+        Ok(())
+    }
+
+    fn reject_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
+        self.canceled.insert(connection_id);
```
dmitry-markin, line 453:
```suggestion
```

I don't think we should insert the connection ID into `canceled`, as this is not an outbound connection established without negotiation (see https://github.com/paritytech/litep2p/issues/197 for a detailed explanation).

dmitry-markin (approved):
Nicely done! :+1:

src/transport/websocket/mod.rs
lexnv: Yep that makes sense! Nice catch 🙏

Issue #195, opened 2024-08-06
kad: Extend `MemoryStore` with provider records


dmitry-markin:
Extend [`MemoryStore`](https://github.com/paritytech/litep2p/blob/3b664f7bef7cbf86176f8929bac349f9340462d8/src/protocol/libp2p/kademlia/store.rs#L42) with [provider records](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#content-provider-advertisement-and-discovery). The records should expire after the [expiration interval](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#content-provider-advertisement-and-discovery), which should be configurable.

Part of https://github.com/paritytech/litep2p/issues/196.

PR #200, opened 2024-08-12
kad: Add support for provider records to `MemoryStore`

dmitry-markin:
Resolves https://github.com/paritytech/litep2p/issues/195.

### DIFF
--- a/src/protocol/libp2p/kademlia/record.rs
+++ b/src/protocol/libp2p/kademlia/record.rs
@@ -19,7 +19,10 @@
 // FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 // DEALINGS IN THE SOFTWARE.
 
-use crate::PeerId;
+use crate::{
+    protocol::libp2p::kademlia::types::{Distance, Key as KademliaKey},
+    Multiaddr, PeerId,
+};
 
 use bytes::Bytes;
 use multihash::Multihash;
@@ -117,4 +120,35 @@ pub struct PeerRecord {
 
     /// The provided record.
     pub record: Record,
+}
+
+/// A record keeping information about a content provider.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct ProviderRecord {
+    /// Key of the record.
+    pub key: Key,
+
+    /// Key of the provider, based on its peer ID.
+    pub provider: PeerId,
+
+    /// Cached addresses of the provider.
+    pub addresses: Vec<Multiaddr>,
+
+    /// The expiration time of the record. The provider records must always have the expiration
+    /// time.
+    pub expires: Instant,
+}
+
+impl ProviderRecord {
+    /// The distance from the provider's peer ID to the provided key.
+    pub fn distance(&self) -> Distance {
+        // Note that the record key is raw (opaque bytes). In order to calculate the distance from
+        // the provider's peer ID to this key we must first hash both.
+        KademliaKey::from(self.provider).distance(&KademliaKey::new(self.key.clone()))
+    }
+
+    /// Checks whether the record is expired w.r.t. the given `Instant`.
+    pub fn is_expired(&self, now: Instant) -> bool {
+        now >= self.expires
+    }
 }
--- a/src/protocol/libp2p/kademlia/store.rs
+++ b/src/protocol/libp2p/kademlia/store.rs
@@ -427,12 +427,42 @@ mod tests {
     }
 
     #[test]
+    fn closest_providers_kept() {
+        let mut store = MemoryStore::with_config(MemoryStoreConfig {
+            max_providers_per_key: NonZeroUsize::new(10).unwrap(),
+            ..Default::default()
+        });
+        let key = Key::from(vec![1, 2, 3]);
+        let providers = (0..20)
+            .map(|_| ProviderRecord {
+                key: key.clone(),
+                provider: PeerId::random(),
+                addresses: vec![multiaddr!(Ip4([127, 0, 0, 1]), Tcp(10000u16))],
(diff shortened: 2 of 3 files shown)


src/protocol/libp2p/kademlia/record.rs
```diff
@@ -118,3 +121,34 @@ pub struct PeerRecord {
...
+
+    /// Key of the provider, based on its peer ID.
+    pub provider: PeerId,
+
+    /// Cached addresses of the provider.
+    pub addresses: Vec<Multiaddr>,
+
+    /// The expiration time of the record. The provider records must always have the expiration
+    /// time.
+    pub expires: Instant,
```
lexnv, line 139: Nice! I wonder if we should do a similar thing for the `Record`. It may not be entirely compliant with the spec, however it will force records out of the store. Do you think it would be beneficial to cap `Record::expires` at say 24h (or other number)?

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -115,6 +122,101 @@ impl MemoryStore {
...
+    /// this `key`, the new provider is only inserted if its closer to `key` than
+    /// the furthest already inserted provider. The furthest provider is then discarded.
+    ///
+    /// Returns `true` if the provider was added, `false` otherwise.
+    pub fn put_provider(&mut self, provider_record: ProviderRecord) -> bool {
+        // Make sure we have no more than `max_provider_addresses`.
+        let provider_record = {
+            let mut record = provider_record;
+            record.addresses.truncate(self.config.max_provider_addresses);
+            record
```
lexnv, line 155: nit: Could this be written on a single line as: 

```
pub fn put_provider(&mut self, mut provider_record: ..

provier_record.addresses.truncate(self.config.max_provider_addresses);
```

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -115,6 +122,101 @@ impl MemoryStore {
...
+        };
+
+        let can_insert_new_key = self.provider_keys.len() < self.config.max_provider_keys;
+
+        match self.provider_keys.entry(provider_record.key.clone()) {
+            Entry::Vacant(entry) =>
+                if can_insert_new_key {
```
lexnv, line 162: nit: Could move the check here `if self.provider_keys.len() < self.config.max_provider_keys`

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -115,6 +122,101 @@ impl MemoryStore {
...
+                // Providers under every key are sorted by distance, with equal distances meaning
+                // peer IDs (more strictly, their hashes) are equal.
+                let provider_position =
+                    providers.binary_search_by(|p| p.distance().cmp(&provider_record.distance()));
+
+                match provider_position {
+                    Ok(i) => {
+                        // Update the provider in place.
+                        let existing_provider =
+                            providers.get_mut(i).expect("index was found in the vector; qed");
```
lexnv, line 187: nit: would `providers[i] = provider_record` work here?

src/protocol/libp2p/kademlia/record.rs
dmitry-markin: Not sure about `Record`. It's specifically highlighted in the libp2p protocol spec that provider records must expire, because providers can go offline.

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: It is not allowed by the borrow checker, as we already borrowed `providers_keys` as mutable.

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: I didn't want to make `provider_record` mutable just to truncate addresses.

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -115,6 +122,101 @@ impl MemoryStore {
...
+                        *existing_provider = provider_record;
+
+                        true
+                    }
+                    Err(i) => {
+                        // `Err(i)` contains the insertion point.
+                        if i == self.config.max_providers_per_key {
+                            tracing::trace!(
+                                target: LOG_TARGET,
```
lexnv, line 196: Should we also try to add the provider if any provierds already inserted are expired? Could probably tackle this later

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: Probably yes, wanted to highlight that the index is correct. But this should be obvious.

lexnv (approved):
LGTM! Nice job here 👍

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: Sure, makes sense. Will depend on whether we will implement dropping expired providers lazily when modifying the store, or implement an "active" garbage collector (`MemoryStore` already has `next_event` event loop support).

dmitry-markin:
> LGTM! Nice job here 👍

Thanks for review! :heart:

src/protocol/libp2p/kademlia/record.rs
dmitry-markin: Aah, I see what you mean. Yes, we can introduce some maximum TTL for records and require publishers to renew them, to not store stale records forever. Let's think if there are applications for long-lived records where the original publisher goes offline. I.e., we can run a command once and publish a record — are there scenarios where this can be useful?

dmitry-markin: May be some credential storing, or a DNS-like system?

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: Done

Issue #205, opened 2024-08-14
transports: Remove `ConnectionError::Timeout` variant


lexnv:
The `ConnectionError::Timeout` variant can be deduced from the `FailedToNegotiate::SubstreamError`.

https://github.com/paritytech/litep2p/blob/c9e7affa1131fce1dd6a5bffcb9236c66facdd90/src/transport/tcp/connection.rs#L85-L106

Issue #204, opened 2024-08-14
error: Refactor the overarching litep2p::Error


lexnv:
The litep2p::Error contains multiple variants that grow over time.

Clean up the error into something more coherent and group errors by individual components / use-case, similar to the `DialError`.

This will help us provide more descriptive error messages and identify easily the place of occurrence:

https://github.com/paritytech/litep2p/blob/c9e7affa1131fce1dd6a5bffcb9236c66facdd90/src/error.rs#L43

Issue #203, opened 2024-08-14
identify: Report observer addresses of peers that succeeded dial attempts


lexnv:
Correlate `DialFailure` and `ListDialFailures` attempts with the Identify response provided to peers.

The addresses the node could not dial should be removed from the list of addresses we provide back to the peer.

This ensures the remote peer has a healthy view of its addresses and leads to better connectivity over time.
Libp2p uses a similar approach, caching individual peer addresses and removing the addresses the node failed to dial.

dmitry-markin:
I'm not sure I understand this issue correctly, but here is my understanding of the Identify operation. As per libp2p [spec](https://github.com/libp2p/specs/tree/master/identify#observedaddr), `observed_addr` is the connection source address of a peer initiating the connection. It is reported in any case and is part of Identify protocol.

In libp2p, Identify protocol implementation keeps a cache of remote peer addresses to provide them when dialing peers, and this is why this list is cleaned up of unreachable addresses. But the observed address is always reported back.

In litep2p peer addresses are discovered entirely through Kademlia DHT routing table, without caching the remote peer listen addresses in the Identify protocol implementation.

So, IMO we shouldn't modify the Identify protocol implementation in litep2p. If we need to check the reachability of external addresses after applying the "many peers have seen the same address" heuristic, it should be done using a different protocol, similar to [AutoNAT](https://docs.libp2p.io/concepts/nat/autonat/).

dmitry-markin:
Also, the heuristic of not reporting back the failed addresses won't work in case of restricted cone NATs, as in this case the dial attempts of the peer previously dialed by another peer behind NAT will succeed, while no other peers will be able to reach the peer behind NAT using discovered address and port. AutoNAT tries to solve this issue by using different IP to probe the addresses.

Issue #128, opened 2024-05-28
error: Merge `NegociatedError` with `SubstreamError`


lexnv:
https://github.com/paritytech/litep2p/blob/467a457c78785819e91ecdb7a5a926b7cd1c3d25/src/error.rs#L42-L121

PR #206, opened 2024-08-14
errors: Add `DialError` error and `ListDialFailures` event for better error reporting

lexnv:
The purpose of this PR is to pave the way for making the Identify protocol more robust, which is currently linked with the low number of peers and connective issues over a long period of time
- https://github.com/paritytech/polkadot-sdk/issues/4925

This PR adds a coherent `DialError` that exposes the minimal information users need to know about dial failures.
- https://github.com/paritytech/polkadot-sdk/issues/5239

A new litep2p event is added for reporting multiple dial errors that occur on different protocols back to the user:

```rust
    /// A list of multiple dial failures.
    ListDialFailures {
        /// List of errors.
        ///
        /// Depending on the transport, the address might be different for each error.
        errors: Vec<(Multiaddr, DialError)>,
    },
```

This event eases the debugging of substrate connectivity issues. At the same time, it can be used in a future PR to inform back to the Identify protocol which self-reported addresses of some peers are unreachable:
- https://github.com/paritytech/litep2p/issues/203

### Next Steps
- Add more tests
- Warp sync + sync full nodes since this is touching individual transports

### Future Work
- The overarching `litep2p::Error` needs a closer look and a refactoring:
  - https://github.com/paritytech/litep2p/issues/204
  - https://github.com/paritytech/litep2p/issues/128
  
- ConnectionError event for individual transports can be simplified:
  - https://github.com/paritytech/litep2p/issues/205
  
- I've observed some inconsistencies in handling TCP vs WebSocket connection timeouts. I believe that we can have another pass and share even more code between them:
  - https://github.com/paritytech/litep2p/issues/70

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -297,7 +297,7 @@ pub enum DnsError {
     ///
     /// For example, DNSv4 was expected but DNSv6 was provided.
     #[error("DNS type is different from the provided IP address")]
-    MismatchDnsVersion,
+    IpVersionMismatch,
 }
 
 impl From<MultihashGeneric<64>> for Error {
--- a/src/error.rs
+++ b/src/error.rs
@@ -146,8 +146,8 @@ pub enum AddressError {
     #[error("Address not available")]
     AddressNotAvailable,
     /// The provided address contains an invalid multihash.
-    #[error("Invalid multihash: `{0:?}`")]
-    InvalidMultihash(Multihash),
+    #[error("Multihash does not contain a valid peer ID : `{0:?}`")]
+    InvalidPeerId(Multihash),
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -299,7 +299,7 @@ pub enum DnsError {
 
 impl From<MultihashGeneric<64>> for Error {
     fn from(hash: MultihashGeneric<64>) -> Self {
-        Error::AddressError(AddressError::InvalidMultihash(hash))
+        Error::AddressError(AddressError::InvalidPeerId(hash))
     }
 }
 
@@ -365,7 +365,7 @@ impl From<ParseError> for Error {
 
 impl From<MultihashGeneric<64>> for AddressError {
     fn from(hash: MultihashGeneric<64>) -> Self {
-        AddressError::InvalidMultihash(hash)
+        AddressError::InvalidPeerId(hash)
     }
 }
 
--- a/src/error.rs
+++ b/src/error.rs
@@ -203,6 +203,11 @@ pub enum NegotiationError {
     StateMissmatch,
     #[error("Peer ID mismatch: expected `{0}`, got `{1}`")]
     PeerIdMismatch(PeerId, PeerId),
+
+    #[cfg(feature = "quic")]
+    #[error("Failed to negotiate QUIC: `{0}`")]
+    Quinn(quinn::ConnectionError),
+
     // TODO: Convert tokio_tungstenite::accept_async into `NegotiationError` for some cases (ie
     // ConnectionClosed).
     #[error("Other error occurred: `{0}`")]
@@ -247,6 +252,21 @@ pub enum DialError {
     #[cfg(feature = "websocket")]
     #[error("WebSocket error: `{0}`")]
     WebSocket(#[from] tokio_tungstenite::tungstenite::error::Error),
+
+    #[cfg(feature = "quic")]
+    #[error("QUIC error: `{0}`")]
+    Quic(#[from] QuicError),
+}
+
+#[cfg(feature = "quic")]
+#[derive(Debug, thiserror::Error)]
+pub enum QuicError {
+    #[error("Invalid certificate")]
+    InvalidCertificate,
+    #[error("Failed to negotiate QUIC: `{0}`")]
+    ConnectionError(#[from] quinn::ConnectionError),
+    #[error("Failed to connect to peer: `{0}`")]
+    ConnectError(#[from] quinn::ConnectError),
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/src/error.rs
+++ b/src/error.rs
@@ -291,7 +291,7 @@ pub enum QuicError {
 #[derive(Debug, thiserror::Error)]
 pub enum DnsError {
     /// The DNS resolution failed to resolve the provided URL.
-    #[error("Dns failed to resolve url `{0}`")]
+    #[error("DNS failed to resolve url `{0}`")]
     ResolveError(String),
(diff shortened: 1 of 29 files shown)


src/protocol/protocol_set.rs
```diff
@@ -274,27 +277,36 @@ impl ProtocolSet {
...
 
         let (protocol, fallback) = match self.fallback_names.get(&protocol) {
             Some(main_protocol) => (main_protocol.clone(), Some(protocol)),
             None => (protocol, None),
         };
 
-        self.protocols
-            .get_mut(&protocol)
-            .ok_or(Error::ProtocolNotSupported(protocol.to_string()))?
+        let Some(protocol_context) = self.protocols.get(&protocol) else {
```
dmitry-markin, line 288: nit: Actually, I liked how it was implemented rust-way with `ok_or()`. `let .. else` feels somewhat C++-ish.

May be add some `#[from]` macros to the types and keep `.ok_or()?`?

src/error.rs
```diff
@@ -183,14 +247,57 @@ pub enum NotificationError {
...
+#[derive(Debug, thiserror::Error)]
+pub enum DnsError {
+    /// The DNS resolution failed to resolve the provided URL.
+    #[error("Dns failed to resolve url `{0}`")]
+    ResolveError(String),
+    /// The DNS expected a different IP address version.
+    ///
+    /// For example, DNSv4 was expected but DNSv6 was provided.
+    #[error("DNS type is different from the provided IP address")]
+    MismatchDnsVersion,
```
dmitry-markin, line 300: nit: may be `DnsError::IpVersionMismatch`?

src/transport/manager/mod.rs
```diff
@@ -1706,13 +1713,19 @@ impl TransportManager {
...
+                                    return Some(TransportEvent::OpenFailure { connection_id, errors: grouped_errors });
+                                }
+                                Ok(None) => {
+                                    tracing::trace!(
+                                        target: LOG_TARGET,
+                                        ?connection_id,
+                                        "open failure, but not the last transport",
+                                    );
+
+                                    self.opening_errors.entry(connection_id).or_default().extend(errors);
```
dmitry-markin, line 1727: Just thinking out loud: if the connection fails, we report all the errors for all the addresses, but if the connection succeeds, we don't report the errors for other addresses tried and failed. Feels like a little bit of asymmetry to me. Do we really need to report individual errors for all the addresses?

src/transport/websocket/connection.rs
```diff
@@ -310,13 +313,12 @@ impl WebSocketConnection {
...
             if peer != dialed_peer {
-                return Err(Error::PeerIdMismatch(dialed_peer, peer));
+                return Err(NegotiationError::PeerIdMismatch(dialed_peer, peer));
             }
         }
 
-        let stream: NoiseSocket<BufferedStream<_>> = stream;
-
         tracing::trace!(target: LOG_TARGET, "noise handshake done");
+        let stream: NoiseSocket<BufferedStream<_>> = stream;
```
dmitry-markin, line 321: nit: Why move this below `trace!`?

src/error.rs
```diff
@@ -127,48 +127,112 @@ pub enum Error {
...
 }
 
+/// Error type for address parsing.
 #[derive(Debug, thiserror::Error)]
 pub enum AddressError {
+    /// The provided address does not correspond to the transport protocol.
+    ///
+    /// For example, this can happen when the address used the UDP protocol but
+    /// the handling transport only allows TCP connections.
     #[error("Invalid protocol")]
```
dmitry-markin, line 137:
```suggestion
    #[error("Invalid address for protocol")]
```
?

src/transport/websocket/mod.rs
```diff
@@ -618,22 +603,15 @@ impl Stream for WebSocketTransport {
...
-                                connection_id,
-                                address,
-                                error: error.error,
-                            })),
-                        None => {
-                            tracing::debug!(target: LOG_TARGET, ?error, "failed to establish connection")
-                        }
-                    },
-                    None => {
-                        tracing::debug!(target: LOG_TARGET, ?error, "failed to establish connection")
```
dmitry-markin, line 634: nit: May be keep these log messages to catch the state mismatch if it happens? May be even add `debug_assert!(false)`?

dmitry-markin (approved):
Good job and well done here! :tada:

src/error.rs
```diff
@@ -251,6 +430,30 @@ impl From<quinn::ConnectionError> for Error {
...
+    }
+}
+
+#[cfg(feature = "websocket")]
+impl From<tokio_tungstenite::tungstenite::error::Error> for DialError {
+    fn from(error: tokio_tungstenite::tungstenite::error::Error) -> Self {
+        DialError::NegotiationError(NegotiationError::WebSocket(error))
+    }
+}
+
```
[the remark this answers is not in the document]
lexnv: Yep, have replaced some of them with `map_err`, the issue would be that `thiserror` does not handle the conversion between nested variants. For example, it cannot deduce the `impl From<WebSocketErr> for DialError`

src/error.rs
```diff
@@ -127,48 +127,112 @@ pub enum Error {
...
+    /// The provided address contains an invalid multihash.
+    #[error("Invalid multihash: `{0:?}`")]
+    InvalidMultihash(Multihash),
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum ParseError {
+    /// Cannot parse the multihash from the provided bytes.
     #[error("Invalid multihash: `{0:?}`")]
     InvalidMultihash(Multihash),
```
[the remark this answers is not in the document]
lexnv: Yep makes sense, have renamed it to `InvalidPeerId`

src/transport/websocket/mod.rs
lexnv: After looking a bit into this, we only have a `pending_dial` associated with the connection ID if the connection was outbound. For inbound connections, we don't track which address failed here

src/transport/manager/mod.rs
lexnv: That sounds reasonable, we can probably look into extending a DialSuccess with a list of failed to dial addresses. Don't quite remember if libp2p is doing a similar thing, but that sounds familiar. Yep, wanted to catch a bit more details into why we have seen so many failures from our metrics, the `UnknownError` was a bit too cryptic :D

Polkadot-Forum:

https://forum.polkadot.network/t/litep2p-network-backend-updates/9973/1

PR #207, opened 2024-08-14
cargo: Enable all features by default

lexnv:
This PR enables all transport features by default.

This eases the development process in the repo, ensuring we don't miss adjustments to individual transports.

At the same time, developers/consumers who are interested in only specific features can disable the default ones.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,10 +86,18 @@ tracing-subscriber = { version = "0.3.16", features = ["env-filter"] }
 futures_ringbuf = "0.4.0"
 
 [features]
-custom_sc_network = []
+# Enable all transports by default.
+default = ["websocket", "quic", "webrtc"]
+
+# Enable QUIC support.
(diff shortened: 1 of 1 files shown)


dmitry-markin:
I wonder why this is better than passing `--all-features` during development/testing, which is already done automatically on CI for `cargo check` & `cargo test`?

lexnv:
Indeed this is a bit opinionated. Basically, my reasons for this are:
- ease the local development (ie not create custom vscode or other editor configs to run clippy with other features) to ensure we capture all components before pushing a PR
- enable all transports by default (substrate-nodes will probably make use of all of them) so that most users don't have to worry about feature flags unless they want to enable just a few bits of them selectively 

No strong opinions here, could probably go the route of creating custom configs :D

dmitry-markin (approved):

lexnv:
That makes sense, no strong reason to change this. Will revisit if it ever comes an issue, thanks for the input 🙏

Issue #191, opened 2024-07-31
identify: Update local node listen addresses


lexnv:
The identify protocol provides as response a `listen_addresses` field.
The remote peer uses this field to insert self reported addresses into the Kademlia DHT.

Substrate makes use of the listen addresses [here](https://github.com/paritytech/polkadot-sdk/blob/00a5cb3a0b43492391141f0a98f5e1666ebfd3cf/substrate/client/network/src/litep2p/mod.rs#L916-L918).

The problem is that the identify protocol starts with a static set of addresses, which may expire or become invalid:

https://github.com/paritytech/litep2p/blob/2d1a4b48bb7f0f768b825ce88d456fe741c922df/src/protocol/libp2p/identify.rs#L272-L276

This might be one of the reasons why we've seen low connectivity with peers over the long term (~90 days).

The Identify protocol can be extended to receive verified addresses for healthier listen addresses.
For the Substrate perspective, when we deem a "remote address" as valid (after 2 encounters), that address needs to be reported back to the Identify protocol.
Libp2p already has a similar mechanism of discovery via the `ExternalAddressDiscovered` and confirming addresses.

dmitry-markin:
Yeah, it shouldn't be like this. With libp2p backend `Identify` "listen addresses" field actually contains discovered external addresses, no matter when they were discovered.

Issue #211, opened 2024-08-20
Make `litep2p` aware of external addresses


dmitry-markin:
Currently, `litep2p` knows only about listen addresses of the node. External addresses handling is left to the client code.

Let `litep2p` know external addresses or even add some heuristic mechanisms of discovering external addresses (based on `Identify` protocol) to move the logic from the client code to the library.

This is needed to implement https://github.com/paritytech/litep2p/issues/191 and to report the addresses of the local node acting as a content provider in Kademlia https://github.com/paritytech/litep2p/issues/196.

PR #212, opened 2024-08-20
Expose API for `PublicAddresses` 

lexnv:
This PR exposes an object for handling public addresses.
- `add_address` - adds an address to the given set, inserting the local peer ID or returning an error on different peer Ids
- `remove_address` - removes an exact address from the set
- `get_addresses` - list of actual addresses

Users of litep2p can obtain the given list via litep2p::public_addresses method.
The lists are exposed internally by a similar API on the transport handler (the handler that communicates with the transport manager).

The identify protocol provides a concatenated list of:
- user provided addresses via the config
- listen addresses 
- external addresses

This ensures that Substrate can use a custom heuristic to update the list of external addresses.

### Testing Done
- Added an integration test with 2 litep2p nodes

Closes: https://github.com/paritytech/litep2p/issues/191

cc @paritytech/networking

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2870,7 +2870,7 @@ version = "0.6.2"
 dependencies = [
  "async-trait",
  "asynchronous-codec 0.7.0",
- "bs58 0.4.0",
+ "bs58 0.5.1",
  "bytes",
  "cid",
  "ed25519-dalek",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ prost-build = "0.13"
 
 [dependencies]
 async-trait = "0.1.81"
-bs58 = "0.4.0"
+bs58 = "0.5.1"
 bytes = "1.6.1"
 cid = "0.10.1"
 ed25519-dalek = { version = "2.1.1", features = ["rand_core"] }


src/protocol/libp2p/identify.rs
```diff
@@ -98,20 +148,27 @@ impl Config {
         protocol_version: String,
         user_agent: Option<String>,
         public_addresses: Vec<Multiaddr>,
-    ) -> (Self, Box<dyn Stream<Item = IdentifyEvent> + Send + Unpin>) {
+    ) -> (
+        Self,
+        IdentifyPublicAddresses,
+        Box<dyn Stream<Item = IdentifyEvent> + Send + Unpin>,
```
altonen, line 154: This could return `IdentifyHandle` which would hold `IdentifyPublicAddresses` and would implement `Stream<Item = IdentifyEvent>`, following the design of other protocols.

lexnv: I like that idea, will certainly implement it! Thanks for the review!

Have put this on pending shortly, I have a hunch we'll need to expose an `ExternalAddrConfirmed` (like libp2p) to suffice other usecases (https://github.com/paritytech/litep2p/issues/211). Will wait for Dmitry to confirm that these are needed for the bootnodes on DHT🙏

dmitry-markin: Sorry to have not responded here. We discussed https://github.com/paritytech/litep2p/issues/211 with @lexnv, and my idea was to introduce a globally accessible list of public addresses in litep2p. If some protocol needs external addresses, it can fetch them. This includes `Identify` and `Kademlia` protocols now.

As a first step, we can manually notify `litep2p` about confirmed external addresses from the client code. As a second step, we can move the heuristic of confirming external addresses seen by three or more peers from `polkadot-sdk` to `litep2p`. I have updated https://github.com/paritytech/litep2p/issues/211 to be more clear about this.

I don't think `ExternalAddrConfirmed` event will help with providing up-to-date peer addresses for local content providers in Kademlia from within `litep2p`, but the option to get addresses from the global list will.

altonen: Your plan sounds good. The easiest place to store the external addresses is probably `TransportManager`, partly since there already exists code for storing known addresses via `TransportManagerHandle` and since every protocol has a handle to `TransportManager`, it should require little architectural changes.

src/public_addresses.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Remove the exact public address.
+    ///
+    /// The provided address must contain the local peer ID.
+    pub fn remove_public_address(&self, address: &Multiaddr) -> bool {
+        self.inner.write().remove(address)
+    }
+
+    /// Returns a vector of the available listen addresses.
+    pub fn get_addresses(&self) -> Vec<Multiaddr> {
```
dmitry-markin, line 74: As the struct is called `PublicAddresses`, may be remove `public` from all function names for consistency:
- `add_address()`
- `remove_address()`
- `get_addresses()`

src/public_addresses.rs
```diff
@@ -0,0 +1,145 @@
...
+        self.inner.read().iter().cloned().collect()
+    }
+
+    /// Checks if the address contains the local peer ID.
+    fn check_local_peer(&self, mut address: Multiaddr) -> Result<Multiaddr, Multiaddr> {
```
dmitry-markin, line 79:
```suggestion
    fn ensure_local_peer(&self, mut address: Multiaddr) -> Result<Multiaddr, Multiaddr> {
```
As we also append the local peer ID?

src/lib.rs
```diff
@@ -138,7 +139,7 @@ pub struct Litep2p {
     local_peer_id: PeerId,
 
     /// Listen addresses.
-    listen_addresses: Vec<Multiaddr>,
+    listen_addresses: PublicAddresses,
```
dmitry-markin, line 142: I don't think this is quite correct. Public addresses may be completely different form the listen addresses if we are behind NAT. We will also receive an error when try to bind to a public NAT address.

I would keep a clear separation between existing listen addresses and public addresses introduced in this PR.

src/lib.rs
```diff
@@ -315,9 +315,6 @@ impl Litep2p {
 
             for address in transport_listen_addresses {
                 transport_manager.register_listen_address(address.clone());
-                listen_addresses.push(address.with(Protocol::P2p(
```
dmitry-markin, line 318: Here and below we should keep old listen addresses separate from public addresses.

src/protocol/libp2p/identify.rs
```diff
@@ -200,16 +201,31 @@ pub(crate) struct Identify {
...
+                if peer_id != local_peer_id {
+                    return None;
+                }
+
+                return Some(address);
+            }
+
+            Some(address.with(Protocol::P2p(local_peer_id.into())))
+        });
+        listen_addresses.inner.write().extend(filtered_public_addr);
```
dmitry-markin, line 222: I don't think we should modify a global `libp2p` struct from Identify protocol.

src/protocol/libp2p/identify.rs
```diff
@@ -265,15 +281,13 @@ impl Identify {
             }
         };
 
+        let listen_addrs =
```
dmitry-markin, line 284: To comply with libp2p implementation, we should include both public addresses and listen addresses here.

src/transport/manager/mod.rs
```diff
@@ -267,13 +267,13 @@ impl TransportManager {
...
         let (event_tx, event_rx) = channel(256);
-        let listen_addresses = Arc::new(RwLock::new(HashSet::new()));
+        let listen_addresses = PublicAddresses::new(local_peer_id);
         let handle = TransportManagerHandle::new(
             local_peer_id,
             peers.clone(),
             cmd_tx,
             supported_transports,
-            Arc::clone(&listen_addresses),
+            listen_addresses.clone(),
```
dmitry-markin, line 276: Is `clone()` needed here?

src/transport/manager/mod.rs
```diff
@@ -381,16 +381,16 @@ impl TransportManager {
...
-        assert!(!address.iter().any(|protocol| std::matches!(protocol, Protocol::P2p(_))));
-
-        let mut listen_addresses = self.listen_addresses.write();
-
-        listen_addresses.insert(address.clone());
-        listen_addresses.insert(address.with(Protocol::P2p(
-            Multihash::from_bytes(&self.local_peer_id.to_bytes()).unwrap(),
-        )));
+        if let Err(address) = self.listen_addresses.add_public_address(address) {
+            tracing::warn!(target: LOG_TARGET, ?address, "failed to register listen address");
```
dmitry-markin, line 392: May be extend the error returned from `add_public_address()` to indicate a reason? Not super important, feel free to ignore.

src/transport/manager/mod.rs
```diff
@@ -267,13 +267,13 @@ impl TransportManager {
         let peers = Arc::new(RwLock::new(HashMap::new()));
         let (cmd_tx, cmd_rx) = channel(256);
         let (event_tx, event_rx) = channel(256);
-        let listen_addresses = Arc::new(RwLock::new(HashSet::new()));
+        let listen_addresses = PublicAddresses::new(local_peer_id);
```
dmitry-markin, line 270: Again, it looks like there is some confusion between public addresses and listen addresses.

dmitry-markin (commented):
The `PublicAddresses` struct is quite nice and has a convenient API, but I think we need to keep clear separation between listen addresses and public addresses (see inline comments).

dmitry-markin:
The PR description and title don't match the PR now :)

src/lib.rs
lexnv: Yep that makes sense, I've introduced a similar object for ListenAddresses as for PublicAddresses, let me know what you think of it 🙏

src/protocol/libp2p/identify.rs
lexnv: Yep that makes sense, have decided to keep the:
- user provided addresses
- listen addresses
- external addresses

src/transport/manager/mod.rs
lexnv: Yep, we need a reference in the transport manager as well

src/transport/manager/mod.rs
lexnv: Yep I like that idea, have added a small error enum

src/addresses.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// Set of the addresses the local node listens on.
+///
+/// The format of the addresses stored in the set contain the local peer ID.
+/// This requirement is enforced by the [`ListenAddresses::add_address`] method.
+///
+/// The listen addresses are populated during the construction of the Litep2p object.
+#[derive(Debug, Clone)]
+pub struct ListenAddresses {
```
dmitry-markin, line 86: Actually, do we need a different struct? May be we can go with just one `Addresses`, and instantiate it to manage `listen_addresses` and `public_addresses`?

One thing with `ListenAddresses` though, is that ideally we want to start listening on the newly added address there. So, I'm not sure if we need to provide API for this struct, or just completely keep the old implementation of `listen_addresses` and use a dedicated struct only for `PublicAddresses`.

src/protocol/libp2p/identify.rs
```diff
@@ -183,8 +183,14 @@ pub(crate) struct Identify {
...
 
+    /// Listen addresses.
+    listen_addresses: ListenAddresses,
+
     /// Public addresses.
-    listen_addresses: PublicAddresses,
+    public_addresses: PublicAddresses,
+
+    /// User provided list of addresses.
+    user_addresses: Vec<Vec<u8>>,
```
dmitry-markin, line 193: I don't think we need one more place to specify public addresses. Getting public addresses via a handle from `TransportManager` should be enough.

src/addresses.rs
lexnv: That makes sense. Yep ideally we should be able to say `litep2p.register_listen_addr` and then give that information to individual protocol listeners. In this case, I'll just keep public addresses for this PR since a `ListenAddr` struct would complicate that implementation, thanks!

src/protocol/libp2p/identify.rs
lexnv: Oki doki, have removed these thanks! 🙏

dmitry-markin: I probably was not quite clear on this, but I meant we don't need to let user pass addresses directly to Identify config if we already have an API to notify litep2p about public addresses.

dmitry-markin (approved):
Looks good! As now we can notify litep2p about public addresses, IMO we can safely remove public addresses from the Identify config.

Issue #201, opened 2024-08-13
kad: Implement `ADD_PROVIDER` & `GET_PROVIDERS` RPCs


dmitry-markin:
Spec: https://github.com/libp2p/specs/tree/master/kad-dht#content-provider-advertisement-and-discovery

Part of: https://github.com/paritytech/litep2p/issues/196

PR #213, opened 2024-08-21
kad: Handle `ADD_PROVIDER` & `GET_PROVIDERS` network requests

dmitry-markin:
This implements the `ADD_PROVIDER` & `GET_PROVIDER` Kademlia RPC without initiating the queries itself. This should already allow participating in the DHT with content providers and serve all network requests, but won't allow initiating queries (like getting the content provider & starting providing) locally.

Implements "server side" of https://github.com/paritytech/litep2p/issues/201.

### Next steps
- Initiate content provider queries in `litep2p`
- Once the client API for queries is exposed, add conformance tests with `libp2p`


lexnv (approved):
LGTM! Nice job here 👍

PR #217, opened 2024-08-26
substream: Use write_all instead of manually writing bytes   

lexnv:
This PR refactors the substream writing to replace the `while remaining { write(byte) }` with the `write_all` method.
- introduce a dedicated function that is shared between substream implementation that handles the varint payload
- replace hardcoded buffer [0; 10] with the encoded buffer of `usize` from the unsigned_varint crate
- flush the identity protocol similar to the varint protocol after successfully writing data

This was driven by the fact that the identity protocol (ie plain bytes without length prefix) uses write_all; and the higher level functions never use that partially written information. I expect the performance improvement to be negligible, however this helps with code maintenance and consistency. 

cc @paritytech/networking

### DIFF
--- a/src/substream/mod.rs
+++ b/src/substream/mod.rs
@@ -375,6 +375,7 @@ impl Substream {
         io.flush().await.map_err(From::from)
     }
 
+    /// Send unsigned varint payload to remote peer.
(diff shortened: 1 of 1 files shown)


dmitry-markin (approved):
:rocket:

PR #219, opened 2024-08-26
query-executor: Save the task waker on empty futures

lexnv:
This PR ensures that we save the task `waker` when the executor does not have any pending futures (messages queued to send / receive from the individual substreams).

When the `self.futures` are empty we returned `Poll::Pending`.
However, the given `QueryExecutor::poll_next` would never wake up again.
To mitigate this behavior, save the current waker and notify it upon modifications to `self.futures`.

This was working before because we relied on other tasks to make progress via the `loop { tokio::select! }`. However this ensures that the query executor is efficiently waking up and can work in isolation.

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/executor.rs
+++ b/src/protocol/libp2p/kademlia/executor.rs
@@ -79,6 +79,14 @@ pub struct FuturesStream<F> {
 }
 
 impl<F> FuturesStream<F> {
+    /// Create new [`FuturesStream`].
+    pub fn new() -> Self {
+        Self {
+            futures: FuturesUnordered::new(),
+            waker: None,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


dmitry-markin:
Yeah, this is a nasty feature of `FuturesUnordered`. In substrate/sync I solved it introducing a wrapper for `FuturesUnordered` with familiar API. May be we can use such abstraction here as well:

https://github.com/paritytech/polkadot-sdk/blob/master/substrate/client/network/sync/src/futures_stream.rs

lexnv:
That looks like a lovely API, nicely done there :D Will take inspiration and bring this into litep2p, thanks 🙏

dmitry-markin (approved):

PR #221, opened 2024-08-27
kad: Fix not retrieving local records

dmitry-markin:
This PR makes sure we always return a locally stored record if we have it.

This changes the existing behavior, so I am unsure if it breaks any assumptions in the client code. We should mention this breaking change in the release notes.


src/protocol/libp2p/kademlia/handle.rs
```diff
@@ -203,12 +203,12 @@ pub enum KademliaEvent {
...
 pub enum RecordsType {
-    /// Record was found in the local store.
+    /// Record was found in the local store and [`Quorum::One`] was used.
     ///
     /// This contains only a single result.
     LocalStore(Record),
 
-    /// Records found in the network.
+    /// Records found in the network. This can include the locally found record.
     Network(Vec<PeerRecord>),
```
lexnv, line 212: nit: Thinking out loud, since we include the local records and the records found from the network, we could try to remove the `RecordsType` enum all together.
Then, we could mention in the `KademliaEvent::GetRecordSuccess` that records fetched from the local store  are populated first in the list of records ie `Vec<PeerRecord>` and that users can distinguish between local and network records by comparing the peer ID of the record.

lexnv (approved):
LGTM! Nice catch here!

I think it is fine to mention any breaking changes in the release notes, we'll try to be more mindful in the future, especially after stabilizing litep2p in substrate 👍

PR #222, opened 2024-08-27
kad: Fix substream opening and dialing race

lexnv:
This PR makes the substream opening and dialing of query commands more robust.

In the past, the logic of a query action would:
- T0; try to open a substream
- T1; try to dial the peer

There is a race between the moment when a substream is opened and when the peer is dialed (T0 and T1), where the peer could connect to us. The race manifested after extensively running `FindNode` commands via [subp2p-explorer bench-cli](https://github.com/lexnv/subp2p-explorer/tree/main/bench-cli)

To mitigate this race, retry to open the substream on recoverable errors from dialing (ie Error::PeerAlreadyConnected).

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -642,54 +642,71 @@ impl Kademlia {
         }
     }
 
-    /// Handle next query action.
-    async fn on_query_action(&mut self, action: QueryAction) -> Result<(), (QueryId, PeerId)> {
-        match action {
-            QueryAction::SendMessage { query, peer, .. } => loop {
-                match self.service.open_substream(peer) {
-                    Err(_) => {
-                        tracing::trace!(target: LOG_TARGET, ?query, ?peer, "dial peer");
+    /// Open a substream with a peer or dial the peer.
+    fn open_substream_or_dial(
+        &mut self,
+        peer: PeerId,
+        action: PeerAction,
+        query: Option<QueryId>,
+    ) -> Result<(), Error> {
+        match self.service.open_substream(peer) {
+            Ok(substream_id) => {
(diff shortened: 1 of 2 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -642,50 +642,71 @@ impl Kademlia {
...
-                        .pending_actions
-                        .insert(substream_id, PeerAction::SendFindNode(query));
+                    Err(error) => {
+                        tracing::trace!(target: LOG_TARGET, ?query, ?peer, ?error, "Failed to dial peer");
+                        return Err(error);
+                    }
+                };
 
-                    Ok(())
+                // Dial returned `Error::AlreadyConnected`, retry opening the substream.
```
dmitry-markin, line 677: nit: Does this block indents badly if put into `Err(Error::AlreadyConnected)` arm? This way we won't need returns and it would be slightly easier to follow the logic. Feel free to ignore, it's likely a matter of personal taste.

dmitry-markin (approved):
LGTM! Other protocols are likely also racy when opening a substream?

src/protocol/libp2p/kademlia/mod.rs
lexnv: Yep that makes things clearer! Thanks for the suggestion 🙏

lexnv:
> Other protocols are likely also racy when opening a substream?

After a quick look, it seems this is the only place where we tried to open a substream and then dial on failure.
I've opened this issue to try to better handle the cases where we dial a peer, however we get back a `AlreadyConnected` error (such as request-response and notification protocols):
- https://github.com/paritytech/litep2p/issues/225

PR #224, opened 2024-08-30
transport: Implement `TransportService::local_peer_id()`

altonen:
Previously `local_peer_id` was declared `pub(crate)` which made it inaccessible to protocols implemented with `UserProtocol`

Implement getter for local peer ID, allowing all protocols having access to `TransportService` to query the local peer ID


lexnv (approved):
LGTM! Thanks 🙏

dmitry-markin (approved):

Issue #188, opened 2024-07-30
request-response: Provide more details into rejected requests for a high request failure


lexnv:
Substrate presents many rejected connections, especially for the `sync` protocol.
The metrics are configured [here](https://github.com/paritytech/polkadot-sdk/blob/15c056be52d9372efc457b9aa82397a47ef73505/substrate/client/network/src/litep2p/shim/request_response/mod.rs#L375) in substrate. 

![Screenshot 2024-07-30 at 15 17 25](https://github.com/user-attachments/assets/27bf0ac1-ebab-4cf0-9021-76624514aae4)

Running libp2p side by side with litep2p, we can observe from the previous figure a very high number of requests that are marked as "rejected".

Provide more details into rejected requests by enriching the `RequestResponseError::Rejected` with specific details:
- is the failure due to opening a substream?
- is the connection closed and we reject all active requests?
- is this a dial failure and we reported back immediately? etc

Detected while testing and debugging:
- https://github.com/paritytech/polkadot-sdk/pull/5029

PR #227, opened 2024-09-02
errors: Introduce immediate dial error and request-response rejection reasons

lexnv:
This PR makes several changes to the errors in general, with the main goal of extracting the reject-reason from the request response protocols:
- an `ImmediateDial` error is introduced for failing to dial peers due to internal errors (no addresses available, tried to dial self, already connected etc) to distinguish between network dial failures
- opening substreams now return a `SubstreamError` instead of using the overarching litep2p::Error
- substreams are now implementing `Stream<Error = SubstreamError>` for consistency with opening substreams
- Reject reasons include immediate dialing errors. Ideally, we could expose the same level of information that is exposed via `ListDialErrors` (provide a tuple of addresses with individual dial errors), however that would require a bigger refactor to the code. For now this information is enough for Substrate metrics to provide more information and align with litep2p metrics.

This PR is part of a bigger effort to simply the overarching error enum:
- https://github.com/paritytech/litep2p/issues/204

Closes: https://github.com/paritytech/litep2p/issues/188

### DIFF
--- a/src/protocol/request_response/mod.rs
+++ b/src/protocol/request_response/mod.rs
@@ -371,7 +371,7 @@ impl RequestResponseProtocol {
                     fallback_protocol,
                     Err(RequestResponseError::Timeout),
                 ),
-                Ok(Err(Error::IoError(ErrorKind::PermissionDenied))) => {
+                Ok(Err(SubstreamError::IoError(ErrorKind::PermissionDenied))) => {
                     tracing::warn!(
                         target: LOG_TARGET,
                         ?peer,
@@ -390,7 +390,7 @@ impl RequestResponseProtocol {
                     peer,
                     request_id,
                     fallback_protocol,
-                    Err(RequestResponseError::NotConnected),
+                    Err(RequestResponseError::NotConnected)
                 ),
                 Ok(Ok(_)) => {
                     tokio::select! {
--- a/src/protocol/request_response/mod.rs
+++ b/src/protocol/request_response/mod.rs
@@ -386,11 +386,11 @@ impl RequestResponseProtocol {
                         Err(RequestResponseError::TooLargePayload),
                     )
                 }
-                Ok(Err(_error)) => (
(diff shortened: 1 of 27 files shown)


tests/protocol/request_response.rs
```diff
@@ -314,7 +314,7 @@ async fn reject_request(transport1: Transport, transport2: Transport) {
         RequestResponseEvent::RequestFailed {
             peer: peer2,
             request_id,
-            error: RequestResponseError::Rejected
+            error: RequestResponseError::Rejected(RejectReason::ConnectionClosed)
```
dmitry-markin, line 317: Is it actually `ConnectionClosed`, or "SubstreamClosed"?

dmitry-markin (approved):
Very good work on errors self-descriptiveness!

src/protocol/request_response/mod.rs
```diff
@@ -388,7 +390,7 @@ impl RequestResponseProtocol {
                     peer,
                     request_id,
                     fallback_protocol,
-                    Err(RequestResponseError::NotConnected),
```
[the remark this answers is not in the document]
lexnv: I've changed this to return the `SubstreamOpenError` instead. In doing so, I've created a small `From<>` implementation to make sure that we report `io::ErrorKind::NotConnected` as a nicer reject reason

tests/protocol/request_response.rs
lexnv: That makes sense! Have added a new variant `SubstreamClosed`, the `ConectionClosed` should now be constructed only when the peer disconnects to distinguish between dropping the substream and completely closing the connection

I've also added PartialEq on a few other errors, since we are most likely comparing ErrorKind's anyway :D Thanks for the review 🙏

Polkadot-Forum:

https://forum.polkadot.network/t/litep2p-network-backend-updates/9973/1

PR #228, opened 2024-09-03
substream-set: Poll substreams fairly with poll indexes

lexnv:
This PR refactors the substream-set (used by request-response protocols):
- the context waker is saved to before returning poll::pending
- a poll index and vec<keys> is introduced to provide more fairness during the polling of substreams

### DIFF
--- a/src/substream/mod.rs
+++ b/src/substream/mod.rs
@@ -47,7 +47,7 @@ use std::{
     hash::Hash,
     io::ErrorKind,
     pin::Pin,
-    task::{Context, Poll},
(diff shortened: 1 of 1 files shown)


lexnv:
I'll do more testing with this PR since one test is failing because we expect fewer polls to happen

lexnv:
In this revision, I've replaced the `HashMap ++ Vec<Keys>` with an `indexMap::IndexMap`.

One test was previously failing due to the poll_index not incrementing properly

lexnv:
Planning to remove this entirely in the future, this is no longer needed

PR #229, opened 2024-09-03
Use static variables for DNS resolver and crypto noise parameters

lexnv:
This PR uses `std::sync::OnceLock` to create reusable variables only once:
- crypto/noise parameters are parsing the provided string only once
- tokio DNS resolver uses the same variable to leverage the DNS internal caching

cc @paritytech/networking


lexnv:
### Updates
- I've deprioritized this PR in favor of other refactoring and bug fixing work
- planning to remove the statics, so I will leave this open for a while

PR #234, opened 2024-09-03
kad: Providers part 3: publish provider records (start providing)

dmitry-markin:
Implement `ADD_PROVIDER` network request and execute `FIND_NODE` query to publish local provider to the target peers.

Builds upon https://github.com/paritytech/litep2p/pull/200 & https://github.com/paritytech/litep2p/pull/213.

Next PR in the series: https://github.com/paritytech/litep2p/pull/235.

### Follow-up
 - Make sure unit and integration tests do not reveal issues and merge together with other PRs in the series.

PR #235, opened 2024-09-03
kad: Providers part 4: refresh local providers

dmitry-markin:
Republish local providers every 22 hours.

Builds upon https://github.com/paritytech/litep2p/pull/234.

Next PR in the series: https://github.com/paritytech/litep2p/pull/236.

### Follow-up
 - Make sure unit and integration tests do not reveal issues and merge together with other PRs in the series.


src/protocol/libp2p/kademlia/handle.rs
```diff
@@ -319,89 +320,25 @@ impl KademliaHandle {
...
-    }
-
-    /// Try to initiate `FIND_NODE` query and if the channel is clogged, return an error.
-    pub fn try_find_node(&mut self, peer: PeerId) -> Result<QueryId, ()> {
-        let query_id = self.next_query_id();
-        self.cmd_tx
-            .try_send(KademliaCommand::FindNode { peer, query_id })
-            .map(|_| query_id)
-            .map_err(|_| ())
-    }
```
altonen, line 356: Sad to see these removed. They were convenient in places where `.await` was not available, e.g., inside a `Stream::poll_next()` implementation and allowed an easy way to deal with backpressure. The alternative is of course to call `KademliaHandle::find_node()` and manually pin the future but it's not quite as ergonomic.

One thing worth pointing w.r.t. blocking handle -> protocol calls. I believe the program can deadlock if, e.g., `KademliaProtocol` is blocking on an `event_tx.send()` call (because the event channel is full) and if the command queue towards the protocol is full. Next call to, e.g., `KademliaHandle::find_node()` will also block and that could deadlock the program because both the handle and the protocol implementation are waiting on each other to drain the queues.

dmitry-markin: Thanks for looking into this!

Basically, the alternative was to introduce either a mutex for generating the `QueryID`s, or introduce a compound
```
enum QueryId {
    User(UserQueryId),
    Local(LocalQueryId),
}
```
So we can generate `UserQueryId` in the handle and `LocalQueryId` in the protocol and they don't clash. But the implementation was too ugly with a lot of `match`es in places where such `QueryId` was used.

IDK, may be the plain mutex is a way to go.

altonen: Request-response uses `Arc<AtomicUsize>` that's passed to the protocol via `RequestResponseConfig` to allocate `RequestId`s both in the handle and in the protocol implementation:

https://github.com/paritytech/litep2p/blob/e3a22d5cfbf255e74be1acceb3ea6ec0baa43c09/src/protocol/request_response/config.rs#L59

Would that work?

dmitry-markin: Of course! Didn't think about atomics :see_no_evil:

dmitry-markin: Should be resolved. Thanks again for review and suggestions!

altonen: Thank you for bringing them back, much appreciated!

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -31,14 +31,21 @@ use crate::{
...
+};
 
 /// Default TTL for the records.
 const DEFAULT_TTL: Duration = Duration::from_secs(36 * 60 * 60);
 
 /// Default provider record TTL.
 const DEFAULT_PROVIDER_TTL: Duration = Duration::from_secs(48 * 60 * 60);
 
+/// Default provider republish interval.
+pub(super) const DEFAULT_PROVIDER_REFRESH_INTERVAL: Duration = Duration::from_secs(22 * 60 * 60);
```
lexnv, line 47: dq: What is rust-libp2p doing by default for this?

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -660,4 +756,6 @@ mod tests {
         assert_eq!(store.get_providers(&provider2.key), vec![provider2]);
         assert_eq!(store.get_providers(&provider3.key), vec![]);
     }
+
+    // TODO: test local providers.
```
lexnv, line 760: nit: Maybe add an issue about this if not already?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -1101,6 +1125,31 @@ impl Kademlia {
                         None => return Err(Error::EssentialTaskClosed),
                     }
                 },
+                action = self.store.next_action() => match action {
```
lexnv, line 1128: One other super small thingie not for this PR, we can try to move away from writing code in `tokio::select` since that breaks our indentation (ie use a `fn hande_store_action(..)` ). We can come later and tidy-up all other places

lexnv (approved):
LGTM! Nice work here! 🙏

src/protocol/libp2p/kademlia/config.rs
dmitry-markin: The same)

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: It's already addresses in the providers unit tests branch. I will remove this TODO in the upcoming PR ("part 6").

PR #244, opened 2024-09-13
req-resp: Refactor to move functionality to dedicated methods

lexnv:
This PR refactors the `RequstResponse::run()` method to move away from writing code in `loop { tokio::select! }`.
Instead, the functionality is moved to dedicated functions.

While at it, have remove the `async` from functions that were plain sync and added comments to make the polling a bit clearer.

cc @paritytech/networking


dmitry-markin (approved):
It is indeed much nicer this way!

PR #236, opened 2024-09-03
kad: Providers part 5: `GET_PROVIDERS` query

dmitry-markin:
Implement iterative `GET_PROVIDERS` query.

Builds upon https://github.com/paritytech/litep2p/pull/235.

### Follow-up
 - Extend unit and integration tests of content providers subsystem of Kademlia.
 - Perform local / testnet tests.
 - Make sure tests do not reveal issues and merge together with other PRs in the series.

PR #245, opened 2024-09-13
kad: Providers part 6: stop providing

dmitry-markin:
Allow removing a local content provider and stopping refreshing it. This PR should make the implementation of DHT content providers in litep2p feature-complete.

Builds upon https://github.com/paritytech/litep2p/pull/236.


src/protocol/libp2p/kademlia/store.rs
```diff
@@ -261,6 +261,46 @@ impl MemoryStore {
         }
     }
 
+    /// Remove local provider for `key`.
+    pub fn remove_local_provider(&mut self, key: Key) {
+        if self.local_providers.remove(&key).is_none() {
+            tracing::warn!(?key, "trying to remove nonexistent local provider",);
```
lexnv, line 267: nit: Could remove the trailing comma here and below?

src/protocol/libp2p/kademlia/handle.rs
```diff
@@ -364,6 +370,15 @@ impl KademliaHandle {
         query_id
     }
 
+    /// Stop providing the key on the DHT.
+    ///
+    /// This will stop republishing the provider, but won't
+    /// remove it instantly from the nodes. It will be removed from them after the provider TTL
+    /// expires, set by default to 48 hours.
```
lexnv, line 377: dq: Do you think we should consider an eviction strategy from the memory store that removes older but not yet expired records similar to LRU?

lexnv (approved):
LGTM! Nicely done here! 🙏

PR #248, opened 2024-09-20
WIP: manager: Improve address handling and refactor PeerState

lexnv:

PR #250, opened 2024-09-24
address_store: Improve address tracking and add eviction algorithm

lexnv:
This PR improves the transport manager's address tracking to keep a healthier view of addresses.

General changes:
- Fixes a bug where `listener` addresses were tracked instead of `dialing` addresses (incoming connections may be established with ephemeral ports and there's no guarantee the remote is listening on them)
- `PeerIdMismatch` error coming from the noise handshake redirects the address to the appropriate peer for healthier addresses
- Addresses are tracked first, regardless of the peer state to ensure we update our view of the addresses (reachable or not)

Address Store changes:
- The store is bounded to a maximum of 64 tracked addresses
- Removing and reinsertion of addresses into the store is prone to error and instead the address store updates addresses in place
- Introduces an eviction algorithm for tracking addresses with higher score

### Testing Done
- added extra tests to the address store
- tested with subp2p-explorer for discovering kusama

This PR is part of a bigger refactor to keep track of healthier addresses.
It is essentially a breakdown of https://github.com/paritytech/litep2p/pull/248 for an easier review process.


src/transport/manager/mod.rs
```diff
@@ -724,6 +724,49 @@ impl TransportManager {
...
+
+        let score = AddressStore::error_score(error);
+
+        // Check if the address corresponds to a different peer ID than the one we're
+        // dialing. This can happen if the node operation restarts the node.
+        //
+        // In this case the address is reachable, however the peer ID is different.
+        // Keep track of this address for future dials.
+        //
+        // Note: this is happening quite often in practice and is the primary reason
```
dmitry-markin, line 739: The comment seems to be unfinished.

src/transport/manager/mod.rs
```diff
@@ -724,6 +724,49 @@ impl TransportManager {
...
+        if let DialError::NegotiationError(NegotiationError::PeerIdMismatch(_, provided)) = error {
+            let context = peers.entry(*provided).or_insert_with(|| PeerContext::default());
+
+            if !std::matches!(address.iter().last(), Some(Protocol::P2p(_))) {
+                address.pop();
+            }
+            context
+                .addresses
+                .insert(AddressRecord::new(&provided, address.clone(), score, None));
```
dmitry-markin, line 748: May be I don't understand something, but not sure if this is a good idea:
1. If we don't discover the "new" `PeerId` through Kademlia, we won't dial this address anyway.
2. Is it possible to generate `PeerIdMismatch` with arbitrary `PeerId` without cryptographically proving you own the private key? Then malicious actor can poison the address store of arbitrary peer by joining the DHT and advertising random `PeerId`, but on connection from the poisoned node provide the victim's `PeerId` via `PeerIdMismatch`.

src/transport/manager/mod.rs
```diff
@@ -724,6 +724,49 @@ impl TransportManager {
...
+
+            return;
+        }
+
+        // Extract the peer ID at this point to give `NegotiationError::PeerIdMismatch` a chance to
+        // propagate.
```
dmitry-markin, line 754: Can you explain this comment, please? Haven't we already handled the `NegotiationError::PeerIdMismatch` above?

src/transport/manager/address.rs
```diff
@@ -186,52 +209,84 @@ impl AddressStore {
...
+        // This algorithm has the following implications:
+        //  - it keeps the best addresses in the store.
+        //  - if the store is at capacity, the worst address will be evicted.
+        //  - an address that is not dialed yet (with score zero) will be preferred over an address
+        //  that already failed (with negative score).
+        if num_addresses >= self.max_capacity {
+            // No need to keep track of negative addresses if we are at capacity.
+            if record.score < 0 {
+                return;
+            }
```
dmitry-markin, line 264: Wouldn't it better to displace an address with even lower score from the list?

src/transport/manager/address.rs
```diff
@@ -18,12 +18,37 @@
...
+/// Maximum number of addresses tracked for a peer.
+const MAX_ADDRESSES: usize = 64;
+
+/// Scores for address records.
+pub mod scores {
+    /// Score indicating that the connection was successfully established.
+    pub const CONNECTION_ESTABLISHED: i32 = 100i32;
+
+    /// Score for a connection with a peer using a different ID than expected.
+    pub const DIFFERENT_PEER_ID: i32 = 50i32;
```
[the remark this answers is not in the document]
lexnv: Yep makes sense! Have simplified the code kept around 1 connection failure and 1 address failure which punishes more, thanks for the suggestion! 🙏

src/transport/manager/handle.rs
```diff
@@ -180,50 +180,58 @@ impl TransportManagerHandle {
...
+            }
+
+            // Check the peer ID if present.
+            if let Some(Protocol::P2p(multihash)) = address.iter().last() {
+                // Ignore the address if the peer ID is invalid.
+                let Ok(peer_id) = PeerId::from_multihash(multihash.clone()) else {
+                    continue;
+                };
+
+                // This can correspond to the provided peerID or to a different one.
```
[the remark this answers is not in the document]
lexnv: Have added an warning here, good point 🙏

src/transport/manager/handle.rs
```diff
@@ -180,50 +180,58 @@ impl TransportManagerHandle {
...
+                // It is important to keep track of all addresses to have a healthy
+                // address store to dial from.
+                peer_addresses.entry(peer_id).or_insert_with(HashSet::new).insert(address);
+                continue;
+            }
 
-            return 0usize;
+            // Add the provided peer ID to the address.
+            let address = address.with(Protocol::P2p(multihash::Multihash::from(peer.clone())));
```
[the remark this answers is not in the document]
lexnv: Have refactored the `continue` with an `if else` branch, looked a bit confusing indeed.
I think we check that the last element is protocol::p2p, let me know if this is sounds right

src/transport/manager/address.rs
```diff
@@ -18,12 +18,29 @@
...
+    /// Score indicating that the connection was successfully established.
+    pub const CONNECTION_ESTABLISHED: i32 = 100i32;
+
+    /// Score for failing to connect due to an invalid or unreachable address.
+    pub const CONNECTION_FAILURE: i32 = -100i32;
+
+    /// Score for providing an invalid address.
+    ///
+    /// This address can never be reached.
+    pub const ADDRESS_FAILURE: i32 = 5 * CONNECTION_FAILURE;
```
dmitry-markin, line 42: ?
```suggestion
    pub const ADDRESS_FAILURE: i32 = i32::MIN;
```

src/transport/manager/handle.rs
```diff
@@ -180,50 +180,66 @@ impl TransportManagerHandle {
...
+                        ?peer,
+                        ?address,
+                        "Added known address that corresponds to a different peer ID",
+                    );
+                }
 
-            return 0usize;
+                // It is important to keep track of all addresses to have a healthy
+                // address store to dial from.
+                peer_addresses.entry(peer_id).or_insert_with(HashSet::new).insert(address);
```
dmitry-markin, line 213: Why not discard the address with invalid peer ID? If the address with invalid peer ID ended up here, something bad has already happened. In theory, we could rewrite the peer ID part, but ideally we should have returned an error to the user long before coming here with invalid peer ID.

This would also simplify the code below by inserting only one peer in the map.

src/transport/manager/address.rs
```diff
@@ -186,52 +201,63 @@ impl AddressStore {
...
+        // The eviction algorithm favours addresses with higher scores.
+        //
+        // This algorithm has the following implications:
+        //  - it keeps the best addresses in the store.
+        //  - if the store is at capacity, the worst address will be evicted.
+        //  - an address that is not dialed yet (with score zero) will be preferred over an address
+        //  that already failed (with negative score).
+        if num_addresses >= self.max_capacity {
+            let Some(min_record) = self.addresses.values().min().cloned() else {
+                return;
```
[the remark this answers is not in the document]
lexnv: The `num_addresses` is initialized to `self.addresses.len()` 🤔

src/transport/manager/address.rs
lexnv: Yep that makes sense, since the address will not be reached

src/transport/manager/address.rs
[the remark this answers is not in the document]
lexnv: Have removed the `num_addreses` variable

src/transport/manager/handle.rs
lexnv: Oki doki that makes sense. I think I'm convinced that tracking all peer addresses is not the best thing here, since the user explicitly provided addresses of a single "peer". I've discarded addresses that belong to a different peer,this should make the code a bit easier to follow, thanks for the suggestion 🙏

src/transport/manager/address.rs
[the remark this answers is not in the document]
dmitry-markin: Is there a chance of getting `None` here? May be `expect()` with a comment then?

src/transport/manager/handle.rs
```diff
@@ -180,51 +180,67 @@ impl TransportManagerHandle {
...
+                    continue;
+                };
+
+                // This can correspond to the provided peerID or to a different one.
+                if peer_id != *peer {
+                    tracing::warn!(
+                        target: LOG_TARGET,
+                        ?peer,
+                        ?address,
+                        "Added known address that corresponds to a different peer ID",
```
dmitry-markin, line 207:
```suggestion
                        "Refusing to add known address that corresponds to a different peer ID",
```

src/transport/manager/handle.rs
```diff
@@ -180,51 +180,67 @@ impl TransportManagerHandle {
...
+
+            // Check the peer ID if present.
+            if let Some(Protocol::P2p(multihash)) = address.iter().last() {
+                // Ignore the address if the peer ID is invalid.
+                let Ok(peer_id) = PeerId::from_multihash(multihash.clone()) else {
+                    continue;
+                };
+
+                // This can correspond to the provided peerID or to a different one.
+                if peer_id != *peer {
```
dmitry-markin, line 202: nit: We can rewrite this as `multihash != peer.as_ref()` without the need to upcast the multihash above.

Issue #237, opened 2024-09-05
transport/manager: Investigate connection closed for a nonexistent peer


lexnv:
Substrate panic triggered on `debug_assert`: 
```
Thread 'tokio-runtime-worker' panicked at 'assertion failed: false', /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/litep2p-0.6.2/src/protocol/transport_service.rs:220
```

Code path:
https://github.com/paritytech/litep2p/blob/a27d0072913be67cce904e2e8c465cc37ec5223a/src/protocol/transport_service.rs#L212-L222

Substrate issue: https://github.com/paritytech/polkadot-sdk/issues/5595
Grafana link: https://grafana.teleport.parity.io/goto/sR83XkeSg?orgId=1

PR #251, opened 2024-09-24
peer_state: Robust state machine transitions

lexnv:
This PR refactors the peer state. Previously the `AddressStore` and `PeerState` where intertwined. The `AddressStore` contained an optional `ConnectionID`, while the `PeerState` states contained a mandatory score. 

This PR separates the address store and the peer state, while moving the logic of the peer state transitioning to a separate module. While at it, have added documentation and testing for state machine transitions.

Changes include:
- Secondary connection from the PeerContext is merged into the `PeerState`
- Connection Ids are removed from the Address Store
- Transport manager is refactored to use the new state machine, this keeps a minimal code around in the manager 

### Testing Done
- added extra tests to the peer state
- tested with subp2p-explorer for discovering kusama

This builds upon https://github.com/paritytech/litep2p/pull/250 for a bigger refactor effort to track addresses in a healthier 

Aims at improving:
- https://github.com/paritytech/litep2p/issues/239
- https://github.com/paritytech/litep2p/issues/238
- https://github.com/paritytech/litep2p/issues/237


src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+                    connection_id,
+                    transports,
+                };
+                return StateDialResult::Ok;
+            }
+            state => panic!(
+                "unexpected state: {:?} validated by Self::can_dial; qed",
+                state
+            ),
+        }
```
dmitry-markin, line 213:
```suggestion
        match self.can_dial() {
            StateDialResult::Ok => {
                *self = PeerState::Opening {
                    addresses,
                    connection_id,
                    transports,
                };
                StateDialResult::Ok
            }
            reason => reason
        }
```

src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+                *self = Self::Connected {
+                    record: connection,
+                    secondary: None,
+                };
+
+                return true;
+            }
+
+            // Accept the incoming connection.
+            Self::Opening { .. } => {
```
dmitry-markin, line 318: I don't think we can "jump" from `Opening` directly to `Connected` following the transport manager logic. May be add a warning?

src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+    }
+
+    /// Returns `true` if the connection was opened.
+    pub fn on_connection_opened(&mut self, record: ConnectionRecord) -> bool {
+        match self {
+            Self::Opening { .. } => {
+                // TODO: Litep2p did not check previously if the
+                // connection record is valid or not, in terms of having
+                // the same connection ID and the address part of the
+                // address set.
```
dmitry-markin, line 408: May be just add a warning for such cases?

src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+    pub fn from_endpoint(peer: PeerId, endpoint: &Endpoint) -> Self {
+        Self {
+            address: Self::ensure_peer_id(peer, endpoint.address().clone()),
+            connection_id: endpoint.connection_id(),
+        }
+    }
+
+    /// Ensures the peer ID is present in the address.
+    fn ensure_peer_id(peer: PeerId, address: Multiaddr) -> Multiaddr {
+        if !std::matches!(address.iter().last(), Some(Protocol::P2p(_))) {
```
dmitry-markin, line 463: May be overwrite the peer ID and add a warning in case of mismatch to catch potential problems?

src/transport/manager/mod.rs
```diff
@@ -415,6 +411,29 @@ impl TransportManager {
         self.transport_manager_handle.add_known_address(&peer, address)
     }
 
+    /// Return multiple addresses to dial on supported protocols.
+    fn open_addresses(addresses: &[Multiaddr]) -> HashMap<SupportedTransport, Vec<Multiaddr>> {
```
dmitry-markin, line 415: Nit:
```suggestion
    fn supported_transports_addresses(addresses: &[Multiaddr]) -> HashMap<SupportedTransport, Vec<Multiaddr>> {
```

src/transport/manager/mod.rs
```diff
@@ -430,157 +449,50 @@ impl TransportManager {
...
 
-        for (address, record) in &mut records {
-            record.set_connection_id(connection_id);
+        // Dialing addresses will succeed because the `context.state.can_dial()` returned `Ok`.
+        let result = context.state.dial_addresses(
+            connection_id,
+            dial_addresses.iter().cloned().collect(),
+            transports.keys().cloned().collect(),
+        );
+        assert_eq!(result, StateDialResult::Ok);
```
dmitry-markin, line 484: Looks dangerous...

src/transport/manager/mod.rs
```diff
@@ -648,77 +567,44 @@ impl TransportManager {
...
             let mut peers = self.peers.write();
 
-            match peers.entry(remote_peer_id) {
-                Entry::Occupied(occupied) => {
-                    let context = occupied.into_mut();
+            let context = peers.entry(remote_peer_id).or_insert_with(|| PeerContext::default());
 
-                    context.addresses.insert(record.clone());
+            // Keep the provided record around for possible future dials.
+            context.addresses.insert(address_record.clone());
```
dmitry-markin, line 593: Won't this lead to duplicate records if we repeatedly dial an address?

src/transport/manager/mod.rs
```diff
@@ -1105,239 +795,55 @@ impl TransportManager {
...
-                            dial_record: None,
-                        },
-                        addresses: AddressStore::new(),
-                        secondary_connection: None,
-                    },
-                );
+                // since an inbound connection was removed, the outbound connection can be
+                // removed from pending dials
+                //
+                // TODO: This may race in the following scenario:
```
dmitry-markin, line 833: Can't understand now, is it difficult to fix / how bad is it? Do we need an issue for this?

src/transport/manager/mod.rs
```diff
@@ -909,140 +700,40 @@ impl TransportManager {
...
         self.connection_limits.on_connection_closed(connection_id);
 
         let mut peers = self.peers.write();
-        let Some(context) = peers.get_mut(&peer) else {
+        let context = peers.entry(peer).or_insert_with(|| PeerContext::default());
+
+        let previous_state = context.state.clone();
+        let connection_closed = context.state.on_connection_closed(connection_id);
+
+        if context.state == previous_state {
```
dmitry-markin, line 714: Is this the same as checking for `connection_closed`?

src/transport/manager/mod.rs
```diff
@@ -1360,107 +866,83 @@ impl TransportManager {
...
+            ..
+        } = previous_state
+        else {
             tracing::warn!(
                 target: LOG_TARGET,
                 ?peer,
                 ?connection_id,
-                "connection opened but peer doesn't exist",
+                state = ?context.state,
+                "State missmatch in opening expected by peer state transition",
```
dmitry-markin, line 904:
```suggestion
                "State mismatch in opening expected by peer state transition",
```

dmitry-markin (approved):
With the state machine extracted into `PeerState` the code indeed easier to comprehend :tada: 

I will also need to review https://github.com/paritytech/litep2p/pull/250 before you are able to merge both PRs into master.

src/transport/manager/peer_state.rs
lexnv: We can still get from `Opening` to `Connected` in the following scenario:
- we are dialing on multiple addresses the remote (opening state outbound)
- the remote establishes a connection with us (inbound)

For this situation, we then cancel the opening futures from the transport manager.

Have added a log trace here and some more state machine documentation, thanks 🙏

src/transport/manager/mod.rs
lexnv: Yep, have removed the assert-eq, thanks 🙏

src/transport/manager/mod.rs
lexnv: With the refactored address store we can update address scores without duplicates 🙏

src/transport/manager/mod.rs
lexnv: Hmm yep, with one exception: We might have a connected state with a secondary connection `Sefl::Connected { record, secondary: Some(...) }` and in that case we'll transition to the connected state without a secondary

src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+            address.with(Protocol::P2p(
+                Multihash::from_bytes(&peer.to_bytes()).expect("valid peer id"),
+            ))
```
dmitry-markin, line 493:
```suggestion
            address.with(Protocol::P2p(peer.as_ref()))
```

src/transport/manager/peer_state.rs
```diff
@@ -0,0 +1,928 @@
...
+        } else {
+            address
```
dmitry-markin: There is still a chance we can get here with a different peer ID. I would replace the last component of multiaddr with a provided `peer` in this case and log a warning.

dmitry-markin: I somehow added this comment to an old revision, please ignore.

lexnv:
Double checked with subp2p-explorer bench-cli, everything looks good, thanks for the review 🙏

PR #254, opened 2024-09-25
transport_service: Improve logs and move code from tokio::select macro

lexnv:
Tiny PR to enhance the transport service logs with protocol and endpoints.

While at it, have refactored the code to move away from the tokio select and use dedicated fns.


dmitry-markin (approved):
Nicely done!

PR #255, opened 2024-09-25
transport: Abort canceled dial attempts for TCP, WebSocket and Quic

lexnv:
The `TransportManager` initiates a dialing process on multiple addresses and multiple transports (ie TCP WebSocket).
The first established connection is reported back from the Transport layer to the `TransportManger`.
Then, the `Transport Manager` cancels all ongoing dialing attempts on the remaining layers.

Previously, cancelling implies storing a `ConnectionID` to a HashSet.
This has the downside that all ongoing dial attempts are polled to completion.

In this PR, the futures that establish socket connections are now aborted directly.

### Example
```bash

Manager ---> TCP (dial address A, B, C)
        ---> WebSocket ( dial address D, E, F)
```

T0. Manager initiates dialing on A, B, C for TCP and D, E, F on WebSocket
T1. Established socket connection on address A from TCP (B and C are dropped)
T2. Manager cancels D, E, F from WebSocket

### Before

T2 implies adding a connectionID to a hashset:
https://github.com/paritytech/litep2p/blob/14dc4cc133e4c09f06b75119970583973a8353f0/src/transport/tcp/mod.rs#L518-L519

The worst case scenario:
- wait for all D, E, F to establish connection, then emit back event to manager if the connection ID was not canceled
The best case scenario:
- wait for one of D, E, F to establish connection, then emit back event to manager if the connection ID was not canceled

https://github.com/paritytech/litep2p/blob/d50ec1014479d9a49aabbb9ae1c8587b702d0314/src/transport/tcp/mod.rs#L536-L542

### After

The future that handles dialing is abortable. This way, we don't have to wait for (worst case) all addresses to fail to connect, or (best case) one address to connect.


dmitry-markin (approved):
Nicely done)

PR #256, opened 2024-09-26
yamux: Update yamux with window tuning and backport fixes

lexnv:
This PR updates the internal yamux crate to include the following: 
- yamux window auto tuning similar to QUIC
- Wake up readers to not miss EOF
- pending frames are replace with pending read and write frame for propagating backpressure

Updates are backported from upstream to work with the current `yamux::Control` and `yamux::ControlledConnection`.
The `Control` and `ControlleConnection` are used by litep2p and have been removed by upstream. This will lead to a major litep2p refactor which may take significantly longer than backporting.


dmitry-markin:
Can you remind why we carry our own yamux fork with litep2p? I remember some functionality was removed from the upstream?

May be we can extract yamux as `litep2p-yamux` and try keeping it in sync with upstream, only with the needed patch applied? Backporting every yamux change in-tree of litep2p seems to be a lot of work, and they will diverge sooner or later anyway.

altonen:
Yamux was included in litep2p because there was a memory leak which had been fixed in a later version but that version also had quite significant API breakage which would've required at least a partial rewrite of TCP/WebSocket connections handlers. More context here: https://github.com/paritytech/litep2p/pull/55

src/yamux/connection.rs
```diff
@@ -801,6 +801,10 @@ impl<T: AsyncRead + AsyncWrite + Unpin> Active<T> {
             shared.increase_send_window_by(frame.header().credit());
             if is_finish {
                 shared.update_state(self.id, stream_id, State::RecvClosed);
+
+                if let Some(w) = shared.reader.take() {
+                    w.wake()
+                }
```
dmitry-markin, line 824: Was this backported from the upstream yamux?

src/yamux/connection/flow_control.rs
dmitry-markin, line 1: Was https://github.com/paritytech/litep2p/pull/256/commits/066dae3bea6375dd68249109e8f477a8a2d20be6 cherry picked completely from upstream yamux? If not, what parts were updated manually?

src/yamux/connection.rs
dmitry-markin, line 1: Was https://github.com/paritytech/litep2p/pull/256/commits/8301fffd52e62b9bfc960d46a765018b8b4cec38 cherry-picked from the upstream? What parts were manually modified?

lexnv:
Closing in favor of:
- https://github.com/paritytech/litep2p/pull/320

PR #259, opened 2024-09-30
Fix `mdns_and_ping` example filename

kaichaosun:
The filename of examples/mnds_and_ping is incorrect.


dmitry-markin (approved):

dmitry-markin:
Thanks for catching!

Issue #253, opened 2024-09-25
TransportService: Improve connection stability by downgrading connections on substream inactivity


lexnv:
### Context

The transport service handler intent is to downgrade connections (leading to possibly closing them), if substreams where not open in a given timeframe: 

https://github.com/paritytech/litep2p/blob/d50ec1014479d9a49aabbb9ae1c8587b702d0314/src/protocol/transport_service.rs#L126-L130

When a connection is established, the timeout is properly tracked:

https://github.com/paritytech/litep2p/blob/d50ec1014479d9a49aabbb9ae1c8587b702d0314/src/protocol/transport_service.rs#L213-L216

When the timeout expires, the connection is downgraded:
https://github.com/paritytech/litep2p/blob/d50ec1014479d9a49aabbb9ae1c8587b702d0314/src/protocol/transport_service.rs#L414-L427

### Issue

Opening of substreams are not taken into account for downgrading connections:

https://github.com/paritytech/litep2p/blob/d50ec1014479d9a49aabbb9ae1c8587b702d0314/src/protocol/transport_service.rs#L410

Here we should match for `Some(InnerTransportEvent::SubstreamOpened` and move forward the timeout future.

With the current implementation, we downgrade connections at 5 seconds intervals.

### Implications

We need to double-check if the protocol name matches the substream open protocol.
Further, `InnerTransportEvent::SubstreamOpened` needs to be extended with a `connectionID` because the `TransportService` holds up to 2 connections (primary and secondary). We should properly advance the timeout of the connection ID.

PR #260, opened 2024-09-30
transport_service: Improve connection stability by downgrading connections on substream inactivity

lexnv:
This PR advances the keep-alive timeout of the transport service.

Previously, the keep-alive timeout was triggered 5 seconds after the connection was reported to the transport service regardless of substream activity.

- (0secs) T0: connection established; keep-alive timeout set to 5seconds in the future
- (4secs) T1: substream A, B, C opened
- (5secs) T2: keep-alive timeout triggered and the connection is downgraded. T1 was not taken into account, otherwise, the keep-alive timeout should be triggered at second 9 (T1 at 4 seconds + keepalive 5 seconds)
- (6secs) T3: substreams A, B, C closed -> connection closes
- (7secs) T4: cannot open new substreams even if we expected the connection to be kept alive for longer

In this PR:
- `KeepAliveTracker` structure to forward the keep-alive timeout of connections.
- Connection ID is forwarded to `SubstreamOpened` events to identify properly substream Ids. This is needed because the `ConnectionContext` contains up to two connections (primary and secondary)

### Testing Done
- test to ensure keepalive downgrades the connection after 5 seconds
- test to ensure keepalive is forwarded on substream activity
- test to ensure a downgraded connection with dropped substreams is closed

Closes https://github.com/paritytech/litep2p/issues/253.


lexnv (commented):
Will have another pass over `KeepAliveTracker`. Im not entirely happy at the moment that we need a `service` notification to advance the polling here, will have to see how this plays with low-connectivity environments

src/protocol/transport_service.rs
```diff
@@ -96,6 +96,167 @@ impl ConnectionContext {
...
+                        peer = ?key.0,
+                        connection_id = ?key.1,
+                        "Last activity no longer tracks the connection (closed event triggered)",
+                    );
+
+                    // We have effectively ignored this `Poll::Ready` event. To prevent the
+                    // future from getting stuck, we need to tell the executor to poll again
+                    // for more events.
+                    cx.waker().wake_by_ref();
+                    return Poll::Pending;
```
lexnv, line 218: A simplified example in the rust playground where I've reproduced this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=182bf56449de7827c8f83b649dbba4fb

src/protocol/transport_service.rs
```diff
@@ -96,6 +96,167 @@ impl ConnectionContext {
...
+        if self.pending_keep_alive_timeouts.is_empty() {
+            // No pending keep-alive timeouts.
+            self.waker = Some(cx.waker().clone());
+            return Poll::Pending;
+        }
+
+        match self.pending_keep_alive_timeouts.poll_next_unpin(cx) {
+            Poll::Ready(Some(key)) => {
+                // Check last-activity time.
+                let Some(when) = self.last_activity.get(&key) else {
```
dmitry-markin, line 206: Nit for easier understanding of what time point `when` is referring to:
```suggestion
                let Some(last_activity) = self.last_activity.get(&key) else {
```

src/protocol/transport_service.rs
```diff
@@ -96,6 +96,167 @@ impl ConnectionContext {
...
+                    // We have effectively ignored this `Poll::Ready` event. To prevent the
+                    // future from getting stuck, we need to tell the executor to poll again
+                    // for more events.
+                    cx.waker().wake_by_ref();
+                    return Poll::Pending;
+                };
+
+                // Keep-alive timeout not reached yet.
+                if when.elapsed() < self.keep_alive_timeout {
+                    let timeout = self.keep_alive_timeout - when.elapsed();
```
dmitry-markin, line 223: Nit:
```suggestion
                let inactive_for = last_activity.elapsed();
                if inactive_for < self.keep_alive_timeout {
                    let timeout = self.keep_alive_timeout - inactive_for;
```

src/protocol/transport_service.rs
```diff
@@ -96,6 +96,167 @@ impl ConnectionContext {
...
+
+                    tracing::debug!(
```
dmitry-markin, line 225: I guess this is going to be triggered quite often, given the number of connections / peers.
```suggestion
                    tracing::trace!(
```
I would may be removed this log completely after initial testing confirms it works as expected.

src/protocol/transport_service.rs
```diff
@@ -96,6 +96,167 @@ impl ConnectionContext {
...
+                    }));
+
+                    // This is similar to the `last_activity` check above, we need to inform
+                    // the executor that this object may produce more events.
+                    cx.waker().wake_by_ref();
+                    return Poll::Pending;
+                }
+
+                // Keep-alive timeout reached.
+                tracing::trace!(
```
dmitry-markin, line 246: nit: This one seems to be less frequent.
```suggestion
                tracing::debug!(
```

dmitry-markin (approved):
LGTM!

It looks like we don't track actual data transmission when rescheduling keep alive timeouts, do we? Then, is it possible we downgrade the connections and disconnect peers if no requests are sent/received, as Notifications protocol works over long-lived substreams and doesn't open new ones?

src/protocol/transport_service.rs
lexnv: Nice catch! Otherwise we can overflow if the second invocation of `when.elapsed()` is advanced 🙏

lexnv:
> Then, is it possible we downgrade the connections and disconnect peers if no requests are sent/received, as Notifications protocol works over long-lived substreams and doesn't open new ones?

Yep absolutely! Offhand, the easiest way may be to use `BandwidthSink` to expose more info, will work on this as a followup to unblock this PR, thanks!

PR #273, opened 2024-10-23
identify: Fix memory leak of unused `pending_opens`

lexnv:
The identify protocol implementation leaked SubstreamIds and PeerIds via the `pending_opens` hashMap.
Objects were only inserted in the `pending_opens`, however they were never removed.

The only possible purpose of `pending_opens` is to double-check the events coming from the service layer: `TransportEvent::SubstreamOpened`. However this is not needed, as illustrated by the current implementation. 

Part of endeavors to fix memory leaks: https://github.com/paritytech/polkadot-sdk/issues/6149

### Testing Done

- custom patched litep2p to dump the internal state of identify protocol running in kusama

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/identify.rs
+++ b/src/protocol/libp2p/identify.rs
@@ -181,9 +181,6 @@ pub(crate) struct Identify {
     protocols: Vec<String>,
 
     /// Pending outbound substreams.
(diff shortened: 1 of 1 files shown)


dmitry-markin (approved):
Nice catch!

Issue #231, opened 2024-09-03
kad: Include a timeout for `QueryExecutor::send_message` 


lexnv:
Sending a plain message using the query executor does not take into account timeouts:

https://github.com/paritytech/litep2p/blob/e3a22d5cfbf255e74be1acceb3ea6ec0baa43c09/src/protocol/libp2p/kademlia/executor.rs#L136-L139

The other methods of the query executor bound the total time to the `READ_TIMEOUT` constant:
https://github.com/paritytech/litep2p/blob/e3a22d5cfbf255e74be1acceb3ea6ec0baa43c09/src/protocol/libp2p/kademlia/executor.rs#L161-L166

Introduce a new variable `WRITE_TIMEOUT` and ensure writes are timedout as well.

PR #277, opened 2024-10-29
kad/executor: Add timeout for writting frames

lexnv:
This PR adds a `WRITE_TIMEOUT` for sending frames to remote peers.

The `WRITE_TIMEOUT` is set to 15 seconds to mirror the `READ_TIMEOUT`.

Closes: https://github.com/paritytech/litep2p/issues/231

### DIFF
--- a/src/protocol/libp2p/kademlia/executor.rs
+++ b/src/protocol/libp2p/kademlia/executor.rs
@@ -35,6 +35,8 @@ use std::{
 
 /// Read timeout for inbound messages.
 const READ_TIMEOUT: Duration = Duration::from_secs(15);
+/// Write timeout for outbound messages.
+const WRITE_TIMEOUT: Duration = Duration::from_secs(15);
 
 /// Query result.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/protocol/libp2p/kademlia/executor.rs
```diff
@@ -91,16 +93,24 @@ impl QueryExecutor {
     /// Send message to remote peer.
     pub fn send_message(&mut self, peer: PeerId, message: Bytes, mut substream: Substream) {
         self.futures.push(Box::pin(async move {
-            match substream.send_framed(message).await {
-                Ok(_) => QueryContext {
+            match tokio::time::timeout(WRITE_TIMEOUT, substream.send_framed(message)).await {
```
dmitry-markin, line 96: Not sure if you already considered this, but a timeout is typically not needed for write operations, because `send_framed(message)).await` will return an error if ACK is not received for sent data on the stream level.

On the other hand, the timeout can complicate sending bigger payloads on slow connections, so I would be extra careful when adding it.

dmitry-markin: Thinking about it again, it should be safe to add a 15-sec timeout for writing if we already have a 15-sec timeout for reading :thinking:

lexnv: I think I missed that ACK on the stream level implementation 🤔 Yep, I would keep the timeout more as a safety net in case we do anything wrong on the lower level implementations 🙏

PR #278, opened 2024-10-30
crypto/noise: Verify crypto/noise signature payload

lexnv:
This PR verifies at the litep2p level the `NoiseHandshakePayload` crypto/noise's signature.

### DIFF
--- a/src/crypto/noise/mod.rs
+++ b/src/crypto/noise/mod.rs
@@ -832,7 +832,7 @@ mod tests {
 
     #[test]
     fn invalid_peer_id_schema() {
-        match parse_peer_id(&vec![1, 2, 3, 4]).unwrap_err() {
+        match parse_peer_id(&vec![1, 2, 3, 4], &[0]).unwrap_err() {
             NegotiationError::ParseError(_) => {}
             _ => panic!("invalid error"),
         }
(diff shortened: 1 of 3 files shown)


src/crypto/noise/mod.rs
```diff
@@ -656,17 +670,35 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncWrite for NoiseSocket<S> {
...
+        payload.identity_sig.ok_or(NegotiationError::BadSignature).map_err(|err| {
+            tracing::debug!(target: LOG_TARGET, "payload without signature");
+            err
+        })?;
+
+    let peer_id = PeerId::from_public_key(&remote_public_key);
+
+    if !remote_public_key.verify(
+        &[STATIC_KEY_DOMAIN.as_bytes(), keypair_public].concat(),
+        &remote_key_signature,
```
dmitry-markin, line 690: I am a little bit confused: if `remote_public_key` is a network public key of a remote peer, then what is `keypair_public` and why we don't check them to be equal?

lexnv: Will have a look, we can include this in the next release 👍

dmitry-markin:
@lexnv thank you for the explanation and the links to the docs!

Issue #282, opened 2024-11-08
transport-manager: Too many file descriptors crash


lexnv:
### Issue
Litep2p crashes after 4/5 hours when tested with a high number of inbound / outbound connections (500 in Kusama).

```
2024-11-07 21:03:37.155 ERROR tokio-runtime-worker grandpa: GRANDPA voter error:
could not complete a round on disk:
Database error: IO error: While open a file for appending: ../dbs/2024-11-04_07-55-18-9e0cebad9f60d62c72c933340c8a3f9e3af63c76-litep2p/chains/ksmcc3/db/full/591987.log:

Too many open files   
```

### Investigation

This may happen because  both TCP and WebSocket transports are polling the listener socket which produces sockets (file descriptors) that await negotiation:

https://github.com/paritytech/litep2p/blob/af0535cff63733c68908ef866d1f6e2e2b701bbe/src/transport/tcp/mod.rs#L526-L530

The transport manager looks like its not consuming the events: 

```bash
2024-11-07 20:13:39.554  INFO tokio-runtime-worker litep2p::websocket: pending_inbound_connections=0

2024-11-07 20:13:59.556  INFO tokio-runtime-worker litep2p::websocket: pending_inbound_connections=217
```

In the span of 20 seconds we received roughly 1653 pending inbound and handled 1436. 
![Screenshot 2024-11-08 at 14 52 17](https://github.com/user-attachments/assets/e5c986c9-ae15-4d04-bbdb-180d52423a62)
![Screenshot 2024-11-08 at 15 04 06](https://github.com/user-attachments/assets/a9eb3ac1-5713-4977-9660-806fb80c9000)

### Possible Solutions

- Don't poll the socket listener (TCP and WebSocket) if we get above a number of `pending_inbound_connections` configurable by TcpConfig and WebSocketConfig
- Ensure TCP and WebSocket are robust wrt "too many open files" errors from the listener
- investigate further polling of futures (transports / transport-manager): it looks offhand like we were able to sustain this pace in the past

lexnv:
The first graph highlights litep2p supporting many inbound connections when substrate is configured for 500/1k in peers.
The second graph shows a memory leak in our tracking of connections which is included in one of our latest releases.

This does not reproduce anymore. I suspect this was an artifact of having multiple polkadot workspaces opened with vscode and rustanalizer, while running 2 substrates nodes on the same machine.

PR #283, opened 2024-11-11
manager: Avoid overflow on stream implementation for `TransportContext`

lexnv:
This PR ensures that the stream implementation of `TransportContext` does not overflow.
Instead, this PR ensures a round-robin polling strategy with the index capped at the number of elements registered to the `TransportContext`. 

While at it, added a test to ensure polling functionality works with round-robin expectations. 

Discovered during: https://github.com/paritytech/litep2p/issues/282

cc @paritytech/networking

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -1322,6 +1322,152 @@ mod tests {
         (dial_address, connection_id)
     }
 
+    struct MockTransport {
+        rx: tokio::sync::mpsc::Receiver<TransportEvent>,
+    }
+
+    impl MockTransport {
+        fn new(rx: tokio::sync::mpsc::Receiver<TransportEvent>) -> Self {
+            Self { rx }
+        }
+    }
+
+    impl Transport for MockTransport {
+        fn dial(&mut self, _connection_id: ConnectionId, _address: Multiaddr) -> crate::Result<()> {
+            Ok(())
+        }
+
+        fn accept(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+            Ok(())
+        }
+
+        fn accept_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+            Ok(())
+        }
+
+        fn reject_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+            Ok(())
+        }
+
+        fn reject(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+            Ok(())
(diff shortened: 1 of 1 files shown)


src/transport/manager/mod.rs
```diff
@@ -170,27 +170,31 @@ impl Stream for TransportContext {
...
-            self.index += 1;
+        if self.transports.is_empty() {
+            // Terminate if we don't have any transports installed.
+            return Poll::Ready(None);
+        }
 
-            let (key, stream) = self.transports.get_index_mut(index).expect("transport to exist");
+        let len = self.transports.len();
+        for index in 0..len {
+            let current = (self.index + index) % len;
```
dmitry-markin, line 180: nit: what about updating `self.index` before the loop to not care about it in every return branch? In the end, it doesn't matter if we store the last index or the new one between the `poll_next` calls.

```suggestion
        let len = self.transports.len();
        self.index = (self.index + 1) % len;
        for index in 0..len {
            let current = (self.index + index) % len;
```

src/transport/manager/mod.rs
```diff
@@ -170,27 +170,31 @@ impl Stream for TransportContext {
...
             match stream.poll_next_unpin(cx) {
                 Poll::Pending => {}
-                Poll::Ready(None) => return Poll::Ready(None),
-                Poll::Ready(Some(event)) => return Poll::Ready(Some((*key, event))),
-            }
-
-            if self.index == start_index + len {
-                break Poll::Pending;
+                Poll::Ready(None) => {
+                    self.index = (self.index + 1) % len;
```
dmitry-markin, line 185:
```suggestion
```

src/transport/manager/mod.rs
```diff
@@ -170,27 +170,31 @@ impl Stream for TransportContext {
...
+                    return Poll::Ready(None);
+                }
+                Poll::Ready(Some(event)) => {
+                    let event = Some((*key, event));
+                    self.index = (self.index + 1) % len;
```
dmitry-markin, line 190:
```suggestion
```

src/transport/manager/mod.rs
```diff
@@ -170,27 +170,31 @@ impl Stream for TransportContext {
...
+                    return Poll::Ready(event);
+                }
             }
         }
+
+        self.index = (self.index + 1) % len;
```
dmitry-markin, line 196:
```suggestion
```

dmitry-markin (approved):

PR #284, opened 2024-11-11
manager: Log when polling returns Ready(None)

lexnv:
This tiny PR logs when the receiving handlers return none, indicating that they shouldn't be polled further:
- installed protocols
- user commands
- installed transports


dmitry-markin (approved):
LGTM! Does it log these errors every time the node is stopped, or the termination rder is such that the receiver is shut down first?

lexnv:
I've not seen these logs triggered yet from my testing, I think there's a small possibility of them happening tho 🤔

PR #285, opened 2024-11-11
transports: Fix missing Poll::Ready(None) event from listenener

lexnv:
This PR handles the following:
- Align Webscoket listener with TCP listener to not miss out on `Poll::Ready(None)` events and enhance logging

- Add warnings whenever pending connection IDs do not exist

- ~~Filtering out `Poll::Ready` events into `Poll::Pending` without calling the context waker will result in the scheduler not polling in the future. This was happening, for example, for both TCP and Websocket when receiving a connection established event that was previously canceled~~

cc @paritytech/networking

Discovered during: https://github.com/paritytech/litep2p/issues/282

### DIFF
--- a/src/transport/tcp/mod.rs
+++ b/src/transport/tcp/mod.rs
@@ -570,7 +570,14 @@ impl Stream for TcpTransport {
             };
         }
 
+        // Whenever we are receiving events by `Poll::Ready` and we choose to not propagate
+        // them to the higher levels, we should wake up the context to poll us again.
+        // Otherwise, the scheduler will not know that we are ready to be polled again.
+        let mut should_wake_up = false;
+
         while let Poll::Ready(Some(result)) = self.pending_raw_connections.poll_next_unpin(cx) {
+            should_wake_up |= true;
+
             tracing::trace!(target: LOG_TARGET, ?result, "raw connection result");
 
             match result {
@@ -633,6 +640,8 @@ impl Stream for TcpTransport {
         }
 
         while let Poll::Ready(Some(connection)) = self.pending_connections.poll_next_unpin(cx) {
+            should_wake_up |= true;
+
             match connection {
                 Ok(connection) => {
                     let peer = connection.peer();
@@ -657,6 +666,11 @@ impl Stream for TcpTransport {
                     }
                 }
             }
+        }
+
+        // We have filtered out all `Poll::Ready` events.
+        if should_wake_up {
+            cx.waker().wake_by_ref();
         }
 
         Poll::Pending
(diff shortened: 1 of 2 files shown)


src/transport/tcp/mod.rs
```diff
@@ -554,9 +584,16 @@ impl Stream for TcpTransport {
             };
         }
 
+        // Whenever we are receiving events by `Poll::Ready` and we choose to not propagate
+        // them to the higher levels, we should wake up the context to poll us again.
+        // Otherwise, the scheduler will not know that we are ready to be polled again.
+        let mut should_wake_up = false;
```
lexnv, line 590: I think this should be reverted, ideally either `self.listener` / `self.pending_raw_connections` / `self.pending_connections` would save the waker.
Wdyt? @dmitry-markin 🙏

src/transport/tcp/mod.rs
```diff
@@ -529,7 +543,23 @@ impl Stream for TcpTransport {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         if let Poll::Ready(event) = self.listener.poll_next_unpin(cx) {
             return match event {
-                None | Some(Err(_)) => Poll::Ready(None),
+                None => {
+                    tracing::error!(
+                        target: LOG_TARGET,
+                        "TCP listener terminated, ignore if the node is stopping",
+                    );
```
dmitry-markin, line 550: Is this error actually ever emitted?

src/transport/tcp/mod.rs
dmitry-markin: It doesn't seem we swallow any pending events from `self.listener`? As for `self.pending_raw_connections` / `self.pending_connections` we better save and trigger a waker only when new connections are added, and not in every `poll_next` call.

src/transport/websocket/mod.rs
```diff
@@ -683,6 +728,11 @@ impl Stream for WebSocketTransport {
             }
         }
 
+        // We have filtered out all `Poll::Ready` events.
+        if should_wake_up {
+            cx.waker().wake_by_ref();
+        }
+
```
dmitry-markin, line 735: The same here: we better save and trigger a waker only when adding new elements to `pending(_raw)_connections`.

dmitry-markin (approved):

src/transport/tcp/mod.rs
lexnv: Actually, I'll revert this for now and come with a follow-up. I believe this is needed and that's part of the reason why we may experience connection stability issues  🙏

PR #286, opened 2024-11-11
manager: Fix connection limits tracking of rejected connections 

lexnv:
This PR ensures that connection IDs are properly tracked to enforce connection limits.

After ~1/2 days the substrate litep2p running node is starting to reject multiple pending socket connections.

This is related to the fact that accepting of established connection is a two-part process:
- step 1. Can we accept the connection based on number of connections?
- step 2. The transport manager transitions the peer state
- step 3. If the transition succeeds, the connection is tracked (inserted in a hashmap).

If step 2 fails, we were previously leaking the connection ID. The connection ID is also counting towards the maximum number of connections the node can sustain.

This PR effectively modifies the connection limits API, `on_connection_established` is split into two methods:
  - `can_accept_connection()`
  - `accept_established_connection()`

This fixes a subtle memory leak bounded at 10000 Connection IDs.
However, even more important, this fixes connection stability for long-running nodes.

Discovered during the investigation of:
- https://github.com/paritytech/litep2p/issues/282

cc @paritytech/networking

### DIFF
--- a/src/transport/manager/limits.rs
+++ b/src/transport/manager/limits.rs
@@ -113,9 +113,10 @@ impl ConnectionLimits {
     }
 
     /// Called when a new connection is established.
-    pub fn on_connection_established(
+    ///
+    /// Returns an error if the connection cannot be accepted due to connection limits.
+    pub fn can_accept_connection(
         &mut self,
-        connection_id: ConnectionId,
         is_listener: bool,
     ) -> Result<(), ConnectionLimitsError> {
         // Check connection limits.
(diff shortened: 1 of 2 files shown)


lexnv:
After further testing, it looks like we have duplicate connection IDs or missing out on Established events. 
I would wait a bit until I completely sort this out

lexnv:
Coming back with more data, the PR is ready for review.

The discrepancy I've seen was related to the fact that we don't enable outbound connection limits from substrate. Therefore, the connection limits will count the inbound peers. This was leading to some confusion from my part, as I was expected the limits (66) to reflect the total connections (667): 

```
status limits_in=66 limits_out=0 num_connected=667 num_inbound=66 num_outbound=601
```

dmitry-markin (approved):

PR #287, opened 2024-11-12
transport: Fix waking up on filtered events from `poll_next` 

lexnv:
This PR ensures that the transport (TCP / WebSocket) implementation wakes up the waker `FuturesUnordered` on adding futures to the set.

This is one of the async rough edges in Rust, to properly illustrate the issue I've compiled this rust example (comment and uncomment like 43 to see the issue in action):

### Example 1
The future is not polled again: 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1e801c4864599c54a98bb9f30be1fdd5

### Example 2
The future is delayed (from 2s to 3s) because it relies on some other implementation to wake up the context waker:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=951080892763ca7d02ae9c5617da6c5d

### Particular implementation

The `poll_next` implementation polls 3 objects:
- `listener`: accepts incoming socket connections
- `pending_raw_connections`: dialing futures to establish socket connection
- `pending_connections`: negotiating futures to establish libp2p protocols

When adding futures to either `pending_raw_connections` or `pending_connections`, the futures are not polled immediately. Instead, they will get polled the first time an already present future finishes execution.

This effectively leads to delays in processing both raw connections and pending connections.
This PR in combination with https://github.com/paritytech/litep2p/pull/286 has improved the connection stability of litep2p for a high number of connections (500 in and 500 out). The decreasing line represents release 0.8 configured with 500 in and out peers, while the stable line from 11/12 represents litep2p with incoming PR patches. (I've restarted the node 3/4 times for more debug logs):

Screenshot 2024-11-12 at 14 49 43

### Next Steps

- There are a few other places were this may be happening, I'll have a look at them after we sort out connection limits. 
- @dmitry-markin suggested an even more optimized way of handling this by polling each future individually to feed them a context waker before adding to the inner FuturesUnordered 🙏

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,6 +79,7 @@ pub mod types;
 pub mod yamux;
 
 mod bandwidth;
+mod utils;
 mod multistream_select;
 
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -79,8 +79,8 @@ pub mod types;
 pub mod yamux;
 
 mod bandwidth;
-mod utils;
 mod multistream_select;
+mod utils;
 
 #[cfg(test)]
 mod mock;
--- a/src/protocol/libp2p/kademlia/executor.rs
+++ b/src/protocol/libp2p/kademlia/executor.rs
@@ -19,9 +19,8 @@
 // DEALINGS IN THE SOFTWARE.
 
 use crate::{
-    protocol::libp2p::kademlia::{futures_stream::FuturesStream, query::QueryId},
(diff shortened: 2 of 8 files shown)


lexnv:
I was previously, waking up from `poll_next` instead of only from adding elements to the FuturesUnordered. 
Thanks @dmitry-markin for suggesting a more optimized way of doing this 🙏

dmitry-markin (approved):

Issue #295, opened 2024-12-02
notifications: Investigate possible memory leak with pending outbound substream tracking


lexnv:
The metrics implementation exposed an abnormal behaviour in our state tracking:
- https://github.com/paritytech/litep2p/pull/294

<img width="1482" alt="Screenshot 2024-12-02 at 11 17 17" src="https://github.com/user-attachments/assets/5bd0c453-6956-42cc-9538-8397f59b970a">

It looks like the following data structure grows over time:
https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L261-L262

Elements are added to the hashmap on:
- https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L919
- https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L1093

Elements are removed from the hashmap on:
- https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L514
- https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L816

### Possible causes

It looks like we should remove the state on the connection closed as well:

https://github.com/paritytech/litep2p/blob/28d42c664f6a36a2b132adf06d85f654f283fc16/src/protocol/notification/mod.rs#L445

PR #294, opened 2024-11-29
metrics: Expose litep2p metrics in an agnostic manner

lexnv:
This PR optionally exposes litep2p metrics in an agnostic manner.

### API Design
Litep2p supports at the moment the following primitives to register and operate on metrics:

```rust
/// A registry for metrics.
pub trait MetricsRegistryT: Send + Sync {
    /// Register a new counter.
    fn register_counter(&self, name: String, help: String) -> Result<MetricCounter, Error>;

    /// Register a new gauge.
    fn register_gauge(&self, name: String, help: String) -> Result<MetricGauge, Error>;
}


/// Represents a metric that can only go up.
pub trait MetricCounterT: Send + Sync {
    /// Increment the counter by `value`.
    fn inc(&self, value: u64);
}

/// Represents a metric that can arbitrarily go up and down.
pub trait MetricGaugeT: Send + Sync {
    /// Set the gauge to `value`.
    fn set(&self, value: u64);

    /// Increment the gauge.
    fn inc(&self);
...
}
```

Around these primitives, substrate can expose `prometheus` metrics, and if need be update seamlessly to other metric crates (like `prometheus-client`).

### Metrics Exposed

The exposed metrics inform the user about the state of litep2p components (like kademlia number of store elements, identify negotiating substreams etc) and help developers detect abnormal behavior (like memory leaks / unbounded growth / stalls in some protocols).

#### Transport Manager
  - number of incoming connections
  - number of outgoing connections
  - number of opening errors
  - number of managed peers 
  - number of pending connections

#### Transport Layer (TCP + Websocket)
  - number of pending inbound connections
  - number of pending open connections
  - number of raw unnegotiated connections
  - number of pending substreams

#### Kademlia
  - number of total peers
  - number of engine running queries
  - number of executor queries
  - number of memory store records
  - number of memory store local providers
  - number of memory store providers
  - number of memory store provider refreshes

### Identify / Ping
  - number of total peers
  - number of pending inbound substreams
  - number of pending outbound substreams
  
### Request Response Protocol

These metrics are exposed for every req-resp protocol.
  - number of connected peers
  - number of pending dials
  - number of pending inbound substreams
  - number of pending inbound requests
  - number of pending outbound
  - number of outbound cancels
  - number of pending outbound responses
  
### Notification Protocol
These metrics are exposed for every notification protocol.
  - number of connected peers
  - number of outbound initiated handshakes
  - number of outbound substreams
  - number of validation
  - number of ready substream handshakes
  - number of timers
  
[...]

PR #296, opened 2024-12-02
notification: Fix memory leak of pending substreams

lexnv:
This PR fixes a subtle memory leak that can happen in the following edge-case situation:
- connection is established and substream outbound is initiated with remote peer
- the substream ID is tracked until the substream either completes successfully or fails
- the connection is closed soon after, leading to no substream events ever being generated

For this edge-cases, we need to remove the tracking of the substream ID when the connection is reported as closed.

This has been detected after running a node for more than 2 days with the following generic metrics PR:
- https://github.com/paritytech/litep2p/pull/294

Closes: https://github.com/paritytech/litep2p/issues/295

cc @paritytech/networking

### DIFF
--- a/src/protocol/notification/mod.rs
+++ b/src/protocol/notification/mod.rs
@@ -435,18 +435,30 @@ impl NotificationProtocol {
                             },
                         );
                     }
+                    (OutboundState::OutboundInitiated { substream }, _) => {
+                        tracing::debug!(
+                            target: LOG_TARGET,
+                            ?peer,
+                            protocol = %self.protocol,
+                            "connection closed outbound substream initiated ",
+                        );
+                        // We need to remove this state to avoid a memory leak.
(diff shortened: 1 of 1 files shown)


src/protocol/notification/mod.rs
```diff
@@ -435,18 +435,30 @@ impl NotificationProtocol {
...
+                        self.pending_outbound.remove(&substream);
+
+                        self.event_handle
+                            .report_notification_stream_open_failure(
+                                peer,
+                                NotificationError::Rejected,
+                            )
+                            .await;
+                    }
                     // user either initiated an outbound substream or an outbound substream was
```
bkchr, line 455: Docs need to be updated.

bkchr (approved):

dmitry-markin (approved):

dmitry-markin (approved):

Issue #300, opened 2024-12-10
kad: Failed to register opened substream to protocol


lexnv:
Kusama validator sees a high number of `failed to register opened substream to protocol`, this is cascading to other warnings.

The validator is running for approx 14 days without interruption `kusama-validator-bhs5-0`, other validators don't seem impacted at the moment.

```
Repo            | Count      | Level      | Triage report

https://github.com/paritytech/polkadot-sdk/ | 35638      | warn_if_frequent | Some network error occurred when fetching erasure chunk
https://github.com/paritytech/litep2p/ | 18736      | error      | failed to register opened substream to protocol
https://github.com/paritytech/polkadot-sdk/ | 475        | warn       | Data unavailable for candidate .*
https://github.com/paritytech/polkadot-sdk/ | 475        | warn       | Recovery of available data failed.
https://github.com/paritytech/polkadot-sdk/ | 9          | warn       | fetch_pov_job
https://github.com/paritytech/polkadot-sdk/ | 9          | warn       | Cluster has too many pending statements, something wrong with our connection to our group peers
https://github.com/paritytech/polkadot-sdk/ | 3          | warn       | Report .*: .* to .*. Reason: .*. Banned, disconnecting. ( Same block request multiple times. Banned, disconnecting.)
https://github.com/paritytech/polkadot-sdk/ | 2          | warn       | Report .*: .* to .*. Reason: .*. Banned, disconnecting. ( A collator provided a collation for the wrong para. Banned, disconnecting.)
https://github.com/paritytech/litep2p/ | 1          | error      | failed to register substream open failure to protocol
https://github.com/paritytech/polkadot-sdk/ | 1          | warn       | .*: .* is already a reserved peer
https://github.com/paritytech/litep2p/ | 1          | error      | failed to send mdns query
```

The warnings are related to not being able to propagate open substreams to kademlia in this case:

```
2024-12-10 16:00:00.096 ERROR tokio-runtime-worker litep2p::tcp::connection: failed to register opened substream to protocol protocol=Allocated("/b0a8d493285c2df73290dfb7e61f870f17b41801197a149ca93654499ea3dafe/kad") peer=PeerId("12D3KooWSKhrnPPYmAAfpA8TzE6NLkDooMzm7bApLYk86TNBmfcp") endpoint=Listener { address: "/ip6/2a02:1210:821b:7f00:ce28:aaff:fe0f:2762/tcp/37594", connection_id: ConnectionId(31845931) } error=ConnectionClosed
[...]

PR #302, opened 2024-12-11
identify: Replace FuturesUnordered with FuturesStream

lexnv:
This PR replaces the identify `FuturesUnordered` with `FuturesStream`. This effectively fixes delays in processing outbound events.
- ensure that identify warns if the transport service is closed (produces no events).
- identify no longer exits on pending outbound events

Related to:
- https://github.com/paritytech/litep2p/pull/287
- https://github.com/paritytech/litep2p/issues/300

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/identify.rs
+++ b/src/protocol/libp2p/identify.rs
@@ -394,7 +394,7 @@ impl Identify {
                             .await;
                     }
                     Some(Err(error)) => tracing::debug!(target: LOG_TARGET, ?error, "failed to read ipfs identify response"),
-                    None => return,
+                    _ => {}
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/protocol/libp2p/identify.rs
```diff
@@ -356,7 +357,10 @@ impl Identify {
         loop {
             tokio::select! {
                 event = self.service.next() => match event {
-                    None => return,
+                    None => {
+                        tracing::warn!(target: LOG_TARGET, "transport service stream ended, terminating identify event loop");
```
iulianbarbu, line 361: Is this a warn because a terminated identify loop is not very expected in normal operation?

lexnv: Yep, we don't expect any of these components to exit under normal circumstances. If they exit however, this should give us enough information about what happened to isolate the issue 🙏

PR #303, opened 2024-12-11
ci: Ensure cargo-machete is working with rust version from CI

lexnv:


skunert (approved):

.github/workflows/ci.yml
```diff
@@ -61,6 +61,13 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
 
+      - name: Install Rust stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
```
iulianbarbu, line 68: Do we want to have the same RUST version as the one used in the polkadot-sdk CI? I think we use nightly 2024-11-19 for some things, and 1.81.0 for others. The latest stable is 1.83 I think.

dmitry-markin (approved):

.github/workflows/ci.yml
lexnv: I believe we should work with latest as long as its not an issue for substrate. At the moment, this is just unblocking cargo-machete and should not impact other fmt / clippy rules. Will have a look at this proly next year or so 🙏

iulianbarbu: Shouldn't be any issue I think, given the stables are backward compatible. Some warnings due to deprecation notices might appear but I don't think it is an issue given litep2p is a dependency, and the checks on polkadot-sdk CI apply to the polkadot-sdk codebase only.

PR #315, opened 2025-01-08
kad: Provide partial results to speedup `GetRecord` queries

lexnv:
This PR provides the partial results of the `GetRecord` kademlia query.
- A new `GetRecordPartialResult` event is introduced for kademlia
- `GetRecordSuccess` is modified to include only the query ID
- Kademlia `GetRecord` implementation no longer stores network records internally and forwards valid (unexpired) ones back to the user

The change is needed to speedup authority discovery for substrate based chains.

More context can be found at: https://github.com/paritytech/polkadot-sdk/issues/7077#issuecomment-2575662449

### Next Steps
- [ ] Adjust testing to API breaking change
- [ ] Investigate CPU impact (as suggested by @dmitry-markin this should be unnoticeable 🙏 )

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -485,11 +485,20 @@ impl Kademlia {
 
                         // update routing table and inform user about the update
                         self.update_routing_table(&peers).await;
-                        self.engine.register_response(
-                            query_id,
-                            peer,
-                            KademliaMessage::GetRecord { key, record, peers },
-                        );
+
+                        // Send partial response to the user.
+                        if let Some(QueryAction::GetRecordPartialResult { query_id, record }) =
+                            self.engine.register_response(
+                                query_id,
+                                peer,
+                                KademliaMessage::GetRecord { key, record, peers },
+                            )
+                        {
+                            let _ = self
+                                .event_tx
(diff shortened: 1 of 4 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -485,11 +485,20 @@ impl Kademlia {
...
+                                query_id,
+                                peer,
+                                KademliaMessage::GetRecord { key, record, peers },
+                            )
+                        {
+                            let _ = self
+                                .event_tx
+                                .send(KademliaEvent::GetRecordPartialResult { query_id, record })
+                                .await;
+                        }
```
dmitry-markin, line 501: nit: to keep the structure of event handling, I would let the `engine` to pass `QueryAction::GetRecordPartialResult` to `on_query_action()` and published the outbound event there. The latency won't be noticeable, but the code structure will be more streamlined.

dmitry-markin (approved):
Nice!

src/protocol/libp2p/kademlia/mod.rs
lexnv: Thanks, that makes sense! 🙏 Let me know if the latest changes are oki with you, planning to do a release later today

PR #320, opened 2025-01-23
yamux: Switch to upstream implementation while keeping the controller API

lexnv:
This PR relies on the libp2p-yamux crate for the core functionality of our multiplexer.
The main goal is to bring complete compatibility between libp2p and litep2p on the yamux layer, and remove 90% of the yamux code in favor of the upstream implementation while keeping the controller API in place.

The upstream crate brings in multiple fixes for substreams while minimally impacting our dependency tree.

The downside of the upstream implementation is that the controller API has been removed. Adjusting to the new API would be a massive breaking change for all transport layers. Therefore, we keep the controller API which integrates seamlessly with the upstream yamux. No other changes were present to the controller API in the upstream implementation.

### Yamux changelog

The changelog includes the fixes from the upstream since the moment we have inlined the crate in litep2p:

```
# 0.13.4

- Fix sending pending frames after closing. See [PR 194](https://github.com/libp2p/rust-yamux/pull/194).

# 0.13.3

- Wake up readers after setting the state to RecvClosed to not miss EOF.
  See [PR 190](https://github.com/libp2p/rust-yamux/pull/190).

- Use `web-time` instead of `instant`.
  See [PR 191](https://github.com/libp2p/rust-yamux/pull/191).

# 0.13.2

- Bound `Active`'s `pending_frames` to enforce backpressure. 
  See [460baf2](https://github.com/libp2p/rust-yamux/commit/460baf2ccb7d5982b266cb3cb9c0bdf75b4fb779)
  
# 0.13.1

- Fix WASM support using `instant::{Duration, Instant}` instead of `std::time::{Duration, Instant}`.
  See [PR 179](https://github.com/libp2p/rust-yamux/pull/179).

# 0.13.0

- Introduce dynamic stream receive window auto-tuning.
  While low-resourced deployments maintain the benefit of small buffers, high resource deployments eventually end-up with a window of roughly the bandwidth-delay-product (ideal) and are thus able to use the entire available bandwidth.
  See [PR 176](https://github.com/libp2p/rust-yamux/pull/176) for performance results and details on the implementation.
- Remove `WindowUpdateMode`.
  Behavior will always be `WindowUpdateMode::OnRead`, thus enabling flow-control and enforcing backpressure.
  See [PR 178](https://github.com/libp2p/rust-yamux/pull/178).

# 0.12.1

- Deprecate `WindowUpdateMode::OnReceive`.
  It does not enforce flow-control, i.e. breaks backpressure.
  Use `WindowUpdateMode::OnRead` instead.
  See [PR #177](https://github.com/libp2p/rust-yamux/pull/177).

# 0.12.0

- Remove `Control` and `ControlledConnection`.
  Users have to move to the `poll_` functions of `Connection`.
  See [PR #164](https://github.com/libp2p/rust-yamux/pull/164).

- Fix a bug where `Stream`s would not be dropped until their corresponding `Connection` was dropped.
  See [PR #167](https://github.com/libp2p/rust-yamux/pull/167).

```

### Next Steps
- [x] deployment in versi-net and monitor metrics / CPU impact (extensively test this)

cc @paritytech/networking


Cargo.toml
```diff
@@ -52,8 +52,7 @@ x25519-dalek = "2.0.0"
 x509-parser = "0.16.0"
 yasna = "0.5.0"
 zeroize = "1.8.1"
-nohash-hasher = "0.2.0"
-static_assertions = "1.1.0"
+libp2p-yamux = { version = "0.13.4", package = "yamux" }
```
dmitry-markin, line 55: Why are we renaming `yamux` to `libp2p-yamux`? 'libp2p-yamux` is an existing crate in `libp2p`: https://github.com/libp2p/rust-libp2p/blob/master/muxers/yamux/Cargo.toml

dmitry-markin (approved):
Awesome! This simplifies things a lot! :tada:

altonen:
@lexnv

Did enabling auto-tuning improve performance in any significant way?

lexnv:
Hi, I've only noticed an improvement in our synthetic benchmarks for the notification protocol on `litep2p/with_backpressure/256KB`

- https://github.com/paritytech/polkadot-sdk/pull/7223

```
test notifications_protocol/libp2p/with_backpressure/64KB ... bench:    44405699 ns/iter (+/- 268459)
test notifications_protocol/libp2p/with_backpressure/256KB ... bench:   353299153 ns/iter (+/- 2417926)
 
test notifications_protocol/litep2p/with_backpressure/64KB ... bench:    47341171 ns/iter (+/- 398339)
test notifications_protocol/litep2p/with_backpressure/256KB ... bench:   342768895 ns/iter (+/- 3704768)
```


Before this PR we saw the following:

```
1k messages
notifications_protocol/libp2p/with_backpressure/64KB:    44421256 ns/iter (+/- 309026)
notifications_protocol/libp2p/with_backpressure/256KB:   359526580 ns/iter (+/- 2611825)
notifications_protocol/litep2p/with_backpressure/64KB:    46879477 ns/iter (+/- 327248)
notifications_protocol/litep2p/with_backpressure/256KB:   413041795 ns/iter (+/- 4181491)
```

This reflects a ~20% improvement on this benchmark (before and after). However, I'm not entirely sure about the stability of the data we get out of this. I've not seen anything else in versi-network. CPU-wise, we look to be in the same ballpark. I would still like to run some tests.

The update's main goal was to filter out compatibility issues between libp2p <-> litep2p.
We have discovered on Kusama validators that sometimes libp2p -> litep2p requests would generate a `SubstreamUpgrade::Io` error, and this was triggering an ugly issue in libp2p where libp2p was not taking into account the protocol timeout
- cc https://github.com/paritytech/polkadot-sdk/issues/7076#issuecomment-2596085096

altonen:
Thanks for the info!

I'm surprised that libp2p is faster, considering I spent a quite a bit of time optimizing the notification paths in litep2p and have a recollection of it being faster, albeit in microbenchmarks. Granted the architectures are very different so could be that this is the price litep2p pays for all the message passing it does or my benchmarks were incorrect. AFAIR the receiver side is quite copy-heavy so if these numbers are unacceptable, I think there are still ways to optimize further.

PR #321, opened 2025-02-03
req-resp: Replace SubstreamSet with FuturesStream

lexnv:
This PR replaces the in-house built `SubstreamSet` with the `FuturesStream`.
- provide fairness for substreams that produce elements based on the `FuturesUnordered` implementation
- clone free implementation for peerId and requested
- substream object is readily available once the substream produces a request (ie there's no need to pay for the hashmap element retrieval)

Substreams are no longer removed from the `FuturesStream` and instead they are left to expire with the `ConnectionClosed` event. This has the side-effect of emitting a `debug` log and it simplifies state tracking.

cc @paritytech/networking

### DIFF
--- a/src/protocol/request_response/mod.rs
+++ b/src/protocol/request_response/mod.rs
@@ -27,8 +27,9 @@ use crate::{
         request_response::handle::{InnerRequestResponseEvent, RequestResponseCommand},
         Direction, TransportEvent, TransportService,
     },
-    substream::{Substream, SubstreamSet},
+    substream::Substream,
     types::{protocol::ProtocolName, RequestId, SubstreamId},
+    utils::futures_stream::FuturesStream,
     PeerId,
 };
 
@@ -162,7 +163,17 @@ pub(crate) struct RequestResponseProtocol {
     pending_inbound: FuturesUnordered<BoxFuture<'static, PendingRequest>>,
 
     /// Pending inbound requests.
-    pending_inbound_requests: SubstreamSet<(PeerId, RequestId), Substream>,
+    pending_inbound_requests: FuturesStream<
+        BoxFuture<
+            'static,
+            (
+                PeerId,
+                RequestId,
+                Result<BytesMut, SubstreamError>,
+                Substream,
+            ),
+        >,
+    >,
 
     /// Pending dials for outbound requests.
     pending_dials: HashMap<PeerId, RequestContext>,
@@ -200,7 +211,7 @@ impl RequestResponseProtocol {
             pending_outbound: HashMap::new(),
             pending_inbound: FuturesUnordered::new(),
             pending_outbound_cancels: HashMap::new(),
-            pending_inbound_requests: SubstreamSet::new(),
+            pending_inbound_requests: FuturesStream::new(),
             pending_outbound_responses: FuturesUnordered::new(),
             max_concurrent_inbound_requests: config.max_concurrent_inbound_request,
         }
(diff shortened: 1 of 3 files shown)


src/protocol/request_response/mod.rs
```diff
@@ -1057,9 +1062,9 @@ impl RequestResponseProtocol {
                 _ = self.pending_outbound_responses.next(), if !self.pending_outbound_responses.is_empty() => {}
 
                 // Inbound requests that are moved to `pending_outbound_responses`.
-                event = self.pending_inbound_requests.next() => match event {
-                    Some(((peer, request_id), message)) => {
-                        if let Err(error) = self.on_inbound_request(peer, request_id, message).await {
+                event = self.pending_inbound_requests.next(), if !self.pending_inbound_requests.is_empty() => match event {
+                    Some((peer, request_id, request, substream)) => {
+                        if let Err(error) = self.on_inbound_request(peer, request_id, request, substream).await {
```
dmitry-markin, line 1067: Is there a way to check here if `substream` points to alive substream over an existing connection, and not process the request if not? May be we can do this check in `on_inbound_request`.

src/protocol/request_response/mod.rs
```diff
@@ -315,11 +326,6 @@ impl RequestResponseProtocol {
                 })
                 .await;
         }
-
-        // remove all pending inbound requests
-        for (request_id, _) in context.active_inbound {
-            self.pending_inbound_requests.remove(&(peer, request_id));
-        }
```
dmitry-markin, line 322: So with the new logic we will handle the request anyway if the connection is closed, hitting an error when we try to send a response?

dmitry-markin (approved):

src/protocol/request_response/mod.rs
lexnv: If the connection has been closed, we'll remove the peer-associated data, which will lead to no further polling on the substream:
https://github.com/paritytech/litep2p/blob/4bc31d03a1af5978059e8a571c4276dae5917718/src/protocol/request_response/mod.rs#L302-L307

The inbound request is directly rejected as a first step here:
https://github.com/paritytech/litep2p/blob/4bc31d03a1af5978059e8a571c4276dae5917718/src/protocol/request_response/mod.rs#L477-L489

If the connection has been closed and detected by the substream, we'll return a `Substream::Closed` error which will not propagate the message:
https://github.com/paritytech/litep2p/blob/4bc31d03a1af5978059e8a571c4276dae5917718/src/protocol/request_response/mod.rs#L501-L511

src/protocol/request_response/mod.rs
lexnv: Yep exactly, if the connection gets closed, we'll still have the substream in the FuturesUnordered and let them hit an error (either when we send out the request, or read the response) 🙏

src/protocol/request_response/mod.rs
dmitry-markin: i mean if we received a request and hold it in `pending_inbound_requests`, but the peer gets disconnected before we poll `pending_inbound_requests`, we will process the request anyway and only hit the error when try to send the response to a closed substream? Is there an easy way to not process the inbound request (as it may require resources on the client/runtime side etc.) in this case and fail early?

lexnv: Had a chat offline: we'll not process the inbound requests because we remove the peer state in `on_connection_closed`, have added a comment to make the code easier to follow : https://github.com/paritytech/litep2p/pull/321/commits/58d574266c76f880a0ad4244793f1422eea75a18 🙏

PR #327, opened 2025-02-18
websocket/stream: Fix unexpected EOF on `Poll::Pending` state poisoning 

lexnv:
This PR updates the following dependencies:
- tokio tungstenite
- URL crate to facilitate the new tungstenite version

The tokio tungstenite update changed the API of webscoket messages, which motivated the following changes:
- When`poll_ready` or `poll_flush` returned `Poll::Pending`, the websocket state would forever remain poisoned. 
  - This has the side effect of returning `UnexpectedEof` on consecutive calls, which led to the connection being terminated
- The state machine of the websocket is simplified to just 3 states
- Writer buffer and pointer are merged under one BytesMut instance
- `poll_read` implementation is simplified as well to make the code more readable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -841,6 +841,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dmitry-markin (approved):
Nice catch with a poisoned state being left!

PR #357, opened 2025-03-10
kad: Set upper limits for kademlia messages 

lexnv:
This PR sets the default kademlia message to 70 KiB

The limit can be configurable using the config builder API.

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/config.rs
+++ b/src/protocol/libp2p/kademlia/config.rs
@@ -273,7 +273,7 @@ impl ConfigBuilder {
 
     /// Set the maximum Kademlia message size.
     ///
-    /// If unspecified, the default maximum message size is 16 KiB.
+    ///  Should fit `MemoryStore` max record size. If unspecified, the default maximum message size is 70 KiB.
     pub fn with_max_message_size(mut self, max_message_size: usize) -> Self {
         self.max_message_size = max_message_size;
         self
--- a/src/protocol/libp2p/kademlia/config.rs
+++ b/src/protocol/libp2p/kademlia/config.rs
@@ -52,8 +52,8 @@ const PROTOCOL_NAME: &str = "/ipfs/kad/1.0.0";
 /// Kademlia replication factor.
 const REPLICATION_FACTOR: usize = 20usize;
 
-/// Kademlia maximum message size.
-const DEFAULT_MAX_MESSAGE_SIZE: usize = 16 * 1024;
+/// Kademlia maximum message size. Should fit 64 KiB value + 4 KiB key.
+const DEFAULT_MAX_MESSAGE_SIZE: usize = 70 * 1024;
 
 /// Kademlia configuration.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/protocol/libp2p/kademlia/config.rs
```diff
@@ -52,6 +52,9 @@ const PROTOCOL_NAME: &str = "/ipfs/kad/1.0.0";
 /// Kademlia replication factor.
 const REPLICATION_FACTOR: usize = 20usize;
 
+/// Kademlia maximum message size.
+const DEFAULT_MAX_MESSAGE_SIZE: usize = 1024 * 1024;
```
skunert, line 56: How big are these messages usually? Maybe makes sense to be more conservative? Or is 1MB realistic?

lexnv: The maximum record from the DHT of Kusama is 1827 bytes, I would say capping this at 8 KiB should be enough for most use-cases. I don't expect the bootnodes on DHT to consume much more resources than this either

bkchr (approved):

skunert (approved):

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -52,6 +52,9 @@ const PROTOCOL_NAME: &str = "/ipfs/kad/1.0.0";
 /// Kademlia replication factor.
 const REPLICATION_FACTOR: usize = 20usize;
 
+/// Kademlia maximum message size.
+const DEFAULT_MAX_MESSAGE_SIZE: usize = 16 * 1024;
```
dmitry-markin, line 56:
```suggestion
/// Kademlia maximum message size. Should fit 64 KiB value + 4 KiB key.
const DEFAULT_MAX_MESSAGE_SIZE: usize = 70 * 1024;
```

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -262,6 +271,14 @@ impl ConfigBuilder {
         self
     }
 
+    /// Set the maximum Kademlia message size.
+    ///
+    /// If unspecified, the default maximum message size is 16 KiB.
```
dmitry-markin, line 276:
```suggestion
    ///  Should fit `MemoryStore` max record size. If unspecified, the default maximum message size is 70 KiB.
```

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -1270,7 +1270,7 @@ mod tests {
         let config = Config {
             protocol_names: vec![ProtocolName::from("/kad/1")],
             known_peers: HashMap::new(),
-            codec: ProtocolCodec::UnsignedVarint(None),
+            codec: ProtocolCodec::UnsignedVarint(Some(16 * 1024)),
```
dmitry-markin, line 1273:
```suggestion
            codec: ProtocolCodec::UnsignedVarint(Some(70 * 1024)),
```

dmitry-markin (approved):

Issue #340, opened 2025-02-27
notification/tests: Check test state and increase code coverage


lexnv:
Increase the code coverage for the notification substream validation testing, and ensure the notification state is checked.

Part of `accept_fails_due_to_closed_substream` test:

https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/protocol/notification/tests/substream_validation.rs#L332

PR #360, opened 2025-03-18
notif/tests: check test state

Imod7:
Relates to https://github.com/paritytech/litep2p/issues/340

### Description
Added the check that the peer's `state` is `Closed`. This is the expected state since:
- in the beginning we check that the peer starts with state [Closed](https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/protocol/notification/tests/substream_validation.rs#L275) 
- then we check that it changes to `state` [Validating](https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/protocol/notification/tests/substream_validation.rs#L295)
- then we have a [NegotiationError](https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/protocol/notification/tests/substream_validation.rs#L325)
- so then we should check again that the peer's state is reset to `Closed`.

### Additional Changes
Replaced also all instances of 
```
_ => panic!("invalid state for peer"),
``` 
with 
```
state => panic!("invalid state for peer: {state:?}"),
``` 
since it might be more helpful for debugging purposes. If not necessary, I can revert these.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,8 @@ network-interface = "1.1.1"
 parking_lot = "0.12.3"
 pin-project = "1.1.0"
 prost = "0.12.6"
-quinn = { version = "0.9.3", default-features = false, features = ["tls-rustls", "runtime-tokio"], optional = true }
 rand = { version = "0.8.0", features = ["getrandom"] }
 rcgen = "0.10.0"
-ring = "0.16.20"
 serde = "1.0.158"
 sha2 = "0.10.8"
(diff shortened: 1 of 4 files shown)


dmitry-markin (approved):
Thanks!

lexnv (approved):
Looks good to me! Thanks! 🙏

PR #361, opened 2025-03-31
bitswap: Add write timeout for sending frames

lexnv:
This PR adds a write timeout of 15 seconds for sending frames 

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/bitswap/mod.rs
+++ b/src/protocol/libp2p/bitswap/mod.rs
@@ -35,7 +35,7 @@ use futures::{future::BoxFuture, stream::FuturesUnordered, StreamExt};
 use prost::Message;
 use tokio::sync::mpsc::{Receiver, Sender};
 
-use std::collections::HashMap;
+use std::{collections::HashMap, time::Duration};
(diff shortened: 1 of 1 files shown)


skunert (approved):

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -54,6 +54,9 @@ mod schema {
 /// Log target for the file.
 const LOG_TARGET: &str = "litep2p::ipfs::bitswap";
 
+/// Write timeout for outbound messages.
+const WRITE_TIMEOUT: Duration = Duration::from_secs(15);
```
dmitry-markin, line 58: Is this the same constant libp2p uses?

lexnv: Should be in the same ballpark as substrate 15s timeout: https://github.com/paritytech/polkadot-sdk/blob/0404a8624964441011730e274c7a02972b63245c/substrate/client/network/src/bitswap/mod.rs#L111

PR #365, opened 2025-04-02
SRLabs: Introduce Fuzzing Harness

R9295:
We use [ziggy](https://github.com/srlabs/ziggy/) to orchestrate fuzzing. Ziggy uses both AFL++ and Honggfuzz under the hood.

This harness performs "structure aware" fuzzing where the messages are serialized using bincode/serde, similar to [arbitrary](https://github.com/rust-fuzz/arbitrary)

Note that I had to update ``cid`` to get the ``serde`` feature support. Hope this is okay.


src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -23,14 +23,15 @@
 use crate::{
     error::Error,
     protocol::{
-        libp2p::bitswap::handle::BitswapCommand, Direction, TransportEvent, TransportService,
```
lexnv, line 26: Nice!

src/protocol/notification/handle.rs
```diff
@@ -437,6 +437,17 @@ impl NotificationHandle {
     pub fn notification_sink(&self, peer: PeerId) -> Option<NotificationSink> {
         self.peers.get(&peer).cloned()
     }
+
+    #[cfg(feature = "fuzz")]
+    /// Expose functionality for fuzzing
+    pub async fn fuzz_send_message(&mut self, command: NotificationCommand) -> crate::Result<()> {
```
lexnv, line 443: dq: Instead of adding this method for fuzzing, could we match on the `command` and call the appropriate method on the handler?

src/types/protocol.rs
```diff
@@ -28,20 +28,30 @@ use std::{
 
 /// Protocol name.
 #[derive(Debug, Clone)]
+#[cfg_attr(feature = "fuzz", derive(serde::Serialize, serde::Deserialize))]
 pub enum ProtocolName {
+    #[cfg(not(feature = "fuzz"))]
```
lexnv, line 33: This was added because the proto name generated by the fuzzer could not be static?

src/protocol/libp2p/bitswap/handle.rs
```diff
@@ -107,6 +109,13 @@ impl BitswapHandle {
     pub async fn send_response(&self, peer: PeerId, responses: Vec<ResponseType>) {
         let _ = self.cmd_tx.send(BitswapCommand::SendResponse { peer, responses }).await;
     }
+
+    #[cfg(feature = "fuzz")]
+    /// Expose functionality for fuzzing
+    pub async fn fuzz_send_message(&mut self, command: BitswapCommand) -> crate::Result<()> {
```
lexnv, line 115: nit: Since the `BitswapCommand` has all the fields exposed, could we instead use `send_response`? 🤔

lexnv (approved):
This looks good to me! Thanks a lot for contributing 🙏 

The fuzzer logic looks solid, I had some minor thoughts on the `fuzz_command` method. @dmitry-markin would love to get your thoughts on this whenever you have some spare time

src/types/protocol.rs
R9295: Serde cannot de-serialize ``&'static str`` and since we use serde to de-serialize fuzzer input, the ``Static`` variant cannot be supported.

R9295:
@lexnv I think the consistency of ``fuzz_send_message`` across all protocols is helpful, especially as the ``Handler`` API may evolve. Generally, matching and calling the handler is only partially doable since  a couple of protocols do not expose  all functionality. 

However, if this is a must, I can implement the changes.

Issue #209, opened 2024-08-16
kad: Consider only adding reached addresses to the routing table


dmitry-markin:
Currently, `litep2p` adds all addresses of peers discovered to the local routing table when handling [`KademliaMessage::FindNode`](https://github.com/paritytech/litep2p/blob/a9f69697315a64cf886af4baff6ec8f25e1e0546/src/protocol/libp2p/kademlia/mod.rs#L407) and [`KademliaMessage::GetRecord`](https://github.com/paritytech/litep2p/blob/a9f69697315a64cf886af4baff6ec8f25e1e0546/src/protocol/libp2p/kademlia/mod.rs#L458)

`libp2p`, on the contrary, only adds the address if we dialed the node and it was reached over it. This is done when handling [`HandlerEvent::ProtocolConfirmed`](https://github.com/libp2p/rust-libp2p/blob/d9ee266bc03e7159e415e406e35ffafbb1660d42/protocols/kad/src/behaviour.rs#L2260) & [`HandlerEvent::ProtocolNotSupported`](https://github.com/libp2p/rust-libp2p/blob/d9ee266bc03e7159e415e406e35ffafbb1660d42/protocols/kad/src/behaviour.rs#L2268).

Consider pros and cons of adding an address only if we dialed and reached the node over it. One thing to take into account is the ability to poison the routing tables of other nodes by advertising rogue addresses of peers over DHT.

Issue #232, opened 2024-09-03
transport/manager: Refill dial addresses after a dial failure to make dialing more robust


lexnv:
Litep2p supports dialing a peer with a maximum of 8 parallel dial addresses.

https://github.com/paritytech/litep2p/blob/e3a22d5cfbf255e74be1acceb3ea6ec0baa43c09/src/transport/manager/mod.rs#L472-L476

At the moment, 8 or less addresses are extracted from the available addresses of a peer.
Then, each protocol is provided a list of addresses and the peer enters an `Opening` state (multiple dial requests):

https://github.com/paritytech/litep2p/blob/e3a22d5cfbf255e74be1acceb3ea6ec0baa43c09/src/transport/manager/mod.rs#L548-L570

This is problematic when the dial fails on initially the provided addresses, although the peer contains more addresses that are rechable.

To make the dialing more robust, refill dial addresses if any, after dial failures:
- option A: Refactor transport trait to modify the `open()` method to return a future
  - with this option, we have complete control over what individual protocols are doing when it comes to dialing
  - an advantage of this is that we can "backpressure" the amount of global dials that happen in litep2p
  - another advantage is that we can refill failed dials immediately with another future 
  - anecdata: my system's ability to dial slows down after around 250-300 parallel dials
  
- option B: Refactor the transport manager to refill the dial addresses
  - this is similar to option A, without refactoring the transport trait
  - this may be easier to implement, although dials can only be refilled after a transport (TCP) fails to dial all provided addresses

I tend to lean towards option A for the long term benefits

dmitry-markin:
Not sure if this should be of high priority. If it happens 8 first addresses of a peer are unreachable, there is something odd with discovery. Under normal circumstances peers should not provide that many external addresses, and the list should be ordered in a way most reliable addresses are returned first (i.e., `--public-addr`).

lexnv:
Yep indeed, I think https://github.com/paritytech/litep2p/issues/209 should be higher on our list. I think this was one of the most important issues that came about after the kad performance investigation that may improve things :D

PR #369, opened 2025-04-08
kad: Improve robustness of addresses from the routing table

lexnv:
This PR improves the robustness of the kademlia addresses stored in the routing table.

The implementation uses an eventually-consistent address store, based on the information provided by the transport manager. This has the benefit of not sharing the transport manager address store with the routing table, therefore not paying for the performance hit of locking the addresses (globally) on kademlia queries.

- transport manager exposes all dialed addresses properly to the installed protocols
- the kademlia peer contains an address store instead of a plain vector (bounded to 32 entries -- please note that the transport manager holds up to 64 entries)
  - similar to the transport manager, the addresses are sorted by score depending on success / failure dials
  - when an address score is negative it is subject to removal
- on successful dial, the score is increased for the address of said peer
- on dial failures, the score of the failed addresses is decreased
- known good addresses are always provided first
- all kademlia queries rely on this information and the addresses are provided sorted by their score

This resolved the following issue, with the mention that undialed or dialed but failed addresses are still provided if the address store has enough capacity. This ensures that addresses that fail due to temporary network errors, or errors specific to our node (ie, behind firewall -- other peers may reach the address) are provided:
- https://github.com/paritytech/litep2p/issues/209

cc @paritytech/networking

### DIFF
--- a/src/protocol/libp2p/kademlia/bucket.rs
+++ b/src/protocol/libp2p/kademlia/bucket.rs
@@ -27,7 +27,7 @@ use crate::{
 };
 
 /// K-bucket entry.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug)]
 pub enum KBucketEntry<'a> {
     /// Entry points to local node.
     LocalNode,
@@ -48,7 +48,7 @@ impl<'a> KBucketEntry<'a> {
         if let KBucketEntry::Vacant(old) = self {
             old.peer = new.peer;
             old.key = Key::from(new.peer);
-            old.addresses = new.addresses;
+            old.address_store = new.address_store;
             old.connection = new.connection;
         }
     }
@@ -91,6 +91,7 @@ impl KBucket {
         for i in 0..self.nodes.len() {
             match self.nodes[i].connection {
                 ConnectionType::NotConnected | ConnectionType::CannotConnect => {
+                    self.nodes[i].address_store.clear();
                     return KBucketEntry::Vacant(&mut self.nodes[i]);
                 }
                 _ => continue,
@@ -105,7 +106,7 @@ impl KBucket {
     pub fn closest_iter<K: Clone>(&self, target: &Key<K>) -> impl Iterator<Item = &KademliaPeer> {
         let mut nodes: Vec<_> = self.nodes.iter().collect();
         nodes.sort_by(|a, b| target.distance(&a.key).cmp(&target.distance(&b.key)));
-        nodes.into_iter().filter(|peer| !peer.addresses.is_empty())
+        nodes.into_iter().filter(|peer| !peer.address_store.is_empty())
     }
 }
 
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -36,6 +36,7 @@ use crate::{
         Direction, TransportEvent, TransportService,
     },
     substream::Substream,
+    transport::Endpoint,
     types::SubstreamId,
     PeerId,
 };
@@ -220,14 +221,18 @@ impl Kademlia {
     }
 
     /// Connection established to remote peer.
-    fn on_connection_established(&mut self, peer: PeerId) -> crate::Result<()> {
+    fn on_connection_established(&mut self, peer: PeerId, endpoint: Endpoint) -> crate::Result<()> {
         tracing::trace!(target: LOG_TARGET, ?peer, "connection established");
 
         match self.peers.entry(peer) {
             Entry::Vacant(entry) => {
-                if let KBucketEntry::Occupied(entry) = self.routing_table.entry(Key::from(peer)) {
-                    entry.connection = ConnectionType::Connected;
-                }
+                // Set the conenction type to connected and potentially save the address in the
+                // table.
+                //
+                // Note: this happens regardless of the state of the kademlia managed peers, because
+                // an alread occuped entry in the `self.peers` map does not mean that we are
+                // no longer interested in the address / connection type of the peer.
+                self.routing_table.on_connection_established(Key::from(peer), endpoint);
 
                 let Some(actions) = self.pending_dials.remove(&peer) else {
                     entry.insert(PeerContext::new());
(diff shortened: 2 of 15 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -395,12 +414,13 @@ impl Kademlia {
...
+            self.service.add_known_address(&info.peer, addresses.clone().into_iter());
 
             if std::matches!(self.update_mode, RoutingTableUpdateMode::Automatic) {
                 self.routing_table.add_known_peer(
                     info.peer,
-                    info.addresses.clone(),
+                    addresses,
                     self.peers
                         .get(&info.peer)
                         .map_or(ConnectionType::NotConnected, |_| ConnectionType::Connected),
```
dmitry-markin, line 426: This should go into `on_connection_established` and be completely removed from here. Instead of decreasing the scores of addresses we fail to dial with an objective to remove them later better to only add addresses we were able to reach from the very beginning.

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -220,14 +221,18 @@ impl Kademlia {
...
-                if let KBucketEntry::Occupied(entry) = self.routing_table.entry(Key::from(peer)) {
-                    entry.connection = ConnectionType::Connected;
-                }
+                // Set the conenction type to connected and potentially save the address in the
+                // table.
+                //
+                // Note: this happens regardless of the state of the kademlia managed peers, because
+                // an alread occuped entry in the `self.peers` map does not mean that we are
+                // no longer interested in the address / connection type of the peer.
+                self.routing_table.on_connection_established(Key::from(peer), endpoint);
```
dmitry-markin, line 235: This should be the place where we insert the address into the routing table for the first time and not just update it's score.

lexnv: There are two edge-cases to adding only dialed successful addresses:

(I) The transport manager will dial 8 addresses in parallel. The first dial success will be reported back to the installed protocols (ie the address). The rest 7 pending dials will abruptly terminate their execution (futures canceled to save resources):
   
   https://github.com/paritytech/litep2p/blob/54a8e0608213943fd70cc80939f73290cbe40b12/src/transport/manager/mod.rs#L814-L820
   
For TCP this results in:
https://github.com/paritytech/litep2p/blob/54a8e0608213943fd70cc80939f73290cbe40b12/src/transport/tcp/mod.rs#L532-L536

This is different than libp2p, IIRC libp2p will let the dials happen and report all addresses. This might be a major way we save resources if addresses are taking too long to dial.
There's yet another difference, we won't go through all dialed addresses, just the first 8 then report back a failure:
- https://github.com/paritytech/litep2p/issues/232


(II) A peer might have no addresses reachable by our network, however the rest of the network might be able to reach the said peer. In this case, it might be beneficial to save addresses with a lower priority.

We could probably try to change the behavior in (I) and consume more CPU resources, but I still think we'd need to refactor the transport manager to achieve this (maybe introduce a new internal event that does not advances state machines and does not cause panics while transitioning our states). That on top of the "refilling of dialing addresses"

Wdyt? 🙏

src/protocol/libp2p/kademlia/routing_table.rs
```diff
@@ -119,6 +120,46 @@ impl RoutingTable {
         self.buckets[index.get()].entry(key)
     }
 
+    /// Update the addresses of the peer on dial failures.
+    ///
+    /// The addresses are updated with a negative score making them subject to removal.
```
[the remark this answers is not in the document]
lexnv: Negative addresses will be subject to removal when the capacity (32) is reached. At that point, a new "undialed" address with score zero will enter the store and kick a negative one

src/protocol/libp2p/kademlia/bucket.rs
```diff
@@ -91,6 +91,7 @@ impl KBucket {
         for i in 0..self.nodes.len() {
             match self.nodes[i].connection {
                 ConnectionType::NotConnected | ConnectionType::CannotConnect => {
+                    self.nodes[i].address_store.clear();
```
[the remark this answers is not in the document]
lexnv: If a connected peer overwrites a disconnected entry (with addresses), that would be fine since the disconnected peer is no longer part of the kademlia (ie subject to eviction and removal at this point).

This will not affect dialing, as the transport manager will keep its view of the AddressStore around. So, even if we overwrite this entry, we'll be able to dial and establish a connection with a disconnected peer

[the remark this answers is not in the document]
lexnv: This makes sense, I've removed the address store clear() call. The address store will eventually get replaced with a new one when a peer is inserted into a vacant entry, therefore there's no risk of polluting peer A addresses with another peer B ones

src/protocol/libp2p/kademlia/routing_table.rs
```diff
@@ -119,6 +120,46 @@ impl RoutingTable {
...
+            ?key,
+            ?addresses,
+            "on dial failure"
+        );
+
+        if let KBucketEntry::Occupied(entry) = self.entry(key) {
+            for address in addresses {
+                entry.address_store.insert(AddressRecord::from_raw_multiaddr_with_score(
+                    address.clone(),
+                    -100i32,
```
dmitry-markin, line 138: May be extract these two constants to the top of the file?

src/protocol/libp2p/kademlia/routing_table.rs
```diff
@@ -119,6 +120,46 @@ impl RoutingTable {
...
+    /// is updated in the store of the peer, making it more likely to be used in the future.
+    pub fn on_connection_established(&mut self, key: Key<PeerId>, endpoint: Endpoint) {
+        tracing::trace!(target: LOG_TARGET, ?key, ?endpoint, "on connection established");
+
+        if let KBucketEntry::Occupied(entry) = self.entry(key) {
+            entry.connection = ConnectionType::Connected;
+
+            if let Endpoint::Dialer { address, .. } = endpoint {
+                entry.address_store.insert(AddressRecord::from_raw_multiaddr_with_score(
+                    address, 100i32,
```
dmitry-markin, line 157: A second constant.

dmitry-markin (approved):

PR #373, opened 2025-04-14
crypto/noise: Set timeout limits for the noise handshake

lexnv:
This PR sets timeout limits for the crypto/noise handshake negotiation.

By default, the timeout is bounded similar to protocol negotiation to the substream open timeout.

To allow this change, the WebSocket internal API has been opened to accept the open timeout. Before this change, we had no granular control and used the connection timeout as a fallback.

cc @paritytech/networking

### DIFF
--- a/src/crypto/noise/mod.rs
+++ b/src/crypto/noise/mod.rs
@@ -709,62 +709,71 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
     role: Role,
     max_read_ahead_factor: usize,
     max_write_buffer_size: usize,
+    timeout: std::time::Duration,
 ) -> Result<(NoiseSocket<S>, PeerId), NegotiationError> {
-    tracing::debug!(target: LOG_TARGET, ?role, "start noise handshake");
+    let handle_handshake = async move {
+        tracing::debug!(target: LOG_TARGET, ?role, "start noise handshake");
 
-    let mut noise = NoiseContext::new(keypair, role)?;
-    let payload = match role {
-        Role::Dialer => {
-            // write initial message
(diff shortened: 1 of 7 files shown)


lexnv:
Test timedout in https://github.com/paritytech/litep2p/actions/runs/14444528198/job/40501930199:

```
 test protocol::notification::send_using_notification_sink_websocket has been running for over 60 seconds
```

Retriggering CI before pushing a new backport of tests for webscoket

lexnv:
Additional tests added in a follow-up PR:
- https://github.com/paritytech/litep2p/pull/374

src/crypto/noise/mod.rs
```diff
@@ -709,62 +709,71 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
...
+                max_write_buffer_size,
+            ),
+            peer,
+        ))
+    };
+
+    match tokio::time::timeout(timeout, handle_handshake).await {
+        Err(_) => Err(NegotiationError::Timeout),
+        Ok(result) => result,
+    }
```
dmitry-markin, line 776: Are you sure we won't end up in the inconsistent state if the future is canceled due to timeout? I.e., we should at least not reuse the `io` object if the future was interrupted?

dmitry-markin (approved):

src/crypto/noise/mod.rs
lexnv: I think this is fine here since we consume the given io `mut io: S` and return a `NoiseSocket<S>` only on success. We have a similar pattern for `/noise` and `/yamux` negotiations 🙏

PR #374, opened 2025-04-15
websocket/stream: Fix state reset and backport connection tests

lexnv:
This PR addresses a bug in the WebSocket stream state machine. The internal state was incorrectly reset to `State::ReadyToSend` upon receiving `Poll::Pending`. The correct behavior is to retain the `State::FlushPending` state until the flush operation completes.

Additionally, the following tests were backported from the TCP connection component:
- yamux_timeout_listener
- yamux_timeout_dialer
- yamux_not_supported_listener
- yamux_not_supported_dialer
- noise_timeout_dialer
- noise_wrong_handshake_listener
- noise_timeout_listener
- noise_not_supported_listener
- noise_not_supported_dialer
- multistream_select_not_supported_listener
- multistream_select_not_supported_dialer

While at it, we had a test guarded by a wrongly named feature flag `webscocket`, which is now also enabled.

cc @paritytech/networking

### DIFF
--- a/src/transport/websocket/stream.rs
+++ b/src/transport/websocket/stream.rs
@@ -34,6 +34,7 @@ use std::{
 const DEFAULT_BUF_SIZE: usize = 8 * 1024;
 
 /// Send state.
+#[derive(Debug)]
 enum State {
     /// State is poisoned.
     Poisoned,
@@ -46,6 +47,7 @@ enum State {
 }
 
 /// Buffered stream which implements `AsyncRead + AsyncWrite`
+#[derive(Debug)]
 pub(super) struct BufferedStream<S: AsyncRead + AsyncWrite + Unpin> {
     /// Write buffer.
     write_buffer: BytesMut,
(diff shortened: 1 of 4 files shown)


src/transport/websocket/stream.rs
```diff
@@ -124,7 +126,7 @@ impl<S: AsyncRead + AsyncWrite + Unpin> futures::AsyncWrite for BufferedStream<S
                         Poll::Ready(Err(_error)) =>
                             return Poll::Ready(Err(std::io::ErrorKind::UnexpectedEof.into())),
                         Poll::Pending => {
-                            self.state = State::ReadyToSend;
+                            self.state = State::FlushPending;
```
dmitry-markin, line 129: Out of curiosity, how did you spot this?

dmitry-markin (approved):

src/transport/websocket/stream.rs
lexnv: Went on a rabbit hole during this flaky test: https://github.com/paritytech/litep2p/pull/373#issuecomment-2804403610 🙈

PR #376, opened 2025-04-17
notifications: Exit protocols on handle drop to save up CPU of `minimal-relay-chains`

lexnv:
This PR ensures that notification tasks are fully terminated when the associated handle is dropped, addressing an issue specific to `minimal-relay-chains`. We have operated from the assumption that the handle is never supposed to get dropped, since protocols cannot be dynamically injected after the backend is started.

This PR effectively reduces the CPU usage of minimal-relay-chains from 7% to 0.1%.

We have noticed that the CPU consumption of minimal-relay-chains has increased compared to libp2p backends. However, our metrics for CPU consumption of full relay chains have drastically improved.

The issue is the following:
- substrate needs a block-announces notification config to initiate the network components.
- the minimal-relay-chains do not need this notification around, and drop the associated handle.
- this behaviour causes a CPU loop that continuously spams the following log line:

```
litep2p::notification: [Parachain] user protocol has exited, exiting protocol=/b0a8d493285c2df73290dfb7e61f870f17b41801197a149ca93654499ea3dafe/block-announces/1
```

CPU before and after:
Screenshot 2025-04-17 at 17 44 23

Fixes: https://github.com/paritytech/polkadot-sdk/issues/7998

### DIFF
--- a/src/protocol/notification/mod.rs
+++ b/src/protocol/notification/mod.rs
@@ -1606,7 +1606,7 @@ impl NotificationProtocol {
     }
 
     /// Handle next notification event.
-    async fn next_event(&mut self) {
+    async fn next_event(&mut self) -> bool {
         // biased select is used because the substream events must be prioritized above other events
         // that is because a closed substream is detected by either `substreams` or `negotiation`
         // and if that event is not handled with priority but, e.g., inbound substream is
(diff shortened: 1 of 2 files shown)


src/protocol/notification/mod.rs
```diff
@@ -1606,7 +1606,7 @@ impl NotificationProtocol {
     }
 
     /// Handle next notification event.
-    async fn next_event(&mut self) {
+    async fn next_event(&mut self) -> bool {
```
bkchr, line 1609:
```suggestion
    ///
    /// Returns `true` when the user command stream was dropped.
    async fn next_event(&mut self) -> bool {
```

bkchr (approved):

bkchr:
Generally could use some test. Make a test that directly drops this stream and then `poll` the `run` function.

PR #392, opened 2025-05-14
crypto/noise: Show peerIDs that fail to decode

lexnv:
This PR enriches the errors coming from the crypto/noise decryption by providing peer IDs.

This is useful for debugging:
- https://github.com/paritytech/polkadot-sdk/issues/8525

### DIFF
--- a/src/codec/identity.rs
+++ b/src/codec/identity.rs
@@ -102,7 +102,7 @@ mod tests {
         let bytes = [3u8; 64];
         let mut bytes = BytesMut::from(&bytes[..]);
 
-        let decoded = codec.decode(&mut bytes);
+        assert!(codec.decode(&mut bytes).unwrap().is_none());
     }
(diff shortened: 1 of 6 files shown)


skunert (approved):

lexnv:
> IMO this should be a temporary fix in a custom branch if this is needed for debugging. We shouldn't print raw data from the remote if it's malformed/malicious

This should only print internal counters and lengths (like expected frame size and buffer length received over the wire, which already happened for trace logs). Ideally, the errors coming from this component should go away once Kusama updates to the upcoming release

I'll expect the following error to pop up in the meanwhile. I'd still like to see if we get different types of errors (edge-cases we might have missed), other than:

```
failed to decrypt message ty=WebSocket peer=PeerId("..") buf_len=12 frame_size=28 error=Decrypt
```

PR #394, opened 2025-05-16
replace duplicate code with common function

seun-ja:


src/protocol/mod.rs
```diff
@@ -141,3 +142,18 @@ pub trait UserProtocol: Send {
     /// Start the the user protocol event loop.
     async fn run(self: Box<Self>, service: TransportService) -> crate::Result<()>;
 }
+
+pub fn sort_address(addresses: impl Iterator<Item = Multiaddr>, peer_id: PeerId) -> Vec<Multiaddr> {
```
lexnv: nit: I would name this method `ensure_address_with_peer` or similar, since it is not really sorting the input 👍

src/protocol/libp2p/kademlia/routing_table.rs
```diff
@@ -188,17 +190,7 @@ impl RoutingTable {
         );
 
         // TODO: https://github.com/paritytech/litep2p/issues/337 this has to be moved elsewhere at some point
```
lexnv: We could probably remove the TODO here? Or do you think there are more places where this may happen? 🤔

seun-ja: Oh yeah, I forgot to remove that. It's all covered now, no more duplicates

src/protocol/mod.rs
seun-ja: Yeah, that's a better name. Would use that instead

src/protocol/transport_service.rs
```diff
@@ -467,15 +468,7 @@ impl TransportService {
...
-        let addresses: HashSet<Multiaddr> = addresses
-            .filter_map(|address| {
-                if !std::matches!(address.iter().last(), Some(Protocol::P2p(_))) {
-                    Some(address.with(Protocol::P2p(Multihash::from_bytes(&peer.to_bytes()).ok()?)))
-                } else {
-                    Some(address)
-                }
-            })
-            .collect();
+        let addresses = ensure_address_with_peer(addresses.into_iter(), *peer);
```
lexnv: There's one tiny issue with this approach: we were previously collecting into a `HashSet`, which deduplicates items. Whereas using the Vec by default, we could have duplicates here

PR #395, opened 2025-05-22
substream/fix: Allow empty payloads with 0-length frame

lexnv:
This PR ensures that zero-length frames are propagated to higher levels.

An issue was noted with the compatibility between litep2p and smoldot, which resulted in smoldot not being accepted by substrate-based chains.

The issue stems from the litep2p substream (high level) implementation, which did not handle zero-length frames.

Smoldot sends during the notification handshake a zero-length frame (ie, `frame [0]`).
However, full nodes (litep2p and libp2p) send the node role `Role::Full` (ie, `frame [1, 1]`).
Substrate nodes expected the node `Role`, however zero-length frames were never punished.

This behavior causes the zero-length frames to go unnoticed and further call into `poll_read`.
In the case of smoldot, the connection is terminated after the handshake timeout of 10 seconds.

### Testing Done
- Reproduced the issue locally with zombinet and papi (https://github.com/polkadot-api/polkadot-api/tree/main/integration-tests/zombie-tests)
- The e2e tests were timing out, because litep2p rejected the `/transactions/1` substream with smoldot (however it accepted the /sync substream which sends a different payload)
- Added a new notification test to illustrate this behavior with different payloads (including empty ones)

Closes: https://github.com/smol-dot/smoldot/issues/2128

Thanks @josepot and @tomaka for repro-case and details 🙏

### DIFF
--- a/src/substream/mod.rs
+++ b/src/substream/mod.rs
@@ -645,6 +645,13 @@ impl Stream for Substream {
                                                 }
 
                                                 this.offset = 0;
+                                                // Handle empty payloads detected as 0-length frame.
+                                                // The offset must be cleared to 0 to not interfere
+                                                // with next framing.
+                                                if size == 0 {
+                                                    return Poll::Ready(Some(Ok(BytesMut::new())));
+                                                }
+
                                                 this.current_frame_size = Some(size);
                                                 this.read_buffer = BytesMut::zeroed(size);
                                             }
--- a/tests/protocol/notification.rs
+++ b/tests/protocol/notification.rs
@@ -2010,6 +2010,240 @@ async fn dialer_fallback_protocol_works(transport1: Transport, transport2: Trans
 }
 
 #[tokio::test]
+async fn zero_byte_handshake_tcp() {
+    // Full node role.
+    zero_byte_handshake(
+        Transport::Tcp(TcpConfig {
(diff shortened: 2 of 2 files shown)


tomaka:
> Smoldot sends during the notification handshake a zero-length frame (ie, frame [0]).
However, full nodes (litep2p and libp2p) send the node role Role::Full (ie, frame [1, 1]).

Ah, I see.
If I recall correctly, my intention as the designer of the protocol was that notifications have an empty payload. In Substrate, however, we're sending a Role as well because at the time the code written there was no way to differentiate between transactions and GrandPa in the code that was sending the handshake, so I simply sent a Role all the time. I guess I overlooked this when writing smoldot.

I'm not really opinionated on this, for what it's worth, and I can also modify smoldot to send a Role if desired.

dmitry-markin (approved):

lexnv:
Sorry for the late response here🙏 

Indeed this was a miss on the litep2p side. Assuming it's a minor change, sending the role of the light-client (https://github.com/paritytech/polkadot-sdk/blob/6ac204d52beab2e5c22588519f639cc03781ec26/substrate/client/network/common/src/role.rs#L93) would still benefit Kusama, which uses litep2p entirely. We'll follow up with a release shortly, but I expect a window of 2-3 weeks until enough nodes update to the fix.

PR #400, opened 2025-06-03
core: Register protocols at runtime

lexnv:
This PR adds the ability to register notification protocols and request-response protocols at runtime.

This is beneficial for users who don't have the full set of protocols during the node initialization. Instead, the protocols will be known at runtime, potentially after calling into the substrate runtime itself.

The counterpart on the substrate side must take into account that each protocol instance has a channel of communication with the rest of the protocols. This might complicate things on the substrate side, but in theory, it's possible to pass a new channel.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,18 @@ pub enum Litep2pEvent {
     },
 }
 
+/// Configuration for adding protocols at runtime (protocols that are started after the Litep2p
+/// instance is created).
+struct RuntimeConfigs {
+    /// The keep-alive timeout for protocols.
+    ///
+    /// Needed for supporting protocols at runtime.
+    keep_alive_timeout: std::time::Duration,
+
+    /// Executor used to run protocols.
+    executor: Arc<dyn crate::executor::Executor>,
+}
+
 /// [`Litep2p`] object.
 pub struct Litep2p {
     /// Local peer ID.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -149,6 +161,11 @@ pub struct Litep2p {
 
     /// Bandwidth sink.
     bandwidth_sink: BandwidthSink,
+
+    /// The keep-alive timeout for protocols.
```
dmitry-markin, line 165:
```suggestion
    /// The executor & keep-alive timeout for protocols.
```

src/lib.rs
```diff
@@ -136,6 +136,18 @@ pub enum Litep2pEvent {
...
+/// Configuration for adding protocols at runtime (protocols that are started after the Litep2p
+/// instance is created).
+struct RuntimeConfigs {
+    /// The keep-alive timeout for protocols.
+    ///
+    /// Needed for supporting protocols at runtime.
+    keep_alive_timeout: std::time::Duration,
+
+    /// Executor used to run protocols.
+    executor: Arc<dyn crate::executor::Executor>,
```
dmitry-markin, line 148: nit: IMO `executor` deserves a dedicated field in `Litep2p`, and may be we don't need a new struct then.

dmitry-markin (approved):
Nice one :+1:

PR #412, opened 2025-07-08
manager: Expose option to forbid dialing private IP addresses

lexnv:
This PR exposes an option to forbid or allow dialing private IP addresses.

For test networks started locally, it is recommended to use private IP addresses to establish connections.

However, for running production validators, the validator node should never attempt to dial a private IP address.
This behavior can be detected by some cloud providers as port scanning. The downstream effect of that depends on the cloud provider, but most likely, the node will become unreachable or the instance will be terminated.

Before this PR, litep2p might receive a local (private) IP address on either the Kademlia or the Identify protocol from a libp2p node. Now, the transport manager will not forward the dialing request to the installed protocols if the private IP option is disabled.

This may be the case of https://github.com/paritytech/polkadot-sdk/issues/8915, where a node operator running a validator using litep2p on the Polkadot network loses connectivity after a few days.

Closes: https://github.com/paritytech/polkadot-sdk/issues/8631

### Next Steps
- [ ] Wait for further information to be provided by the node operator
- [ ] Ensure the PR is properly tested locally, in versi-net and on kusama validators

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -126,7 +126,7 @@ pub struct ConfigBuilder {
     keep_alive_timeout: Duration,
 
     /// True if litep2p should attempt to dial local addresses.
-    local_dialing: bool,
+    use_private_ip: bool,
 }
 
 impl Default for ConfigBuilder {
@@ -160,7 +160,7 @@ impl ConfigBuilder {
             known_addresses: Vec::new(),
             connection_limits: ConnectionLimitsConfig::default(),
             keep_alive_timeout: KEEP_ALIVE_TIMEOUT,
-            local_dialing: true,
+            use_private_ip: true,
         }
     }
 
@@ -282,19 +282,20 @@ impl ConfigBuilder {
         self
     }
 
-    /// Set the local dialing behavior.
+    /// Allow or forbid connecting to private IPv4/IPv6 addresses.
     ///
-    /// When the local dialing is enabled, litep2p will attempt to dial local addresses.
+    /// When the private IP is enabled, litep2p will attempt to dial local addresses.
     /// This is useful for testing or when you want to preserve local connections.
     ///
-    /// However, for production use, it is recommended to disable local dialing
+    /// However, for production use, it is recommended to disable the private IP dialing
     /// to avoid unnecessary local traffic. Furthermore, it is not recommended
-    /// to enable local dialing when running a validator in a cloud provider, as this behavior
+    /// to enable private IP dialing when running a validator in a cloud provider, as this behavior
     /// might be misinterpreted by the cloud provider's network policies as port scanning.
     ///
-    /// By default, local dialing is enabled.
-    pub fn with_local_dialing(mut self, enable: bool) -> Self {
-        self.local_dialing = enable;
+    /// Address allocation for private networks is specified by
+    /// [RFC1918](https://tools.ietf.org/html/rfc1918)).
+    pub fn with_private_ip(mut self, enable: bool) -> Self {
+        self.use_private_ip = enable;
         self
     }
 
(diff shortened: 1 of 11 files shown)


src/config.rs
```diff
@@ -124,6 +124,9 @@ pub struct ConfigBuilder {
 
     /// Close the connection if no substreams are open within this time frame.
     keep_alive_timeout: Duration,
+
+    /// True if litep2p should attempt to dial local addresses.
+    use_private_ip: bool,
```
dmitry-markin:
```suggestion
    dial_global_ips_only: bool,
```
?

private_ip != !global_ip

src/transport/manager/mod.rs
```diff
@@ -194,6 +194,44 @@ impl Stream for TransportContext {
...
+}
+
+impl IpDialingMode {
+    /// Check if the IP address in the given `Multiaddr` is global.
+    pub fn is_address_global(address: &Multiaddr) -> bool {
+        let ip = match address.iter().next() {
+            Some(Protocol::Ip4(ip)) => ip_network::IpNetwork::from(ip),
+            Some(Protocol::Ip6(ip)) => ip_network::IpNetwork::from(ip),
+            Some(Protocol::Dns(_)) | Some(Protocol::Dns4(_)) | Some(Protocol::Dns6(_)) =>
+                return true,
```
dmitry-markin: This will leak addresses like `/dns4/10.10.10.10/tcp/30333/...`, and also dns records can point to local IP addresses.

dmitry-markin:
> One consequence might be that mDNS won't work with `IpDialingMode::GlobalOnly`, because multicast addresses are not considered global.

This is not the case luckily — mDNS directly uses UDP sockets instead of relying on `TransportManager`: https://github.com/paritytech/litep2p/blob/861847032372b9a80de86beadf581371209d715a/src/protocol/mdns.rs#L129

src/transport/manager/address.rs
```diff
@@ -259,6 +259,13 @@ impl AddressStore {
         records.sort_by(|lhs, rhs| rhs.score.cmp(&lhs.score));
         records.into_iter().take(limit).map(|record| record.address).collect()
     }
+
+    /// Similar to [`Self::addresses`], but returns an iterator over the addresses.
+    pub fn addresses_iter(&self) -> impl Iterator<Item = &Multiaddr> {
+        let mut records = self.addresses.values().collect::<Vec<_>>();
+        records.sort_by(|lhs, rhs| rhs.score.cmp(&lhs.score));
+        records.into_iter().map(|record| &record.address)
+    }
```
dmitry-markin: Returning an iterator after allocating a vector "hides" the cost of calling this function. IMO having `addresses()` should be enough.

src/transport/manager/handle.rs
```diff
@@ -260,7 +281,18 @@ impl TransportManagerHandle {
...
+                return Err(ImmediateDialError::NoAddressAvailable);
+            }
+
+            // Check if we can dial at least one address from the store in the current operating
+            // mode. If the mode is `IpDialingMode::All`, this will always return
+            // `true` on the first address.
+            let has_dialable_address = address_store
+                .addresses
+                .iter()
+                .any(|(address, _record)| self.ip_dialing_mode.allows_address(address));
```
dmitry-markin: Currently we check if the address is global two times — in `TransportManagerHandle`, and later in `TransportManager`. May be it's enough to keep the check only in one place?

dmitry-markin (approved):

src/transport/manager/handle.rs
lexnv: Indeed, we could keep the check in `TransportManager`. I think I wanted to propagate the dial error as soon as possible (not entirely sure since itsbeen a while)

lexnv:
From the testing done locally with a patched polkadot binary, this PR ensures that no dials will be made on private IP addresses.

This was tested similarly to the linked issue above `netstat -nputw | grep -E "10\.|\(172\.16\|192\.168\)"`

Issue #134, opened 2024-05-29
ping: Verify received "pong" paylaod


lexnv:
We send a `[0u8; 32]` payload to the peer, then we read the next payload sent by the peer.

After that, we could verify that what we sent (ping) was identical to what we received (pong).

This is to stay compliant with the p2p spec

haikoschol:
This was fixed in #416. @lexnv @dmitry-markin

PR #416, opened 2025-07-19
ping: Conform to the spec & exclude from connection keep-alive

### LINKED ISSUE #415: Ping: repeat requests periodically
Currently, ping request is only sent once when a connection is established. As a mechanism of detecting broken connections ping only makes sense if it is repeated periodically. For example, libp2p repeats pings every [15 seconds by default](https://github.com/libp2p/rust-libp2p/blob/e6b0e34733c11c2d6bc165fda2ce22c995e0805d/protocols/ping/src/handler.rs#L66). We need to do the same.

Care should be taken for ping substreams to not keep connections alive if there are no other substreams. See https://github.com/paritytech/litep2p/issues/414 for more info.

dharjeezy:
This PR makes ping protocol conform to the [spec](https://github.com/libp2p/specs/blob/master/ping/ping.md), specifically:
1. Repeat ping requests instead of sending them only on connect.
2. Reuse ping substreams (both inbound and outbound) for subsequent requests, as required by the spec. This is needed for [smoldot](https://github.com/smol-dot/smoldot) compatibility.
3. Exclude ping substreams from the connection keep-alive mechanism, allowing it to close connection on timeout if only ping (and identify) substreams are open.

Tested litep2p-litep2p, litep2p-(pre-PR)litep2p, litep2p-smoldot.

Closes #415

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -32,14 +32,14 @@ use crate::{
             peer_state::{ConnectionRecord, PeerState, StateDialResult},
             types::PeerContext,
         },
-        Endpoint, Transport, TransportEvent, MAX_PARALLEL_DIALS,
+        Endpoint, Transport, TransportEvent,
     },
     types::{protocol::ProtocolName, ConnectionId},
     BandwidthSink, PeerId,
 };
 
 use address::{scores, AddressStore};
-use futures::{Stream, StreamExt};
+use futures::{future::BoxFuture, stream::FuturesUnordered, Stream, StreamExt};
 use indexmap::IndexMap;
 use multiaddr::{Multiaddr, Protocol};
 use multihash::Multihash;
@@ -211,9 +211,6 @@ pub struct TransportManager {
     /// Bandwidth sink.
     bandwidth_sink: BandwidthSink,
 
-    /// Maximum parallel dial attempts per peer.
-    max_parallel_dials: usize,
-
     /// Installed protocols.
     protocols: HashMap<ProtocolName, ProtocolContext>,
 
@@ -258,6 +255,9 @@ pub struct TransportManager {
 
     /// Opening connections errors.
     opening_errors: HashMap<ConnectionId, Vec<(Multiaddr, DialError)>>,
+
+    /// Pending accept futures with associated connection information.
+    pending_accept: FuturesUnordered<BoxFuture<'static, (PeerId, Endpoint, crate::Result<()>)>>,
 }
 
 /// Builder for [`crate::transport::manager::TransportManager`].
@@ -270,9 +270,6 @@ pub struct TransportManagerBuilder {
 
     /// Bandwidth sink.
     bandwidth_sink: Option<BandwidthSink>,
-
-    /// Maximum parallel dial attempts per peer.
-    max_parallel_dials: usize,
 
     /// Connection limits config.
     connection_limits_config: limits::ConnectionLimitsConfig,
@@ -291,7 +288,6 @@ impl TransportManagerBuilder {
             keypair: None,
             supported_transports: HashSet::new(),
             bandwidth_sink: None,
-            max_parallel_dials: MAX_PARALLEL_DIALS,
             connection_limits_config: limits::ConnectionLimitsConfig::default(),
         }
     }
@@ -314,12 +310,6 @@ impl TransportManagerBuilder {
     /// Set the bandwidth sink
     pub fn with_bandwidth_sink(mut self, bandwidth_sink: BandwidthSink) -> Self {
         self.bandwidth_sink = Some(bandwidth_sink);
-        self
-    }
-
-    /// Set the maximum parallel dials per peer
-    pub fn with_max_parallel_dials(mut self, max_parrallel_dials: usize) -> Self {
-        self.max_parallel_dials = max_parrallel_dials;
         self
     }
 
@@ -355,7 +345,6 @@ impl TransportManagerBuilder {
             local_peer_id,
             keypair,
             bandwidth_sink: self.bandwidth_sink.unwrap_or_else(BandwidthSink::new),
-            max_parallel_dials: self.max_parallel_dials,
             protocols: HashMap::new(),
             protocol_names: HashSet::new(),
             listen_addresses,
@@ -371,6 +360,7 @@ impl TransportManagerBuilder {
             pending_connections: HashMap::new(),
             connection_limits: limits::ConnectionLimits::new(self.connection_limits_config),
             opening_errors: HashMap::new(),
+            pending_accept: FuturesUnordered::new(),
         }
     }
 }
@@ -540,9 +530,6 @@ impl TransportManager {
     pub async fn dial(&mut self, peer: PeerId) -> crate::Result<()> {
         // Don't alter the peer state if there's no capacity to dial.
         let available_capacity = self.connection_limits.on_dial_address()?;
-        // The available capacity is the maximum number of connections that can be established,
-        // so we limit the number of parallel dials to the minimum of these values.
-        let limit = available_capacity.min(self.max_parallel_dials);
 
         if peer == self.local_peer_id {
             return Err(Error::TriedToDialSelf);
@@ -560,7 +547,9 @@ impl TransportManager {
 
         // The addresses are sorted by score and contain the remote peer ID.
         // We double checked above that the remote peer is not the local peer.
-        let dial_addresses = context.addresses.addresses(limit);
+        // Limit addresses by the available connection capacity. The transport layer
+        // handles dial concurrency via `max_parallel_dials`.
+        let dial_addresses = context.addresses.addresses(available_capacity);
         if dial_addresses.is_empty() {
             return Err(Error::NoAddressAvailable(peer));
         }
@@ -1099,6 +1088,35 @@ impl TransportManager {
     pub async fn next(&mut self) -> Option<TransportEvent> {
         loop {
             tokio::select! {
+                (peer, endpoint, result) = self.pending_accept.select_next_some(), if !self.pending_accept.is_empty() => {
+                    match result {
+                        Ok(()) => {
+                            tracing::trace!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                ?endpoint,
+                                "connection accepted and protocols notified",
+                            );
+
+                            return Some(TransportEvent::ConnectionEstablished { peer, endpoint });
+                        }
+                        Err(error) => {
+                            // The pending accept future has failed to inform one of the
+                            // installed protocols about the connection. This can happen when the
+                            // node is shutting down or when the user has dropped the long running protocol.
+                            // To err on the safe side, roll back the state modification done in `on_connection_established`.
+                            self.on_connection_closed(peer, endpoint.connection_id());
+
+                            tracing::error!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                ?endpoint,
+                                ?error,
+                                "failed to notify protocols about connection",
+                            );
+                        }
+                    }
+                }
                 event = self.event_rx.recv() => {
                     let Some(event) = event else {
                         tracing::error!(
@@ -1278,16 +1296,36 @@ impl TransportManager {
                                         "accept connection",
                                     );
 
-                                    let _ = self
+                                    match self
                                         .transports
                                         .get_mut(&transport)
                                         .expect("transport to exist")
-                                        .accept(endpoint.connection_id());
+                                        .accept(endpoint.connection_id())
+                                    {
+                                        Ok(future) => {
+                                            // A ConnectionEstablished is propagated to the user once
+                                            // all protocols have been notified.
+                                            self.pending_accept.push(Box::pin(async move {
+                                                let result = future.await;
+                                                (peer, endpoint, result)
+                                            }));
+                                        }
+                                        Err(error) => {
+                                            // Roll back the state modification done in `on_connection_established` by
+                                            // simulating a closed connection. The transport returns an error
+                                            // while accepting the connection, which can happen if the transport is
+                                            // already closed or the connection is dropped before the accept call.
+                                            self.on_connection_closed(peer, endpoint.connection_id());
 
-                                    return Some(TransportEvent::ConnectionEstablished {
-                                        peer,
-                                        endpoint,
-                                    });
+                                            tracing::debug!(
+                                                target: LOG_TARGET,
+                                                ?peer,
+                                                ?endpoint,
+                                                ?error,
+                                                "failed to accept connection",
+                                            );
+                                        }
+                                    }
                                 }
                                 Ok(ConnectionEstablishedResult::Reject) => {
                                     tracing::trace!(
@@ -1482,8 +1520,11 @@ mod tests {
                 Ok(())
             }
 
-            fn accept(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-                Ok(())
+            fn accept(
+                &mut self,
+                _connection_id: ConnectionId,
+            ) -> crate::Result<BoxFuture<'static, crate::Result<()>>> {
+                Ok(Box::pin(async { Ok(()) }))
             }
 
             fn accept_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1386,6 +1386,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "ip_network"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aa2f047c0a98b2f299aa5d6d7088443570faae494e9ae1305e48be000c9e0eb1"
+
+[[package]]
 name = "ipconfig"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1874,6 +1880,7 @@ dependencies = [
  "hex-literal",
  "hickory-resolver",
  "indexmap",
+ "ip_network",
  "libc",
  "libp2p",
  "mockall",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ ed25519-dalek = { version = "2.1.1", features = ["rand_core"] }
 futures = "0.3.27"
 futures-timer = "3.0.3"
 indexmap = { version = "2.9.0", features = ["std"] }
+ip_network = "0.4"
 libc = "0.2.158"
 mockall = "0.13.1"
 multiaddr = "0.17.0"
--- a/src/config.rs
+++ b/src/config.rs
@@ -265,8 +265,10 @@ impl ConfigBuilder {
     }
 
     /// How many addresses should litep2p attempt to dial in parallel.
+    ///
+    /// The provided number is clamped to a minimum of 1.
     pub fn with_max_parallel_dials(mut self, max_parallel_dials: usize) -> Self {
-        self.max_parallel_dials = max_parallel_dials;
+        self.max_parallel_dials = max_parallel_dials.max(1);
         self
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -175,7 +175,6 @@ impl Litep2p {
             .with_keypair(litep2p_config.keypair.clone())
             .with_supported_transports(supported_transports)
             .with_bandwidth_sink(bandwidth_sink.clone())
-            .with_max_parallel_dials(litep2p_config.max_parallel_dials)
             .with_connection_limits_config(litep2p_config.connection_limits)
             .build();
 
@@ -333,7 +332,8 @@ impl Litep2p {
         }
 
         // enable tcp transport if the config exists
-        if let Some(config) = litep2p_config.tcp.take() {
+        if let Some(mut config) = litep2p_config.tcp.take() {
+            config.max_parallel_dials = litep2p_config.max_parallel_dials;
             let handle = transport_manager.transport_handle(Arc::clone(&litep2p_config.executor));
             let (transport, transport_listen_addresses) =
                 <TcpTransport as TransportBuilder>::new(handle, config, resolver.clone())?;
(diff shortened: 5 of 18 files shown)


src/protocol/libp2p/ping/mod.rs
```diff
@@ -192,6 +195,19 @@ impl Ping {
                     Some(_) => {}
                     None => return,
                 },
+                _ = interval.tick() => {
+                    for peer in &self.peers {
+                        tracing::trace!(target: LOG_TARGET, ?peer, "sending ping");
+                        if let Err(error) = self.service.open_substream(*peer) {
```
dmitry-markin: The libp2p ping protocol spec states that the repeated pings should go on the same substream: https://github.com/libp2p/specs/blob/master/ping/ping.md

Because of this, the issue is actually more difficult due to the need to not keep the connection alive by this ping substream.

I don't think we should break the spec even if it works with the current libp2p implementation. @lexnv what do you think?

lexnv: Indeed, I would comply with the libp2p spec as much as possible. The main issue would be that we cannot distinguish ping susbtreams for the keep alive connection tracker 🤔 

Are we entirely convinced we need to implement this for ping?
- we'll need to adjust the substream API to expose that the substream is ping (accounting towards the keepalive)
- how do we treat failures here? If we cannot open a substream should we terminate the connection? If so, we need to propagate again to the higher levels the signal to terminate the connection
  - offhand, from ping we'll need to propagate towards the transport manager, and from the transport manager we'll need to propagate to individual connections

dharjeezy:
> > Hey, this is going in the right direction, but there should be a way to simplify the implementation by not introducing separate async tasks for inbound/outbound substreams and not adding command channels. It should be possible to implement polling of inbound substreams by putting them into `FuturesUnordered` or, may be, `tokio_stream::StreamMap`. The latter should be quite easy as there is maximum one inbound substream allowed from any remote peer as per ping spec: https://github.com/libp2p/specs/blob/master/ping/ping.md
> 
> `tokio_stream::StreamMap` might not work because we need to also write to substream and not only poll it, but still there should be a way to simplify the code.

@dmitry-markin can you check my current implementation

dmitry-markin (commented):
Hey hey, nice use of `SplitStream`/`SplitSink`! I think what is missing is a cleaner separation between inbound and outbound substreams. On inbound substreams we should only respond to incoming pings, and on outbound send pings ourself. It would be better than relying on the presence of the time in the map, which can get screwed when we both send pings to a peer and receive pings from it.

src/protocol/libp2p/ping/mod.rs
```diff
@@ -60,23 +60,32 @@ pub enum PingEvent {
...
 
-    /// Connected peers.
-    peers: HashSet<PeerId>,
+    /// Sinks we write Pings to.
+    outbound_sinks: HashMap<PeerId, SplitSink<Substream, Bytes>>,
 
-    /// Pending outbound substreams.
-    pending_outbound: FuturesUnordered<BoxFuture<'static, crate::Result<(PeerId, Duration)>>>,
+    /// Streams we read Pings from.
+    /// Keyed by a local counter to handle multiple streams per peer if necessary.
```
haikoschol: As Dmitry [mentioned](https://github.com/paritytech/litep2p/pull/416#pullrequestreview-3360280764), there should be only one inbound ping substream per peer. Can we simplify this? I guess just using the `PeerId` like in `outbound_sinks`?

src/protocol/libp2p/ping/mod.rs
```diff
@@ -24,26 +24,26 @@ use crate::{
...
 
 pub use config::{Config, ConfigBuilder};
-
 mod config;
 
 // TODO: https://github.com/paritytech/litep2p/issues/132 let the user handle max failures
 
 /// Log target for the file.
 const LOG_TARGET: &str = "litep2p::ipfs::ping";
+const PING_TIMEOUT: Duration = Duration::from_secs(10);
```
haikoschol:
```suggestion
```

Since it is unused. However, we probably want to have a timeout for receiving pongs. Arguably out of scope for this PR though.

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
+                    }
                     Some(_) => {}
                     None => return,
                 },
-                _event = self.pending_inbound.next(), if !self.pending_inbound.is_empty() => {}
-                event = self.pending_outbound.next(), if !self.pending_outbound.is_empty() => {
+
+                _ = interval.tick() => {
+                    for (peer, sink) in self.outbound_sinks.iter_mut() {
+                        let payload = vec![0u8; 32];
```
haikoschol: We should keep this `TODO` comment:

> `// TODO: https://github.com/paritytech/litep2p/issues/134 generate random payload and verify it`

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
+
+                        self.ping_times.insert(*peer, Instant::now());
+                        tracing::trace!(target: LOG_TARGET, ?peer, "sending ping");
+
+                        if let Err(error) = sink.send(Bytes::from(payload)).await {
+                             tracing::debug!(target: LOG_TARGET, ?peer, ?error, "failed to send ping");
+
+                        }
```
haikoschol:
```suggestion
                        tracing::trace!(target: LOG_TARGET, ?peer, "sending ping");

                        if let Err(error) = sink.send(Bytes::from(payload)).await {
                             tracing::debug!(target: LOG_TARGET, ?peer, ?error, "failed to send ping");
                        } else {
                             self.ping_times.insert(*peer, Instant::now());
                        }
```

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
+                        Err(error) => {
+                            tracing::debug!(target: LOG_TARGET, ?peer, ?error, "ping substream closed/error");
+                            self.outbound_streams.remove(&peer);
+                            self.outbound_sinks.remove(&peer);
+                            self.ping_times.remove(&peer);
+                        }
+                    }
+                }
+
+                // Handle Outbound Responses (Ping is expected here)
```
haikoschol:
```suggestion
                // Handle Inbound Pings
```

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
+                Some((id, event)) = self.inbound_streams.next() => {
+                    match event {
+                        Ok(payload) => {
+                             if let Some(sink) = self.inbound_sinks.get_mut(&id) {
```
haikoschol: Consider logging when there's no sink for a peer ID.

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
+                                 tracing::trace!(target: LOG_TARGET, ?id, "sending pong");
+                                 if let Err(error) = sink.send(payload.freeze()).await {
+                                     tracing::debug!(target: LOG_TARGET, ?id, ?error, "failed to send pong");
+                                 }
+                             }
+                        }
+                        Err(_) => {
```
haikoschol: Consider logging the error.

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,126 +94,130 @@ impl Ping {
...
                     match event {
-                        Some(Ok((peer, elapsed))) => {
-                            let _ = self
-                                .tx
-                                .send(PingEvent::Ping {
-                                    peer,
-                                    ping: elapsed,
-                                })
-                                .await;
+                        Ok(payload) => {
```
haikoschol:
```suggestion
                        Ok(_payload) => {
```

src/protocol/libp2p/ping/mod.rs
```diff
@@ -24,26 +24,26 @@ use crate::{
     error::{Error, SubstreamError},
```
haikoschol:
```suggestion
```

haikoschol (commented):
FWIW, I ran the following tests, using a litep2p listener and webrtc transport:

| Dialer | Listener -> Dialer Pings | Dialer -> Listener Pings |
|--------|---------------|----------------|
| [libp2p](https://github.com/haikoschol/webrtc-poc/tree/haiko-testing/client) |      ❌       |       ✅       |
| ["minimal smoldot env"](https://github.com/ChainSafe/smoldot/commits/haiko-libp2p-webrtc) | ✅ | ✅ |

The failing outbound pings with libp2p is an unrelated issue (#494).

src/transport/manager/mod.rs
```diff
@@ -348,6 +348,7 @@ impl TransportManager {
             self.next_substream_id.clone(),
             self.transport_manager_handle.clone(),
             keep_alive_timeout,
+            true,
```
haikoschol: This should be `false` in case the protocol being registered is ping, right?

src/protocol/libp2p/ping/mod.rs
haikoschol: @dharjeezy when I click on "Resolve comment", I get the error "Failed to resolve thread". Can you resolve it?

dharjeezy:
@lexnv @haikoschol  @dmitry-markin  please help review

haikoschol (approved):

src/protocol/transport_service.rs
```diff
@@ -298,6 +300,7 @@ impl TransportService {
         next_substream_id: Arc<AtomicUsize>,
         transport_handle: TransportManagerHandle,
         keep_alive_timeout: Duration,
+        counts_towards_keep_alive: bool,
```
haikoschol: @dharjeezy I just noticed that the addition of this parameter broke some tests:

```
litep2p$ cargo test --workspace

[...]

error[E0061]: this function takes 7 arguments but 6 arguments were supplied
    --> src/protocol/libp2p/kademlia/mod.rs:1287:40
     |
1287 |           let (transport_service, _tx) = TransportService::new(
     |  ________________________________________^^^^^^^^^^^^^^^^^^^^^-
1288 | |             peer,
1289 | |             ProtocolName::from("/kad/1"),
1290 | |             Vec::new(),
...    |
1293 | |             KEEP_ALIVE_TIMEOUT,
1294 | |         );
     | |_________- argument #7 of type `bool` is missing
     |
note: associated function defined here
    --> src/protocol/transport_service.rs:296:19
     |
 296 |     pub(crate) fn new(
     |                   ^^^
...
 303 |         counts_towards_keep_alive: bool,
     |         -------------------------------
help: provide the argument
     |
1287 |         let (transport_service, _tx) = TransportService::new(
 ...
1293 |             KEEP_ALIVE_TIMEOUT,
1294 ~             /* bool */,
1295 ~         );
     |

error[E0061]: this function takes 7 arguments but 6 arguments were supplied
   --> src/protocol/notification/tests/mod.rs:63:35
    |
 63 |       let (transport_service, tx) = TransportService::new(
    |  ___________________________________^^^^^^^^^^^^^^^^^^^^^-
 64 | |         peer,
 65 | |         ProtocolName::from("/notif/1"),
 66 | |         Vec::new(),
...   |
 69 | |         KEEP_ALIVE_TIMEOUT,
 70 | |     );
    | |_____- argument #7 of type `bool` is missing
    |
note: associated function defined here
   --> src/protocol/transport_service.rs:296:19
    |
296 |     pub(crate) fn new(
    |                   ^^^
...
303 |         counts_towards_keep_alive: bool,
    |         -------------------------------
help: provide the argument
    |
 63 |     let (transport_service, tx) = TransportService::new(
...
 69 |         KEEP_ALIVE_TIMEOUT,
 70 ~         /* bool */,
 71 ~     );
    |

error[E0061]: this function takes 7 arguments but 6 arguments were supplied
   --> src/protocol/request_response/tests.rs:61:35
    |
 61 |       let (transport_service, tx) = TransportService::new(
    |  ___________________________________^^^^^^^^^^^^^^^^^^^^^-
 62 | |         peer,
 63 | |         ProtocolName::from("/notif/1"),
 64 | |         Vec::new(),
...   |
 67 | |         KEEP_ALIVE_TIMEOUT,
 68 | |     );
    | |_____- argument #7 of type `bool` is missing
    |
note: associated function defined here
   --> src/protocol/transport_service.rs:296:19
    |
296 |     pub(crate) fn new(
    |                   ^^^
...
303 |         counts_towards_keep_alive: bool,
    |         -------------------------------
help: provide the argument
    |
 61 |     let (transport_service, tx) = TransportService::new(
...
 67 |         KEEP_ALIVE_TIMEOUT,
 68 ~         /* bool */,
 69 ~     );
    |

error[E0061]: this function takes 7 arguments but 6 arguments were supplied
   --> src/protocol/transport_service.rs:673:33
    |
673 |           let (service, sender) = TransportService::new(
    |  _________________________________^^^^^^^^^^^^^^^^^^^^^-
674 | |             peer,
675 | |             ProtocolName::from("/notif/1"),
676 | |             Vec::new(),
...   |
679 | |             KEEP_ALIVE_TIMEOUT,
680 | |         );
    | |_________- argument #7 of type `bool` is missing
    |
note: associated function defined here
   --> src/protocol/transport_service.rs:296:19
    |
296 |     pub(crate) fn new(
    |                   ^^^
...
303 |         counts_towards_keep_alive: bool,
    |         -------------------------------
help: provide the argument
    |
673 |         let (service, sender) = TransportService::new(
...
679 |             KEEP_ALIVE_TIMEOUT,
680 ~             /* bool */,
681 ~         );
    |

```

pls fix

haikoschol: @dharjeezy Thanks! Again, I can't resolve this for whatever reason. If you can, please do.

src/protocol/libp2p/ping/config.rs
```diff
@@ -104,13 +117,19 @@ impl ConfigBuilder {
         self
     }
```
lexnv, line 119: Nit: Is cargo doc check ignoring missings docs on public methods? We should also document this and state the defaults (same for above in public interfaces)

src/protocol/libp2p/ping/mod.rs
```diff
@@ -85,93 +96,67 @@ impl Ping {
...
+
+        self.inbound_streams.insert(peer, stream);
+        self.inbound_sinks.insert(peer, sink);
     }
 
     /// Start [`Ping`] event loop.
     pub async fn run(mut self) {
         tracing::debug!(target: LOG_TARGET, "starting ping event loop");
 
+        let mut interval = tokio::time::interval(self.ping_interval);
```
lexnv: nit: This will get fired immediatelly, is this intended or should we disarm the first tick?

src/protocol/libp2p/ping/mod.rs
```diff
@@ -185,26 +170,144 @@ impl Ping {
...
+                    let mut failed_outbound = Vec::new();
+
+                    for (peer, sink) in self.outbound_sinks.iter_mut() {
+                        // TODO: <https://github.com/paritytech/litep2p/issues/134>
+                        //       generate random payload and verify it.
+                        let payload = vec![0u8; 32];
+
+                        tracing::trace!(target: LOG_TARGET, ?peer, "sending ping");
+
+                        // Due to buggy implementation of ping protocol in litep2p before
```
lexnv: I would call it "Due to differences with older litep2p" since this was a decision made for simplicity i suppose

src/protocol/libp2p/ping/mod.rs
```diff
@@ -185,26 +170,144 @@ impl Ping {
...
+                        // <https://github.com/paritytech/litep2p/pull/416>,
+                        // when talking to older litep2p we will:
+                        //  1. Successfully send first ping and receive pong.
+                        //  2. Successfully send second ping, but won't receive a response.
+                        //  3. Fail to send the third ping and reopen a substream, going to (1).
+                        //
+                        // Hence `trace` below.
+                        if let Err(error) = sink.send(Bytes::from(payload)).await {
```
lexnv: This could actually cause head of line blocking if the first peer is blocking. It might be worth fixing this since it could result under heavy load or malicious peers to delay sending the ping frames. Which then given a significant delay cause the remote to disconnect?

I would move this in a futures unordered

src/protocol/libp2p/ping/mod.rs
```diff
@@ -185,26 +170,144 @@ impl Ping {
...
                                     ping: elapsed,
-                                })
-                                .await;
+                                }).await;
+                            }
+                        }
+                        Err(error) => {
+                            // `trace` because this is expected to happen with older versions of
+                            // litep2p.
+                            // TODO: why does this not happen in practice? Do we deliver substream
```
lexnv: Lets open an issue for this so we can investigate further

It looks like this invalidates the assumptions from interval tick above?

src/protocol/connection.rs
```diff
@@ -158,7 +158,14 @@ impl ConnectionHandle {
     }
 }
 
-/// Type which allows the connection to be kept open.
+/// Type which allows to keep the connection opened and not allow the keep-alive mechanism to close
+/// it.
+///
+/// The [`Permit`] is created when beginning to open a substream and passed on until it reaches
+/// [`TransportService`](crate::protocol::TransportService), where the connection is upgraded (what
```
lexnv: nit: which

src/protocol/connection.rs
```diff
@@ -158,7 +158,14 @@ impl ConnectionHandle {
...
+/// means it won't be closed) and the permit is not needed anymore and dropped.
+///
+/// Every use of a permit must be short-lived, make sure to drop it!
```
lexnv: nit: The permit is designed to be short-lived, please ensure this is dropped as soon as its no longer relevant

or similar

src/protocol/transport_service.rs
```diff
@@ -287,6 +296,8 @@ pub struct TransportService {
 
     /// Close the connection if no substreams are open within this time frame.
     keep_alive_tracker: KeepAliveTracker,
+
```
lexnv, line 307: nit: Could add a tiny doc

src/protocol/transport_service.rs
```diff
@@ -648,14 +667,21 @@ impl Stream for TransportService {
...
+                    {
                         self.keep_alive_tracker.substream_activity(peer, connection_id);
                         if let Some(context) = self.connections.get_mut(&peer) {
                             context.try_upgrade(&connection_id);
                         }
                     }
 
+                    // Connection is upgraded, we must now drop the permit.
+                    // This is for the reader, not for compiler.
+                    drop(permit);
```
lexnv: Could the try upgrade still fail now?

src/protocol/protocol_set.rs
```diff
@@ -127,6 +127,10 @@ pub enum InnerTransportEvent {
 
         /// Substream.
         substream: Substream,
+
+        /// Permit that was held while this substream was opening. Must be dropped by
+        /// [`TransportService`](crate::protocol::TransportService) once connection is upgraded.
+        permit: Permit,
```
lexnv: How I understand this:
- we decouple permits from substream
- whenever a substream is received we upgrade the connection (try_upgrade) which may also fail silently (but we hold the permit so that should be fine?)
- we are free to drop the permit
- the transport service checks periodically (5s) if a substream has been opened in this interval
- if no substreams were opened the connection is downgraded
- a downgraded connection is closed the moment the last substream exists (because it was previously holding permits)

Wouldn't this change "close on inactivity the connection once the last substream exists" to "close the connection immediately since all permits are dropped"?

src/protocol/libp2p/ping/config.rs
dmitry-markin: litep2p doesn't have `#![warn(missing_docs)]` set. But this is for another PR :)

src/protocol/libp2p/ping/mod.rs
dmitry-markin: It's fine.

src/protocol/libp2p/ping/mod.rs
dmitry-markin: It doesn't conform to the spec, so it's buggy)

src/protocol/libp2p/ping/mod.rs
dmitry-markin: Yeah, was also thinking about it. Makes sense.

src/protocol/libp2p/ping/mod.rs
dmitry-markin: > It looks like this invalidates the assumptions from interval tick above?

Why? `send()` in interval tick fails, while I would expect it to fail here if the substream reset events were delivered.

src/protocol/connection.rs
haikoschol: "where" seems fine here. Or "which upgrades the connection".

src/protocol/protocol_set.rs
dmitry-markin: Yes, this is what makes me worry about Notifications substreams not keeping the connection open anymore.

May be there is a better way to not cause ping/identify substreams keep the connection open? I.e., `SubstreamKeepAlive::Yes` makes the substream store the permit, and `SubstreamKeepAlive::No` to drop it as soon as substream was opened? What do you think?

src/protocol/transport_service.rs
dmitry-markin: No, because we are guaranteed to hold the valid permit.

src/protocol/protocol_set.rs
lexnv: Yep this approach looks good to me 🙏

dmitry-markin: Done

src/protocol/libp2p/ping/mod.rs
dmitry-markin: Done in https://github.com/paritytech/litep2p/pull/416/commits/b8d73cff66fc66c501b64661c88722fef4bedb37.

Polkadot-Forum:

https://forum.polkadot.network/t/a-deep-dive-into-overhauling-litep2ps-ping-protocol-to-support-periodic-pinging-spec-compliant-payloads-and-a-new-substreamkeepalive-mechanism-for-connection-lifecycle-management/17501/1

PR #422, opened 2025-08-01
chore: update str0m dependency, update code based on breaking changes

### LINKED ISSUE #421: webrtc: Update dependencies
We are using outdated dependencies to implement the webrtc transport layer.

The first step we should take is to ensure that we are running with the latest versions on crates.io:

https://github.com/paritytech/litep2p/blob/077d43598530fd4a988a47d989fcc97527c87c89/Cargo.toml#L66-L68

Ensure that all crates required for the WebRTC implementation are updated to the latest version.

This might not be as trivial as bumping the version, since str0m might contain subtle or direct api breaking changes. It might be required to adjust the str0m crate upstream if it turn out we need more details (that are now private behind the API).

timwu20:
Closes #421

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3484,9 +3484,9 @@ checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "str0m"
-version = "0.6.3"
+version = "0.9.0"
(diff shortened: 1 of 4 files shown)


dmitry-markin (approved):

lexnv (approved):
Nicely done @timwu20! Thanks again for contributing 🙏

dmitry-markin:
P.S. Also, can you avoid force-pushing to preserve the review flow?

PR #423, opened 2025-08-05
Support RSA remote network identity keys

dmitry-markin:
Support remote RSA network identity keys to handle connections to nodes with such keys. The support is gated behind `rsa` feature flag and deliberately doesn't enable the use of RSA keys as a local network identity keys. This preserves backward compatibility and allows for the use of lighter `x509-parser` dependency supporting only X.509 parsing and not serialization.

### DIFF
--- a/src/crypto/tls/certificate.rs
+++ b/src/crypto/tls/certificate.rs
@@ -458,7 +458,7 @@ mod tests {
 
         assert!(parsed_cert.verify().is_ok());
         assert_eq!(
-            crate::crypto::PublicKey::Ed25519(keypair.public()),
+            crate::crypto::RemotePublicKey::Ed25519(keypair.public()),
             parsed_cert.extension.public_key
         );
     }


src/crypto/tls/certificate.rs
```diff
@@ -102,10 +102,13 @@ pub struct P2pCertificate<'a> {
...
 /// and a signature performed using the private host key.
 pub struct P2pExtension {
-    public_key: PublicKey,
+    public_key: RemotePublicKey,
     /// This signature provides cryptographic proof that the peer was
     /// in possession of the private host key at the time the certificate was signed.
     signature: Vec<u8>,
+    /// PeerId derived from the public key. While not being part of the extension, we store it to
+    /// avoid the need to serialize the public key back to protobuf.
+    peer_id: PeerId,
```
lexnv, line 111: There's no way of deriving the PeerID from the `RemotePublicKey` struct without going to the protobuf encoding? And therefore, we need to cache it here? 

dq: Would `public_key.to_peer_id()` suffice here?

src/crypto/mod.rs
```diff
@@ -110,3 +91,57 @@ impl From<ed25519::PublicKey> for PublicKey {
...
+
+    /// Decode a public key from a protobuf structure, e.g. read from storage
+    /// or received from another node.
+    pub fn from_protobuf_encoding(bytes: &[u8]) -> Result<RemotePublicKey, ParseError> {
+        use prost::Message;
+
+        let pubkey = keys_proto::PublicKey::decode(bytes)?;
+
+        pubkey.try_into()
+    }
```
lexnv, line 128: dq: Couldn't we implement the `to_peer_id` here? It looks like this works for ed25519, is there a limitation with rsa?

lexnv (approved):
LGTM! 🚀

src/crypto/mod.rs
dmitry-markin: This is the whole reason why `RemotePublicKey` is introduced. With RSA lighter `x509-parser` doesn't support serializing RSA keys, and we need serialization to derive peer ID from the key. There is a heavier library, `x509-cert` that supports the serialization, but its encoder returns `Result` and this would require completely rewriting everything that relies on `to_peer_id` being infallible.

If we only support remote RSA keys, there is no need to encode X.509, but a separate `RemotePublicKey` type is needed.

src/crypto/tls/certificate.rs
dmitry-markin: `RemotePublicKey` deliberately doesn't support `to_peer_id` as explained in the other comment.

src/crypto/mod.rs
lexnv: That totally makes sense! Thanks for the extra context, the code looks solid! Feel free to merge anytime 🙏

PR #426, opened 2025-08-13
Make `cargo clippy` happy

dmitry-markin:
Make `cargo clippy` & `cargo check` happy in default and `--all-features` configurations.

This is needed due to stable toolchain update and clippy now discovering more issues than before in https://github.com/paritytech/litep2p/pull/422.

### DIFF
--- a/src/crypto/tls/certificate.rs
+++ b/src/crypto/tls/certificate.rs
@@ -126,7 +126,7 @@ pub struct VerificationError(#[from] pub(crate) webpki::Error);
 /// Internal function that only parses but does not verify the certificate.
 ///
 /// Useful for testing but unsuitable for production.
-fn parse_unverified(der_input: &[u8]) -> Result<P2pCertificate, webpki::Error> {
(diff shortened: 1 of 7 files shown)


src/crypto/tls/certificate.rs
```diff
@@ -126,7 +126,7 @@ pub struct VerificationError(#[from] pub(crate) webpki::Error);
 /// Internal function that only parses but does not verify the certificate.
 ///
 /// Useful for testing but unsuitable for production.
-fn parse_unverified(der_input: &[u8]) -> Result<P2pCertificate, webpki::Error> {
+fn parse_unverified<'a>(der_input: &'a [u8]) -> Result<P2pCertificate<'a>, webpki::Error> {
```
lexnv, line 129:
```suggestion
fn parse_unverified(der_input: &[u8]) -> Result<P2pCertificate<'_>, webpki::Error> {
```

dmitry-markin: I would keep it explicit — this is how internal functions returning wrapped references are defined (`from_der`), and this way its more obvious that `P2pCertificate` contains supplied slice with its lifetime.

PR #427, opened 2025-08-18
kademlia: Track success of `PUT_VALUE` queries

dmitry-markin:
Track success of `PUT_VALUE` queries and emit `PutRecordSuccess` event.

This PR add checking of the final stage of `PUT_VALUE` query — storing records at the target peers. When integrating the changes into polkadot-sdk, it should be taken into account that queries that previously unconditionally succeeded can (and will) now fail when not all the target peers return `PUT_VALUE` ACK. In order to keep authority-discovery operational it is advised to reduce the quorum of queries from "all" to 50% of target peers.

### DIFF
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -1304,18 +1304,15 @@ impl Kademlia {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashSet;
-
     use super::*;
     use crate::{
         codec::ProtocolCodec,
-        crypto::ed25519::Keypair,
         transport::{
-            manager::{limits::ConnectionLimitsConfig, TransportManager},
+            manager::{TransportManager, TransportManagerBuilder},
             KEEP_ALIVE_TIMEOUT,
         },
         types::protocol::ProtocolName,
-        BandwidthSink, ConnectionId,
+        ConnectionId,
     };
     use multiaddr::Protocol;
     use multihash::Multihash;
@@ -1329,13 +1326,7 @@ mod tests {
     }
 
     fn make_kademlia() -> (Kademlia, Context, TransportManager) {
-        let (manager, handle) = TransportManager::new(
-            Keypair::generate(),
-            HashSet::new(),
-            BandwidthSink::new(),
-            8usize,
-            ConnectionLimitsConfig::default(),
-        );
+        let manager = TransportManagerBuilder::new().build();
 
         let peer = PeerId::random();
         let (transport_service, _tx) = TransportService::new(
@@ -1343,7 +1334,7 @@ mod tests {
             ProtocolName::from("/kad/1"),
             Vec::new(),
             Default::default(),
-            handle,
+            manager.transport_manager_handle(),
             KEEP_ALIVE_TIMEOUT,
         );
         let (event_tx, event_rx) = channel(64);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.1] - 2025-10-28
+
+This release ensures that polling the yamux controller after an error does not lead to unexpected behavior.
+
+### Fixed
+
+- yamux/control: Ensure poll next inbound is not called after errors  ([#445](https://github.com/paritytech/litep2p/pull/445))
+
 ## [0.11.0] - 2025-10-20
 
 This release adds support for RSA remote network identity keys gated behind `rsa` feature. It also fixes mDNS initialization in the environment with no multicast addresses available and Bitswap compatibility with kubo IPFS client >= v0.37.0.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1822,7 +1822,7 @@ checksum = "241eaef5fd12c88705a01fc1066c48c4b36e0dd4377dcdc7ec3942cea7a69956"
 
 [[package]]
 name = "litep2p"
-version = "0.11.0"
+version = "0.11.1"
 dependencies = [
  "async-trait",
  "bs58 0.5.1",
@@ -1873,7 +1873,7 @@ dependencies = [
  "webpki",
  "x25519-dalek 2.0.1",
  "x509-parser 0.17.0",
- "yamux 0.13.7",
+ "yamux 0.13.8",
  "yasna",
  "zeroize",
 ]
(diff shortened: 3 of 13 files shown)


src/protocol/libp2p/kademlia/query/mod.rs
```diff
@@ -428,6 +460,29 @@ impl QueryEngine {
         query_id
     }
 
+    /// Start `PUT_VALUE` requests tracking.
+    pub fn start_put_record_to_found_nodes_requests_tracking(
```
lexnv, line 464: Nice! IIUC, these changes are only impacting the newly added API via `start_put_record_to_found_nodes_requests_tracking` call?

dmitry-markin: Yes, this PR only affects the last stage of `PUT_VALUE` — tracking individual requests to target peers after calling this function.

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -472,20 +472,44 @@ impl Kademlia {
...
-                let _ = self.event_tx.send(KademliaEvent::IncomingRecord { record }).await;
-            }
+                    if let IncomingRecordValidationMode::Automatic = self.validation_mode {
+                        self.store.put(record.clone());
+                    }
+
+                    // Send ACK even if the record was/will be filtered out to not reveal any
+                    // internal state.
+                    let message = KademliaMessage::put_value(record.clone());
+                    self.executor.send_message(peer, message, substream);
```
lexnv, line 509: dq: Is libp2p sending the ACK as well? Im wondering if this is really needed, since we clone the record here and use some bandwidth?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -472,20 +472,44 @@ impl Kademlia {
...
+                        "handle `PUT_VALUE` response",
+                    );
 
-                if let IncomingRecordValidationMode::Automatic = self.validation_mode {
-                    self.store.put(record.clone());
+                    self.engine.register_response(
+                        query_id,
+                        peer,
+                        KademliaMessage::PutValue {
+                            record: record.clone(),
```
lexnv: nit: Could we avoid the clone here? We should have access to the record

src/protocol/libp2p/kademlia/query/put_record.rs
```diff
@@ -0,0 +1,122 @@
...
+    peers_to_succeed: usize,
+
+    /// Peers we're waiting for responses from.
+    pending_peers: HashSet<PeerId>,
+
+    /// Successfully responded peers.
+    successful_peers: HashSet<PeerId>,
+
+    /// Failed peers.
+    failed_peers: HashSet<PeerId>,
```
lexnv: nit: Do we need these peers around? Offhand, I can't think of an use-case for this

src/protocol/libp2p/kademlia/query/put_record.rs
```diff
@@ -0,0 +1,122 @@
...
+                Quorum::All => cmp::max(peers.len(), 1),
+            },
+            pending_peers: peers.into_iter().collect(),
+            successful_peers: HashSet::new(),
+            failed_peers: HashSet::new(),
+        }
+    }
+
+    /// Register successful response from peer.
+    pub fn register_response(&mut self, peer: PeerId) {
```
lexnv, line 69: nit: I also would add a trace log here and on request resp failure (unless we can deduce this info otherwise)

src/protocol/libp2p/kademlia/query/put_record.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Get next action if the context is finished.
+    pub fn next_action(&self) -> Option<QueryAction> {
+        if self.is_finished() {
+            if self.is_succeded() {
+                Some(QueryAction::QuerySucceeded { query: self.query })
+            } else {
+                Some(QueryAction::QueryFailed { query: self.query })
+            }
+        } else {
+            None
```
lexnv, line 127: nit: I find it easier to read `if self.is_finished { return None }`, up to you (I know this looks a bit like cpp :D)

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -803,20 +827,26 @@ impl Kademlia {
...
+                    ?query,
                     record_key = ?record.key,
                     num_peers = ?peers.len(),
                     "store record to found peers",
                 );
                 let key = record.key.clone();
                 let message = KademliaMessage::put_value(record);
 
-                for peer in peers {
+                for peer in &peers {
```
lexnv, line 847: nit; I got a bit concerned about the perf aspect of cloning the message. It looks like the message.clone() is cheaply clonable, I would leave a comment here not to look into it if we ever want to optimize Kademlia

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: Yes, this is what libp2p does. I don't understand either, why the whole message is sent back instead of a small ACK message.

Due to these ACKs not currentty being sent by litep2p we will likely need a two-stage upgrade process: first start sending out ACKs and use some alternative mechanism for determining if the record was received (not relying on ACKs being received), and after most of the network upgrades switch to using received ACKs. I am going to implement "stage 1" in a follow-up PR.

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: Indeed, I wonder why clippy didn't catch this.

src/protocol/libp2p/kademlia/query/put_record.rs
dmitry-markin: Good catch! I left only tracking of pending peers and the number of succeeded peers, this is enough.

src/protocol/libp2p/kademlia/query/put_record.rs
dmitry-markin: Done.

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: It's `Bytes`, so `message.clone()` might be not super cheaply clonable if the record is big.

dmitry-markin:
@lexnv thanks for the review! All comments make sense, suggestions applied.

As for ACKs, this is part of the protocol, and litep2p currently doesn't behave up to spec. This will cause some difficulties when upgrading — see the inline comment for the explanation.

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: Aah, you are right, `Bytes` is reference counted :+1:

dmitry-markin: Done

Issue #429, opened 2025-08-28
Remove workaround for handling missing `PUT_VALUE` ACKs


dmitry-markin:
Due to litep2p not previously implementing `PUT_VALUE` ACK messages, PR /TBD/ implements a hackish way of dealing with such older nodes. It does two things:

1. Tracks `PUT_VALUE` send successes instead of ACK responses to determine the query success.
2. Ignores send errors when the remote doesn't read the `PUT_VALUE` ACK message.

Once most of the network is on a newer litep2p version, this hackish implementation must be replaced with a proper `PUT_VALUE` ACK handling.

PR #430, opened 2025-08-28
kademlia: Workaround for dealing with not implemented `PUT_VALUE` ACKs

dmitry-markin:
Because previously litep2p didn't send/receive `PUT_VALUE` ACKs, we can't just enable this mechanism — this would lead to failures when older remote peers don't behave up to spec. This PR deals with this situation by implementing a workaround to handle such peers:

1. Track `PUT_VALUE` send successes instead of ACK responses to determine the query success.
2. Ignore send errors when the remote doesn't read the `PUT_VALUE` ACK message.

This should be seen as a temporary measure until most of the networks updates to litep2p with `PUT_VALUE` ACK support (this PR or newer). Suggested timing is in a year after releasing the version with this PR. This is tracked by issue https://github.com/paritytech/litep2p/issues/429.

Builds upon https://github.com/paritytech/litep2p/pull/427.

### DIFF
--- a/src/protocol/libp2p/kademlia/executor.rs
+++ b/src/protocol/libp2p/kademlia/executor.rs
@@ -149,7 +149,7 @@ impl QueryExecutor {
 
     /// Send message and ignore sending errors.
     ///
-    /// This is a hackish way of dealing with older litep2p nodes not exppecting receiving
+    /// This is a hackish way of dealing with older litep2p nodes not expecting receiving
     /// `PUT_VALUE` ACK messages. This should eventually be removed.
     // TODO: remove this as part of https://github.com/paritytech/litep2p/issues/429.
     pub fn send_message_eat_failure(
--- a/src/protocol/libp2p/kademlia/mod.rs
+++ b/src/protocol/libp2p/kademlia/mod.rs
@@ -100,6 +100,16 @@ enum PeerAction {
     SendAddProvider(QueryId, Bytes),
 }
 
+impl PeerAction {
+    fn query_id(&self) -> QueryId {
+        match self {
+            PeerAction::SendFindNode(query_id) => *query_id,
+            PeerAction::SendPutValue(query_id, _) => *query_id,
+            PeerAction::SendAddProvider(query_id, _) => *query_id,
+        }
+    }
+}
+
 /// Peer context.
 #[derive(Default)]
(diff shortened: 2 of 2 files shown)


src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -291,13 +293,26 @@ impl Kademlia {
...
+        // Apart from the failing query, we need to fail all other pending queries for the peer
+        // being disconnected.
         if let Some(PeerContext { pending_actions }) = self.peers.remove(&peer) {
             pending_actions.into_iter().for_each(|(_, action)| {
-                if let PeerAction::SendFindNode(query_id) = action {
-                    self.engine.register_response_failure(query_id, peer);
+                let query_id = match action {
+                    // `SendFindNode` includes `FIND_NODE`, `GET_VALUE` and `GET_PROVIDERS`
+                    // queries.
+                    PeerAction::SendFindNode(query_id)
```
lexnv: nit: Maybe we could add the `fn query_id()` in the PeerAction?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -291,13 +293,26 @@ impl Kademlia {
...
+                    | PeerAction::SendPutValue(query_id, _)
+                    | PeerAction::SendAddProvider(query_id, _) => query_id,
+                };
+
+                // Don't report failure twice for the same `query_id` if it was already reported
+                // above. (We can still have other pending queries for the peer that
+                // need to be reported.)
+                if Some(query_id) != query {
```
lexnv, line 317: dq: Doesn't the `if Some(query_id)` shadows the previous `let query_id = peeraction.query_id()`?

src/protocol/libp2p/kademlia/executor.rs
```diff
@@ -90,24 +113,69 @@ impl QueryExecutor {
...
+                    query_id,
+                    result: QueryResult::SendSuccess { substream },
+                },
+            }
+        }));
+    }
+
+    /// Send message and ignore sending errors.
+    ///
+    /// This is a hackish way of dealing with older litep2p nodes not exppecting receiving
```
lexnv: nit: typo in exppecting

PR #432, opened 2025-09-05
kademlia: Track success of `ADD_PROVIDER` queries

dmitry-markin:
Track success of `ADD_PROVIDER` queries and emit `AddProviderSuccess` & `QueryFailed` events.

This PR adds tracking of the last stage of `ADD_PROVIDER` query: putting provider records to the target peers. Because libp2p spec doesn't have `ADD_PROVIDER` ACK messages, we track only that the requests were sent out to the target peers without errors. `Quorum` is respected when determining whether the query was successful or not.

Builds upon https://github.com/paritytech/litep2p/pull/430.

### DIFF
--- a/src/protocol/libp2p/kademlia/query/mod.rs
+++ b/src/protocol/libp2p/kademlia/query/mod.rs
@@ -88,7 +88,7 @@ enum QueryType {
 
     /// `PUT_VALUE` message sending phase.
     PutRecordToFoundNodes {
-        /// Context for tracking `PUT_VALUE` requests.
+        /// Context for tracking `PUT_VALUE` responses.
         context: PutToTargetPeersContext,
     },
 
@@ -498,7 +498,7 @@ impl QueryEngine {
             target: LOG_TARGET,
             ?query_id,
             num_peers = ?peers.len(),
-            "start `PUT_VALUE` progress tracking"
+            "start `PUT_VALUE` responses tracking"
         );
 
         self.queries.insert(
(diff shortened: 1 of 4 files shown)


src/protocol/libp2p/kademlia/store.rs
```diff
@@ -1,5 +1,3 @@
-// Copyright 2023 litep2p developers
```
lexnv, line 1: nit: Should we do the same for all other files and just keep the copyright without the year?

src/protocol/libp2p/kademlia/message.rs
```diff
@@ -28,14 +28,15 @@ use crate::{
...
+use enum_display::EnumDisplay;
 use prost::Message;
 use std::time::{Duration, Instant};
 
 /// Logging target for the file.
 const LOG_TARGET: &str = "litep2p::ipfs::kademlia::message";
 
 /// Kademlia message.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, EnumDisplay)]
```
lexnv, line 39: dq: This is used to display the message in the logs?

src/protocol/libp2p/kademlia/store.rs
dmitry-markin: Actually, accidentally deleted these lines. Will revert.

Yes, we can live without a year as in polkadot-sdk.

src/protocol/libp2p/kademlia/query/mod.rs
```diff
@@ -503,7 +549,7 @@ impl QueryEngine {
             Some(QueryType::PutRecordToPeers { context, .. }) => {
                 context.register_response_failure(peer);
             }
-            Some(QueryType::PutRecordToFoundNodes { context, .. }) => {
+            Some(QueryType::PutRecordToFoundNodes { context }) => {
```
lexnv, line 552: nit: Maybe we can introduce a `fn context()` on the `QueryType`, such that we can turn this code into:

```rust
match self.queries.get_mut(&query) {
    None => ...,

    Some(query) => query.context.register_response_failure(peer);
}
```

dmitry-markin: This won't work for a more complex `register_response` implementation, so I'll keep it as is for consistency.

PR #434, opened 2025-09-08
mDNS: Do not fail initialization if the socket could not be created

bkchr:
I was on a plane and wanted to run some zombienet tests. The tests were failing because the nodes were unable to setup the mDNS socket. This pull request solves this by moving the socket initialization into the execution loop. If the initialization failed, it is retried again in the future.


dmitry-markin (approved):

lexnv (approved):
LGTM! 🙏

Issue #437, opened 2025-09-16
tcp: Listener terminated with error causing litep2p backend to exit


lexnv:
Some Paseo validators have reported the following error:

```
Sep 06 15:54:25 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:25.480 ERROR tokio-runtime-worker litep2p::tcp: TCP listener terminated with error error=Os { code: 103, kind: ConnectionAborted, message: "Software caused connection abort" }
Sep 06 15:54:25 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:25.480 ERROR tokio-runtime-worker litep2p::transport-manager: Installed transports terminated, ignore if the node is stopping
Sep 06 15:54:25 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:25.480 ERROR tokio-runtime-worker sub-libp2p: Litep2p backend terminated
Sep 06 15:54:25 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:25.588  INFO tokio-runtime-worker sub-authority-discovery: Successfully persisted AddrCache on disk
Sep 06 15:54:26 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:26.370  WARN tokio-runtime-worker parachain::availability-recovery: Recovery of available data failed. candidate\_hash=0xf8be43ec2e00ab7210e6f864e71f5dfcc25b1137b854fc066fcfb7f549aba6ca traceID=>
Sep 06 15:54:26 vo-phx1 polkadot\[957593\]: 2025-09-06 15:54:26.370  WARN tokio-runtime-worker parachain::availability-recovery: Error during recovery: Data is unavailable
```

The TCP Listener has returned an error, which is propagated to the TCP transport as stream terminated (`Some(None)`):

https://github.com/paritytech/litep2p/blob/6723585cd95ae984f0fba1c4daf9aec00bad5251/src/transport/tcp/mod.rs#L572-L580

The downstream effect of this is that the TCP transport will terminate on a None event. Then, the litep2p backend will terminate on the None event propagated from the TCP transport.

This is effectively causing a cascade of events that terminate the litep2p backend. 

The error originates from the Tokio TCP listener:
https://github.com/paritytech/litep2p/blob/6723585cd95ae984f0fba1c4daf9aec00bad5251/src/transport/common/listener.rs#L460-L465

This might happen under unstable network conditions; however, we should retry the operation or reinitialize the listener.

cc @paritytech/networking

[...]

PR #439, opened 2025-09-23
listener: Ensure tcp/websocket do not terminate on listener errors

lexnv:
This PR ensures that both TCP and WebSocket transports do not terminate in case of listener errors.

Recoverable errors like `ConnectionAborted` (which signals that the client connection entered the queue and disconnected before we got a chance to process it) are retried.

Nonrecoverable errors lead to the `TokioListener` being dropped and never polled again (ie, removed from the transport service). 

TCP and Websocket are now robust wrt listener errors and will not terminate the main execution loop unless the listeners return None.

Closes: https://github.com/paritytech/litep2p/issues/437
cc @paritytech/networking

### DIFF
--- a/src/transport/websocket/mod.rs
+++ b/src/transport/websocket/mod.rs
@@ -607,23 +607,24 @@ impl Stream for WebSocketTransport {
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         if let Poll::Ready(event) = self.listener.poll_next_unpin(cx) {
-            return match event {
+            match event {
                 None => {
                     tracing::error!(
                         target: LOG_TARGET,
                         "Websocket listener terminated, ignore if the node is stopping",
                     );
 
-                    Poll::Ready(None)
+                    return Poll::Ready(None);
                 }
                 Some(Err(error)) => {
                     tracing::error!(
                         target: LOG_TARGET,
                         ?error,
-                        "Websocket listener terminated with error",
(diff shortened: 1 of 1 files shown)


src/transport/common/listener.rs
```diff
@@ -455,17 +455,43 @@ impl Stream for SocketListener {
...
+                match listener.poll_accept(cx) {
+                    Poll::Pending => {
+                        break;
+                    }
+                    Poll::Ready(Ok((stream, address))) => {
+                        self.poll_index = (self.poll_index + 1) % len;
+                        return Poll::Ready(Some(Ok((stream, address))));
+                    }
+                    Poll::Ready(Err(error)) => match error.kind() {
+                        ErrorKind::ConnectionAborted | ErrorKind::ConnectionReset => {
```
dmitry-markin, line 473: I would double check that `socket.accept()` can't generate these errors continuously after a socket error — otherwise we would end up busy-looping here.

src/transport/common/listener.rs
```diff
@@ -455,17 +455,43 @@ impl Stream for SocketListener {
...
+                        }
+                        _ => {
+                            self.poll_index = (self.poll_index + 1) % len;
+
+                            tracing::error!(
+                                target: LOG_TARGET,
+                                ?error,
+                                "Fatal error for listener",
+                            );
+                            self.listeners[current] = None;
```
dmitry-markin, line 490: IMO a better way would be to terminate the process and let the restarted process listen again, than stopping accepting incoming connections completely.

Issue #420, opened 2025-07-24
webrtc: Test the connection stability of the webrtc substreams


lexnv:
Ensure that the current implementation of webRtc is stable.

This should be achieved using the following crate, which has received an update to configure the transport layers:
- https://github.com/lexnv/litep2p-perf

The `run_bandwidth.sh` can be adjusted locally to target a webrtc address (similar to the branch origin/lexnv/websocket-tests).

This should resolve any issues we have with the current WebRTC implementation and serve as a building block for future improvements.

lexnv:
Since we have determined that libp2p is not spec-compliant and some version of their webrtc is not working properly, we still should test the stability of our connections:
- 1. The litep2p-perf is generic enough to support a new project that's based on libp2p-go 
- 2. Multistream select should contain spec compliant testing to filter out remainng issues

cc @gab8i @skunert

Issue #75, opened 2024-04-17
`multistream-select` implementation used by WebRTC doesn't conform to the spec


dmitry-markin:
Currently, we encode and send all protocols at once during `multistream-select` phase when opening WebRTC stream:
https://github.com/paritytech/litep2p/blob/337538f8b68d755056fa3f8a8bcf624ae1937c12/src/multistream_select/dialer_select.rs#L273-L278

This is against libp2p [`multistream-select` spec](https://github.com/libp2p/specs/blob/master/connections/README.md#multistream-select), which states that protocols should be negotiated one-by-one.

Also, the received protocols are also expected to be sent as a list:
https://github.com/paritytech/litep2p/blob/337538f8b68d755056fa3f8a8bcf624ae1937c12/src/multistream_select/dialer_select.rs#L294-L295

This also doesn't conform to the spec.

Nevertheless, the above works with smoldot. This issue is about fixing the implementation, and, may be, filing a bug to smoldot if it doesn't conform to the spec either.

Originally discussed in https://github.com/paritytech/litep2p/pull/51#discussion_r1513022095.

Issue #45, opened 2024-02-13
WebRTC backpressure


altonen:
There is virtually no backpressure in either direction. Data channel counts are seemingly not restricted in any way. Substream on the protocol-side should block if remote is not willing to accept more data over a given data channel and the remote should be throttled if local protocol is not able to read data fast enough.

lexnv:
Part of the work in:
- https://github.com/paritytech/litep2p/pull/575

PR #441, opened 2025-10-07
webrtc: update str0m, fix multistream message decoding and encoding

timwu20:
## Changes
- updates the str0m dependency to the commit which contains fixes I co-authored (PR [#1](https://github.com/algesten/str0m/pull/711), [#2](https://github.com/algesten/str0m/pull/712))
  - fixes str0m integration based on breaking public API changes
- fixes the decoding of multistream messages to support messages without trailing linebreak
- removes trailing linebreak when encoding multistream message used for WebRTC
- contributes to https://github.com/paritytech/litep2p/issues/420

### DIFF
--- a/src/multistream_select/protocol.rs
+++ b/src/multistream_select/protocol.rs
@@ -220,6 +220,10 @@ impl Message {
 
             // Skip ahead to the next protocol.
             remaining = &tail[len..];
+            if remaining.len() == 0 {
+                // During negotiation the remote may not append a trailing newline.
+                break;
+            }
         }
 
         Ok(Message::Protocols(protocols))
@@ -230,7 +234,7 @@ impl Message {
 ///
 /// # Note
 ///
-/// This is implementation is not compliant with the multistream-select protocol spec.
+/// This implementation may not be compliant with the multistream-select protocol spec.
 /// The only purpose of this was to get the `multistream-select` protocol working with smoldot.
 pub fn webrtc_encode_multistream_message(
     messages: impl IntoIterator<Item = Message>,
@@ -248,9 +252,6 @@ pub fn webrtc_encode_multistream_message(
         let mut proto_bytes = UnsignedVarint::encode(proto_bytes)?;
         header.append(&mut proto_bytes);
     }
-
-    // For the `Message::Protocols` to be interpreted correctly, it must be followed by a newline.
-    header.push(b'\n');
 
     Ok(BytesMut::from(&header[..]))
 }
@@ -540,6 +541,23 @@ mod tests {
         assert_eq!(
             Message::decode(bytes).unwrap_err(),
             ProtocolError::InvalidMessage
+        );
+    }
+
+    #[test]
+    fn test_decode_multiple_protocols_no_trailing_newline() {
+        let raw: [u8; 38] = [
+            19, 47, 109, 117, 108, 116, 105, 115, 116, 114, 101, 97, 109, 47, 49, 46, 48, 46, 48, 10,
+            17, 47, 105, 112, 102, 115, 47, 112, 105, 110, 103, 47, 49, 46, 48, 46, 48, 10,
+        ];
+        let bytes = Bytes::copy_from_slice(&raw);
+        
+        assert_eq!(
+            Message::decode(bytes).unwrap(),
+            Message::Protocols(vec![
+                Protocol::try_from(Bytes::from_static(b"/multistream/1.0.0")).unwrap(),
+                Protocol::try_from(Bytes::from_static(b"/ipfs/ping/1.0.0")).unwrap(),
+            ])
         );
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -125,7 +125,7 @@ dependencies = [
  "nom",
  "num-traits",
  "rusticata-macros",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
  "time",
 ]
 
@@ -1025,15 +1025,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fxhash"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c31b6d751ae2c7f11320402d34e41349dd1016f8d5d45e48c4312bc8625af50c"
-dependencies = [
- "byteorder",
-]
-
-[[package]]
 name = "generator"
 version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1170,7 +1161,7 @@ dependencies = [
  "once_cell",
  "rand 0.9.2",
  "ring 0.17.14",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
  "tinyvec",
  "tokio",
  "tracing",
@@ -1193,7 +1184,7 @@ dependencies = [
  "rand 0.9.2",
  "resolv-conf",
  "smallvec",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
  "tokio",
  "tracing",
 ]
@@ -1918,7 +1909,7 @@ dependencies = [
  "snow",
  "socket2 0.5.10",
  "str0m",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
  "tokio-tungstenite",
@@ -2213,7 +2204,7 @@ dependencies = [
  "log",
  "netlink-packet-core",
  "netlink-sys",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
 ]
 
 [[package]]
@@ -2237,7 +2228,7 @@ checksum = "862f41f1276e7148fb597fc55ed8666423bebe045199a1298c3515a73ec5cdd9"
 dependencies = [
  "cc",
  "libc",
- "thiserror 2.0.12",
+ "thiserror 2.0.17",
  "winapi",
 ]
 
@@ -2807,7 +2798,7 @@ dependencies = [
  "pin-project-lite 0.2.16",
  "quinn-proto",
  "quinn-udp",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "rustls 0.20.9",
  "thiserror 1.0.69",
  "tokio",
@@ -2824,7 +2815,7 @@ dependencies = [
  "bytes",
  "rand 0.8.5",
  "ring 0.16.20",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "rustls 0.20.9",
  "slab",
  "thiserror 1.0.69",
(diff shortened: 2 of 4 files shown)


src/multistream_select/protocol.rs
```diff
@@ -220,6 +220,10 @@ impl Message {
 
             // Skip ahead to the next protocol.
             remaining = &tail[len..];
+            if remaining.len() == 0 {
+                // During negotiation the remote may not append a trailing newline.
+                break;
```
lexnv, line 225: Was this observed with smoldot integration?

src/multistream_select/protocol.rs
```diff
@@ -249,9 +253,6 @@ pub fn webrtc_encode_multistream_message(
         header.append(&mut proto_bytes);
     }
 
-    // For the `Message::Protocols` to be interpreted correctly, it must be followed by a newline.
-    header.push(b'\n');
```
lexnv, line 253: We could reach out to https://github.com/smol-dot/smoldot to align on the spec, then we could remove/simplify these functions. 

We have diverged a bit from the spec, I wonder if this may break things down the line. Would the next message be interpreted correctly?

src/multistream_select/protocol.rs
timwu20: This was observed when testing against the libp2p native WebRTC transport. During negotiation, the libp2p native WebRTC transport doesn't append an extra newline.

src/multistream_select/protocol.rs
timwu20: We have @freddyli7 working on utilizing smoldot as a client to test against the libp2p server, as well as the litep2p server.  We plan on getting it into `litep2p-perf` to run the performance behaviour.  If things are incorrect in smoldot, we will create PRs upstream to fix them.

timwu20:
Added https://github.com/paritytech/litep2p/pull/441/commits/8cf27e9fa0f512d0cd27189f720b40db17ef2991 which calls `Channel.set_buffered_amount_low_threshold` to set the threshold to 1024 bytes, and listen on this event to close the channel if the protocol behaviour intends to close the substream.

With this change I'm able to get `litep2p-perf` to run the performance behaviour using a litep2p server utilizing WebRTC transport to upload 16MB of data to a libp2p WebRTC client.  Any byte counts higher than that cause the connection to get dropped because of the lack of ICE keep-alive requests.  I believe this has to do with the lack of proper back pressure when writing on the server side.  I think we can address this further in issue #45.

src/transport/webrtc/connection.rs
```diff
@@ -631,6 +631,8 @@ impl WebRtcConnection {
             protocol: protocol.to_string(),
         });
 
+        self.rtc.channel(channel_id).unwrap().set_buffered_amount_low_threshold(1024);
```
haikoschol, line 634: When testing smoldot against a minimal litep2p-server, this caused a panic for me because `self.rtc.channel(channel_id)` returned `None`. Moving this line to [`WebRtcConnection::on_channel_opened()`](https://github.com/paritytech/litep2p/blob/5e2992ac68aa605495e7ec8896753116a932ff17/src/transport/webrtc/connection.rs#L238) prevented the panic.

src/multistream_select/protocol.rs
haikoschol: @lexnv `Message::Protocols` is the response to an `ls` command, right? According to the spec, supporting `ls` [is optional](https://github.com/multiformats/multistream-select/blob/9a1fbe8edadebea223f47102f1913b8026c23a1c/README.md?plain=1#L142) and [smoldot currently does not implement it](https://github.com/smol-dot/smoldot/blob/33d34fb8733c7d8a7e5878f8658bc7a0d0383bae/lib/src/libp2p/connection/multistream_select.rs#L387).

I made a quick attempt to remove this function and it was easy for `WebRtcDialerState::propose()` (by misusing `Message::Protocols` 😅), but it's also used [on the listener side](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/listener_select.rs#L384). It's less clear what the benefit of rewriting that code would be since it would be yet another place where length prefixes are handled.

This is my (simplified?) implementation of `WebRtcDialerState::propose()`:
```rust
pub fn propose(
    protocol: ProtocolName,
) -> crate::Result<(Self, Vec<u8>)> {
    let header = Protocol::try_from(&b"/multistream/1.0.0"[..])
        .expect("valid multitstream-select header");

    let mut protocols = vec![header];
    protocols.extend(std::iter::once(protocol.clone())
        .chain(fallback_names.clone())
        .filter_map(|protocol| Protocol::try_from(protocol.as_ref()).ok())
    );

    let mut buf = BytesMut::with_capacity(64);
    Message::Protocols(protocols).encode(&mut buf).map_err(|_| Error::InvalidData)?;

    Ok((
        Self {
            protocol,
            fallback_names,
            state: HandshakeState::WaitingResponse,
        },
        buf.freeze().to_vec(),
    ))
}
```

There are two differences between an `ls` response message and a protocol request or confirmation message with the header included. The `ls` response has:
1. a length-prefix for the entire message at the beginning
2. a trailing `\n`

Number 1 is not added by `Message::encode()` for `Message::Protocols`. Number 2 is added, but does not throw off smoldot when it expects a protocol request. But strictly speaking using `Message::Protocols` to encode a protocol request or confirmation message is not spec compliant.

src/multistream_select/protocol.rs
haikoschol: The problem is that `Message::decode()` conflates `ls` responses with protocol requests and confirmation messages. As I mentioned in [another comment](https://github.com/paritytech/litep2p/pull/441/files#r2495275465), there are small differences.

`ls` response:
```
<total length><header length>/multistream/1.0.0\n
<proto1 length>/foo/bar/proto1\n
<proto2 length>/foo/bar/proto2\n
\n
```

protocol request with fallback protocols:

```
<header length>/multistream/1.0.0\n
<proto1 length>/foo/bar/proto1\n
<proto2 length>/foo/bar/proto2\n
```

protocol confirmation:

```
<header length>/multistream/1.0.0\n
<proto1 length>/foo/bar/proto1\n
```

When using webrtc transport, there is another length prefix for the protobuf frame. But that is independent of whether the payload is a multistream-select message or not.

edit: @timwu20 pointed me towards #75 which in turn links to [this spec for multistream-select](https://github.com/libp2p/specs/blob/master/connections/README.md#multistream-select) that specifies a back-and-forth process for providing fallback protocols. I've been working off [this spec](https://github.com/multiformats/multistream-select/blob/master/README.md), which glosses over this. So I have to correct my example above: The "protocol request with fallback protocols" does not exist in this form, which means that `Message::decode()` does not need to support it. :)

lexnv: I would not add this code due to possible incompatibilities for current transports.
I see rust-libp2p is using a similar code path to what we did before:
- https://github.com/libp2p/rust-libp2p/blob/b3d31673a0ed3f45f4c95cf4865c8381dba6f471/misc/multistream-select/src/protocol.rs#L224-L227

Instead, we should get in touch with smoldot to adjust any multistream-select inconsistencies (or impl `ls` similarly to litep2p/libp2p 🙏 )

src/multistream_select/protocol.rs
lexnv: Similar to https://github.com/paritytech/litep2p/pull/441/files#r2498425079, I believe the path forward is to align both litep2p <-> smoldot to the expected multistream-select behavior.

From the litep2p perspective, it would mean removing this `webrtc_encode_multistream_message` entirely and relying on the normal DialerState, as with TCP/WebSocket.

The WebRTC code was written quickly as a PoC to validate litep2p interoperability with smoldot; however, it used a few hackish methods that deviated from the spec, which should not be present in production environments (unless technically impossible to avoid).

From the smoldot perspective, we'd need several patches to align with the WebSocket behavior

cc @dmitry-markin

haikoschol: What concretely would you like to see changed on the smoldot side? I believe its current behaviour is in line with the spec.

IMHO, the custom code in litep2p is necessary due to the protobuf framing used in webrtc transport. I agree that it would be nice to handle this in a cleaner way, but that shouldn't require changes in smoldot.

src/multistream_select/protocol.rs
haikoschol: According to [Tim's comment](https://github.com/paritytech/litep2p/pull/441/files#r2411061194), he added this when testing webrtc transport between litep2p and rust-libp2p.

I just tested litep2p <-> smoldot without this change and it doesn't work either. The reason why rust-libp2p <-> smoldot works without this change is that `MessageIO` yields one `Message` at a time, even if the remote sent both header and protocol name in a single webrtc protobuf frame.

@lexnv I completely agree with your concern that we must not break backwards compatibility for other transports. I see two alternatives to including this change:

1. Write webrtc custom code for decoding multistream-select messages analogous to `webrtc_encode_multistream_message`.
2. Get rid of all the webrtc custom code and use `MessageIO`, `DialerSelectFuture`, `ListenerSelectFuture`, etc.

I know that you prefer option 2 based on [your comment here](https://github.com/paritytech/litep2p/pull/441/files#r2498449451) and I'm happy to work on that. But I'm also expected to make progress on litep2p <-> smoldot interop and don't have a clear picture of how to implement option 2 yet. 🫤

haikoschol: There's a third option: Adding the trailing newline in `webrtc_listener_negotiate()` before passing it to `Message::decode()`.

```rust
pub fn webrtc_listener_negotiate<'a>(
    supported_protocols: &'a mut impl Iterator<Item = &'a ProtocolName>,
    payload: Bytes,
) -> crate::Result<ListenerSelectResult> {
    let payload = if payload.len() > 2 && payload[0..payload.len() - 2] != b"\n\n"[..] {
        let mut buf = BytesMut::from(payload);
        buf.extend_from_slice(b"\n");
        buf.freeze()
    } else {
        payload
    };

    let Message::Protocols(protocols) = Message::decode(payload).map_err(|_| Error::InvalidData)?
```

@lexnv Is this acceptable to you as a short-term solution if we implement option 2 once we have achieved interoperability?

lexnv: The third option is adding the quick workaround without modifying the MultiStream Select core logic. 

I'm mostly concerned about subtle changes that will lead to litep2p <-> old litep2p validators misbehaving and causing wide connectivity issues. When this happens, usually a higher level starts to manifest undefined behaviors. The load we are experiencing in kusama / polkadot is not reproducible by our testing setups yet, and spawning large test nets is unsustainable budget-wise 🙏

lexnv:
@haikoschol are you ok with closing this PR in favor of:
- https://github.com/paritytech/litep2p/pull/465

They look similar to me and we could continue the discussions there 🙏 

Feel free to reopen if need be

PR #442, opened 2025-10-07
introduce max message size in webrtc config

### LINKED ISSUE #352: webrtc: Optimize decoding with appropriate buffer size
https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/transport/webrtc/util.rs#L74-L76

dharjeezy:
closes #352

### DIFF
--- a/src/transport/webrtc/config.rs
+++ b/src/transport/webrtc/config.rs
@@ -32,6 +32,7 @@ pub struct Config {
     ///
     /// How many datagrams can the buffer between `WebRtcTransport` and a connection handler hold.
     pub datagram_buffer_size: usize,
+    pub max_message_size: usize,
 }
 
 impl Default for Config {
@@ -41,6 +42,7 @@ impl Default for Config {
                 .parse()
                 .expect("valid multiaddress")],
             datagram_buffer_size: 2048,
+            max_message_size: 512 * 1024,
         }
     }
 }
--- a/src/transport/webrtc/util.rs
+++ b/src/transport/webrtc/util.rs
@@ -72,9 +72,9 @@ impl WebRtcMessage {
     }
 
     /// Decode payload into [`WebRtcMessage`].
-    pub fn decode(payload: &[u8]) -> Result<Self, ParseError> {
+    pub fn decode(payload: &[u8], max_message_size: usize) -> Result<Self, ParseError> {
         // TODO: https://github.com/paritytech/litep2p/issues/352 set correct size
-        let mut codec = UnsignedVarint::new(None);
+        let mut codec = UnsignedVarint::new(Some(max_message_size));
         let mut data = bytes::BytesMut::from(payload);
         let result = codec
             .decode(&mut data)
@@ -95,10 +95,12 @@ impl WebRtcMessage {
 mod tests {
     use super::*;
 
+    const TEST_MAX_SIZE: usize = 512 * 1024;
+
     #[test]
     fn with_payload_no_flags() {
         let message = WebRtcMessage::encode("Hello, world!".as_bytes().to_vec());
-        let decoded = WebRtcMessage::decode(&message).unwrap();
+        let decoded = WebRtcMessage::decode(&message, TEST_MAX_SIZE).unwrap();
 
         assert_eq!(decoded.payload, Some("Hello, world!".as_bytes().to_vec()));
         assert_eq!(decoded.flags, None);
@@ -107,7 +109,7 @@ mod tests {
     #[test]
     fn with_payload_and_flags() {
         let message = WebRtcMessage::encode_with_flags("Hello, world!".as_bytes().to_vec(), 1i32);
-        let decoded = WebRtcMessage::decode(&message).unwrap();
+        let decoded = WebRtcMessage::decode(&message, TEST_MAX_SIZE).unwrap();
 
         assert_eq!(decoded.payload, Some("Hello, world!".as_bytes().to_vec()));
         assert_eq!(decoded.flags, Some(1i32));
@@ -116,7 +118,7 @@ mod tests {
     #[test]
     fn no_payload_with_flags() {
         let message = WebRtcMessage::encode_with_flags(vec![], 2i32);
-        let decoded = WebRtcMessage::decode(&message).unwrap();
+        let decoded = WebRtcMessage::decode(&message, TEST_MAX_SIZE).unwrap();
 
         assert_eq!(decoded.payload, None);
         assert_eq!(decoded.flags, Some(2i32));
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -199,6 +199,8 @@ pub struct WebRtcConnection {
 
     /// Substream handles.
     handles: SubstreamHandleSet,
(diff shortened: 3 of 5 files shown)


src/transport/webrtc/config.rs
```diff
@@ -41,6 +42,7 @@ impl Default for Config {
                 .parse()
                 .expect("valid multiaddress")],
             datagram_buffer_size: 2048,
+            max_message_size: 512 * 1024,
```
dmitry-markin: Where this constant is coming from? What is libp2p way of handling this?

What will happen in case of high throughput transfer — shouldn't we be able to parse the payload split over multiple `max_message_size` buffers?

lexnv: Maybe we should make this optional, not entirely sure if this interferes with @timwu20 upcoming work 🤔

dmitry-markin: I would suggest hardcoding the constant from libp2p spec, as discussed in https://github.com/paritytech/litep2p/issues/352.

lexnv: Oki doki makes sense 🙏 Let's hardcode it instead

dharjeezy: so, @lexnv should i remove the `max_message_size` entirely from the config?

lexnv: Hey yep, let's pick: https://github.com/paritytech/litep2p/issues/352#issuecomment-3443193748 16kIB  for this

src/transport/webrtc/util.rs
```diff
@@ -74,7 +76,7 @@ impl WebRtcMessage {
     /// Decode payload into [`WebRtcMessage`].
     pub fn decode(payload: &[u8]) -> Result<Self, ParseError> {
         // TODO: https://github.com/paritytech/litep2p/issues/352 set correct size
```
lexnv:
```suggestion
```

src/transport/webrtc/util.rs
```diff
@@ -23,6 +23,8 @@ use crate::{codec::unsigned_varint::UnsignedVarint, error::ParseError, transport
 use prost::Message;
 use tokio_util::codec::{Decoder, Encoder};
 
+const MAX_MESSAGE_SIZE: usize = 16 * 1024;
```
haikoschol, line 26: There is already [a constant for this in `webrtc/substream.rs`](https://github.com/paritytech/litep2p/blob/7e048bfa4788c5998dc5e9df38c2e298ee93aff0/src/transport/webrtc/substream.rs#L39). We should either use that one here or the other way round.

Apart from that, this change covers the scope of #352, which seems to be only about optimizing memory usage.

However, it does not achieve spec compliance with respect to handling of max. message size. In order to get there, we would need to [set the `max_message_size` attribute in the SDP "remote description"](https://github.com/libp2p/specs/blob/e87cb1c32a666c2229d3b9bb8f9ce1d9cfdaa8a9/webrtc/webrtc-direct.md?plain=1#L115). Unfortunately, str0m doesn't offer an API for doing that and seems to [hard-code the value to 256 KiB](https://github.com/algesten/str0m/blob/664a6c8b40bd7035724cde7e426fc524ffb1196b/src/change/sdp.rs#L1215).

Another issue worth considering is how oversized messages are handled in the methods called by protocols, which I believe ultimately comes down to [`Substream::poll_read()`](https://github.com/paritytech/litep2p/blob/772840b1c53dc88a0b6f0808c73e7470442f2f95/src/transport/webrtc/substream.rs#L155) and [`Substream::poll_write()`](https://github.com/paritytech/litep2p/blob/772840b1c53dc88a0b6f0808c73e7470442f2f95/src/transport/webrtc/substream.rs#L195), but there is also [`write_all()`](https://github.com/paritytech/litep2p/blob/d35c711dc8d5485542ed7aed9a02b0af170f0706/src/substream/mod.rs#L366), which I have no idea what it does in the case of WebRTC substreams.

For reading, there is [this check](https://github.com/paritytech/litep2p/blob/772840b1c53dc88a0b6f0808c73e7470442f2f95/src/transport/webrtc/substream.rs#L175), which is incorrect, because the maximum length of the payload is smaller than the maximum size of the whole message. Regardless of whether it is correct, I think it is also made redundant by the changes in this PR, because oversized messages would already be rejected [here, in `WebRtcConnection::on_open_channel_data()`](https://github.com/paritytech/litep2p/blob/785a89fec191e0239ac107e4a9028e5a42ca12b1/src/transport/webrtc/connection.rs#L449).

For writing, `Substream::poll_write()` currently silently truncates oversized messages, or rather payloads that are larger than `MAX_FRAME_SIZE` (again, slightly off). This is definitely not what we want. Instead, we either want to return an error or fragment the payload into multiple messages. The latter is probably tricky to get right, especially considering the upcoming changes to the `Substream` implementation regarding flag handling. The former is in line with [the specification of the `send()` function in the WebRTC JavaScript API in browsers](https://w3c.github.io/webrtc-pc/#dom-rtcdatachannel-send):

> 5. If the byte size of data exceeds the value of maxMessageSize on channel's associated RTCSctpTransport, throw a TypeError.

The question then becomes how this would work in higher-level code implementing protocols. That code is not aware of transport-specific limitations and, I assume, expects to be able to send arbitrarily large amounts of data.

@lexnv @dmitry-markin wdyt?

haikoschol: @dharjeezy to clarify, my requested change is to use a single constant for `MAX_MESSAGE_SIZE`/`MAX_FRAME_SIZE` in both `webrtc/substream.rs` and `webrtc/util.rs`.

All the other stuff is important and needs to be addressed, but is probably out of scope for this PR.

haikoschol: Turns out [rust-libp2p also silently truncates](https://github.com/libp2p/rust-libp2p/blob/b187c14ef36744ea6b2f29740321b5fe896a50ef/misc/webrtc-utils/src/stream.rs#L208) (but [correctly calculates the maximum payload size](https://github.com/libp2p/rust-libp2p/blob/b187c14ef36744ea6b2f29740321b5fe896a50ef/misc/webrtc-utils/src/stream.rs#L55)).

lexnv: > @dharjeezy to clarify, my requested change is to use a single constant for MAX_MESSAGE_SIZE/MAX_FRAME_SIZE in both webrtc/substream.rs and webrtc/util.rs. 

Makes sens to me! Thansk for the suggestions @haikoschol

haikoschol: In the meantime I've been working on a PR to enable setting max_message_size in str0m. Will probably open it tomorrow.

haikoschol: > In the meantime I've been working on a PR to enable setting max_message_size in str0m. Will probably open it tomorrow.

Unfortunately, this does not seem as straight forward as I initially thought. Turns out the rust-libp2p folks have been there and figured out how to set the attribute with SDP munging, but from my understanding of the str0m code, the value might not be passed down to the SCTP layer. I've opened an issue instead: https://github.com/algesten/str0m/issues/844

PR #443, opened 2025-10-09
improve efficiency of propagating address failures

### LINKED ISSUE #409: TransportManager: Improve the efficiency of propagating dial failure messages
The transport manager propagates the `InnerTransportEvent::DialFailure` event together with the multiaddress to all installed protocols. In older versions of litep2p, the address was not propagated. However, we needed to have an eventually consistent address store in kademlia for optimization and consistency purposes.

https://github.com/paritytech/litep2p/blob/954d01b3afea6ee73973603a517739f3166a5603/src/transport/manager/mod.rs#L1119-L1121

We can improve the propagation of messages by propagating the addresses only to the kademlia handler. The other protocols could leave without knowing which address has failed (there was no change made that takes the address into account for other protocols).

dharjeezy:
closes #409

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -1213,7 +1213,7 @@ impl TransportManager {
                                                     "dial failure, notify protocol",
                                                 );
 
-                                                let addresses = context.dial_failure_addresses(&[address.clone()]);
+                                                let addresses = context.dial_failure_addresses(std::slice::from_ref(&address));
 
                                                 match context.tx.try_send(InnerTransportEvent::DialFailure { peer, addresses: addresses.clone() }) {
                                                     Ok(()) => {}
@@ -1666,12 +1666,14 @@ mod tests {
             Vec::new(),
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired
         );
         manager.register_protocol(
             ProtocolName::from("/notif/1"),
             Vec::new(),
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired
         );
     }
 
@@ -1686,6 +1688,7 @@ mod tests {
             Vec::new(),
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired,
         );
         manager.register_protocol(
             ProtocolName::from("/notif/2"),
@@ -1695,6 +1698,7 @@ mod tests {
             ],
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired,
         );
     }
 
@@ -1712,6 +1716,7 @@ mod tests {
             ],
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired,
         );
         manager.register_protocol(
             ProtocolName::from("/notif/2"),
@@ -1721,6 +1726,7 @@ mod tests {
             ],
             ProtocolCodec::UnsignedVarint(None),
             KEEP_ALIVE_TIMEOUT,
+            DialFailureAddresses::NotRequired,
         );
     }
 
--- a/src/protocol/protocol_set.rs
+++ b/src/protocol/protocol_set.rs
@@ -438,6 +438,7 @@ mod tests {
     use super::*;
     use crate::mock::substream::MockSubstream;
     use std::collections::HashSet;
+    use crate::transport::manager::DialFailureAddresses;
 
     #[tokio::test]
     async fn fallback_is_provided() {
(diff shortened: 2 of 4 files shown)


src/transport/manager/mod.rs
```diff
@@ -332,6 +336,7 @@ impl TransportManager {
         fallback_names: Vec<ProtocolName>,
         codec: ProtocolCodec,
         keep_alive_timeout: Duration,
+        needs_dial_failure_addresses: bool,
     ) -> TransportService {
```
lexnv, line 431: I would add an explicit enum for this option:

```rust
pub enum DialFailureAddresses {
    Required,
    NotRequired,
}
```


Such that we can easily follow during the protocol registration the purpose of this flag

src/transport/manager/mod.rs
```diff
@@ -1268,10 +1274,7 @@ impl TransportManager {
                                     for (protocol, context) in &self.protocols {
                                         let _ = match context
                                             .tx
-                                            .try_send(InnerTransportEvent::DialFailure {
-                                                peer,
-                                                addresses: addresses.clone(),
-                                            }) {
+                                            .try_send(make_dial_failure_event(&self.address_reporting_protocols, protocol, peer, addresses.clone())) {
```
lexnv: We could introduce a tiny helper over the protocol context:

```rust

impl ProtocolContext {
    fn dial_failure(&self, addresses: &[Multiaddr]) -> Vec<Multiaddr> {
        match self.dial_failure_mode {
            DialFailureMode::Required => addresses.to_vec(),
            DialFailureMode::NotRequired => Vec::new(),
        }
    }
}

 let _ = match context.tx.try_send(
    InnerTransportEvent::DialFailure {
        peer,
        addresses: context.dial_failure(addresses)
    }
 )
```

This way we avoid cloning the addresses until we need them, while keeping the `InnerTransportEvent::DialFailure` event explicit. This will also provide a bit better isolation to avoid free floating functions

src/transport/manager/mod.rs
```diff
@@ -73,6 +73,12 @@ pub(crate) mod handle;
 /// Logging target for the file.
 const LOG_TARGET: &str = "litep2p::transport-manager";
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum DialFailureAddresses {
```
lexnv, line 82: nit: We could add some documentation explaining the purpose of this enum

src/transport/manager/mod.rs
```diff
@@ -106,6 +112,8 @@ pub struct ProtocolContext {
 
     /// Fallback names for the protocol.
     pub fallback_names: Vec<ProtocolName>,
+
+    pub dial_failure_mode: DialFailureAddresses,
```
lexnv, line 124: nit: Missing docs

src/transport/manager/mod.rs
```diff
@@ -1116,10 +1134,10 @@ impl TransportManager {
...
                                                     "dial failure, notify protocol",
                                                 );
-                                                match context.tx.try_send(InnerTransportEvent::DialFailure {
-                                                    peer,
-                                                    addresses: vec![address.clone()],
-                                                }) {
+
+                                                let adresses = context.dial_failure_addresses(&[address.clone()]);
+
+                                                match context.tx.try_send(make_dial_failure_event(peer, adresses.clone())) {
```
lexnv: nit: Is there a way to avoid the cloning here? Maybe changing the signature of the fn to take a `& Multiaddr`?

I would also inline the `make_dial_failure_event`:

```rust
match context.tx.try_send(InnerTransportEvent::DialFailure { peer, addresses })
```

Instead of cloning the addreses here (`addresses.clone()`), we could recreate the addresses on error

dharjeezy: i have changed to this inline way, but can't still do without cloning

lexnv (approved):
Looks good to me! Thanks again for contributing 🙏

src/transport/manager/mod.rs
```diff
@@ -1132,10 +1158,7 @@ impl TransportManager {
                                                         );
                                                         let _ = context
                                                             .tx
-                                                            .send(InnerTransportEvent::DialFailure {
-                                                                peer,
-                                                                addresses: vec![address.clone()],
-                                                            })
+                                                            .send(InnerTransportEvent::DialFailure { peer, addresses })
```
dmitry-markin, line 1231: @dharjeezy Can you also add to the doc of `InnerTransportEvent` & `TransportEvent` that `addresses` is only forwarded if the protocol was registered with `DialFailureAddresses::Required`, please?

dharjeezy: this is done now @dmitry-markin. can we merge it in now?

src/transport/mod.rs
```diff
@@ -154,6 +154,7 @@ pub(crate) enum TransportEvent {
     },
 
     /// Failed to dial remote peer.
+    ///  The addresses are only forwarded to the protocol if it was registered with `DialFailureAddresses::Required`.
```
lexnv:
```suggestion
    /// The addresses are only forwarded to the protocol if it was registered with `DialFailureAddresses::Required`.
```

PR #445, opened 2025-10-15
yamux/control: Ensure poll next inbound is not called after errors

lexnv:
We have recently fixed a spurious issue that caused the rust-yamux component to panic.

Litep2p transitioned the internal state of the connection back to Idle.
This causes litep2p to poll the component again. Then rust-yamux would use the same buffer offset that caused the panic in the first place.

Instead, this PR ensures that litep2p never polls the connection once yamux encounters an error.
This is aligned with the behavior of transport layers that close the connection on the first yamux error.

Part of: https://github.com/paritytech/polkadot-sdk/issues/9169

cc @paritytech/networking

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,26 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.0] - 2025-10-20
+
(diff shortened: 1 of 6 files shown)


src/yamux/control.rs
```diff
@@ -81,7 +83,25 @@ where
...
+                            // substream error. The `connection.poll_next_inbound` should
+                            // not be called again after returning an error. Instead, we
+                            // must close the connection gracefully.
+                            match maybe_stream.as_ref() {
+                                Some(Err(error)) => {
+                                    tracing::debug!(target: LOG_TARGET, ?error, "Inbound stream error, closing connection");
+
+                                    self.state = State::Closing {
+                                        reply: None,
+                                        inner: Closing::ClosingConnection { connection },
```
dmitry-markin: I believe we first should set the state to `Closing::DrainingControlCommands` to respond to all outstanding control commands.

dmitry-markin (approved):

PR #446, opened 2025-10-15
Add x-shadow feature for Shadow simulation compatibility

xDimon:
# Summary

This PR introduces the `x-shadow` feature, which allows building **Litep2p** compatible with the **Shadow** network simulator.
When this feature is enabled, outgoing socket binding is disabled to prevent address collisions that occur under Shadow, which otherwise lead to `AddrInUse` errors when connecting to remote peers.

# Changes

- Added a new Cargo feature: `x-shadow`.
- Disabled explicit binding of outgoing sockets when the feature is active.
- djusted connection logic to comply with **Shadow** environment constraints.
- Updated build configuration and conditional compilation accordingly.

# Motivation

Under the **Shadow** simulator, each node runs in an isolated process namespace but shares the same IP address space.
Explicitly binding outgoing sockets may cause multiple nodes to use the same source address, resulting in `EADDRINUSE` collisions.
The `x-shadow` feature resolves this by skipping the bind step for outgoing connections, ensuring stable network behavior in simulation.

# Compatibility

- Activated via Cargo feature flag: `--features x-shadow`.
- Has no effect on normal (non-Shadow) builds or runtime behavior.
- Safe to merge — default build remains unchanged.

# Testing

1. Build with `--features x-shadow`.
2. Run the binary under Shadow simulation.
3. Verify that connections between peers establish successfully.
4. Ensure no AddrInUse errors occur during parallel outgoing connections.
5. Confirm clean shutdown and resource release after simulation.

### DIFF
--- a/src/transport/tcp/mod.rs
+++ b/src/transport/tcp/mod.rs
@@ -42,6 +42,7 @@ use futures::{
     stream::{AbortHandle, FuturesUnordered, Stream, StreamExt},
     TryFutureExt,
 };
+use hickory_resolver::TokioResolver;
 use multiaddr::Multiaddr;
 use socket2::{Domain, Socket, Type};
 use tokio::net::TcpStream;
@@ -50,6 +51,7 @@ use std::{
     collections::HashMap,
     net::SocketAddr,
     pin::Pin,
+    sync::Arc,
     task::{Context, Poll},
     time::Duration,
 };
@@ -129,6 +131,9 @@ pub(crate) struct TcpTransport {
     /// Connections which have been opened and negotiated but are being validated by the
     /// `TransportManager`.
     pending_open: HashMap<ConnectionId, NegotiatedConnection>,
+
+    /// DNS resolver.
+    resolver: Arc<TokioResolver>,
 }
 
 impl TcpTransport {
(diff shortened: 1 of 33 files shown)


src/transport/tcp/mod.rs
```diff
@@ -213,6 +213,7 @@ impl TcpTransport {
         socket.set_nonblocking(true)?;
         socket.set_nodelay(nodelay)?;
 
+        #[cfg(not(feature = "x-shadow"))]
```
dmitry-markin, line 216: There is an existing runtime configuration for this in `tcp::Config` (and similarly for websocket): https://github.com/paritytech/litep2p/blob/8a02398708959339fb60b18a9cd3461eb152408c/src/transport/tcp/config.rs#L42

xDimon: This configuration only applies to listening sockets, and it is ok. But problem happens when we are trying to bind address for outgoing sockets.

dmitry-markin: It applies both to listen addresses and [dial addresses](https://github.com/paritytech/litep2p/blob/8a02398708959339fb60b18a9cd3461eb152408c/src/transport/common/listener.rs#L319-L325).

Issue #444, opened 2025-10-14
bitswap: reuse inbound substream for subsequent requests


dmitry-markin:
Currently, we drop an inbound substream after receiving a Bitswap request. In the past, kubo IPFS implementation was smart enough to open new substreams for subsequent requests. kubo >= 0.37.0 is more picky and refuses to send subsequent requests over new substreams. This is caused by https://github.com/ipfs/boxo/pull/984 released in boxo v0.33.1 (Bitswap implementation used by kubo) and pulled by kubo v0.37.0.

This leads to first Bitswap "have" request to be received and responded, but a subsequent "block" request to never arrive. As a consequence, no data is served over Bitswap.

In order to retain compatibility with kubo and other IPFS implementations we should keep the inbound substream open and handle any subsequent messages sent over it.

PR #447, opened 2025-10-16
bitswap: Reuse inbound substream for subsequent requests

dmitry-markin:
Fix compatibility with kubo IPFS >= 0.37.0.

Close https://github.com/paritytech/litep2p/issues/444.

### DIFF
--- a/src/protocol/libp2p/bitswap/mod.rs
+++ b/src/protocol/libp2p/bitswap/mod.rs
@@ -130,7 +130,7 @@ impl Bitswap {
     fn on_inbound_substream(&mut self, peer: PeerId, substream: Substream) {
         tracing::debug!(target: LOG_TARGET, ?peer, "handle inbound substream");
 
-        if let Some(_) = self.inbound.insert(peer, substream) {
+        if self.inbound.insert(peer, substream).is_some() {
             // Only one inbound substream per peer is allowed in order to constrain resources.
             tracing::debug!(
                 target: LOG_TARGET,


src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -123,45 +122,61 @@ impl Bitswap {
...
-
-                        (cid.version() == cid::Version::V1
-                            && cid.hash().code() == u64::from(Code::Blake2b256)
-                            && cid.hash().size() == 32)
-                            .then_some((cid, want_type))
-                    })
-                    .collect::<Vec<_>>(),
-            ))
-        }));
+        if let Some(_) = self.inbound.insert(peer, substream) {
```
lexnv: nit: `self.inbound.insert(peer, substream).is_some()`

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -123,45 +122,61 @@ impl Bitswap {
...
+            // Only one inbound substream per peer is allowed in order to constrain resources.
+            tracing::debug!(
+                target: LOG_TARGET,
+                ?peer,
+                "dropping inbound substream as remote opened a new one",
```
lexnv, line 138: feel free to ignore completely: Maybe we can add the `id = %substream.substream_id()` to these logs to group them a bit better 🤔

dmitry-markin: Unfortunately, `substream_id()` is not exposed on a substream, and I wouldn't introduce one just for logs...

Issue #347, opened 2025-02-27
tcp: Merge `open_connection` into the `negotiate_connection` fn


lexnv:
https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/transport/tcp/connection.rs#L212-L231

PR #459, opened 2025-10-23
refactor: merge open_connection into negotiate_connection

iTranscend:
This PR merges `open_connection()` into `negotiate_connection()`.

resolves https://github.com/paritytech/litep2p/issues/347


dmitry-markin (commented):
I don't know what @lexnv had in mind when created the issue, but it seems a free-standing function was replaced with an async block. IMO this reduces readability and I would look into other options to simplify the code.

Issue #342, opened 2025-02-27
substream: Remove the SubstreamSet structure


lexnv:
At the moment, the SubstreamSet is used for testing purposes and in one of our examples.

We have moved away from the SubstreamSet in favor of FuturesUnordered because the SubstreamSet is not fairly polling streams:
- https://github.com/paritytech/litep2p/pull/321

Adjust the example and remove this from our testing

PR #462, opened 2025-10-29
feat: transition from SubstreamSet to FuturesStream

iTranscend:
in partial resolution of https://github.com/paritytech/litep2p/issues/342

This PR replaces the use of the SubstreamSet hashmap in substream's test with the FuturesStream type.


src/substream/mod.rs
```diff
@@ -849,58 +849,33 @@ where
...
-    }
-
-    #[test]
-    fn remove_substream() {
-        let mut set = SubstreamSet::<PeerId, MockSubstream>::new();
-
-        let peer1 = PeerId::random();
-        let substream1 = MockSubstream::new();
-        set.insert(peer1, substream1);
+        let task = async |mut substream: MockSubstream| {
```
lexnv: nit: Maybe we can introduce a helper function for testing since this code is reused

lexnv (approved):
Thanks @iTranscend for contributing 🙏 

This looks good to me, it is nice that we have tests for our `FuturesStream` implementation

src/substream/mod.rs
```diff
@@ -849,58 +849,36 @@ where
...
+            Some(Ok(request)) => Ok(request),
+            Some(Err(error)) => Err(error),
+            None => Err(SubstreamError::ConnectionClosed),
+        };
+
+        (request, substream)
+    }
 
     #[test]
     fn add_substream() {
```
dmitry-markin: IMO this test doesn't make sense anymore or should go to `FuturesStream` tests.

dmitry-markin (changes requested):
These tests were testing the `SubstreamSet`, implemented in this file. As it was replaced with `FuturesStream`, the equivalent tests should go to `utils/futures_stream.rs`.

Also, the `SubstreamSet` should be completely removed.

src/substream/mod.rs
lexnv: Yep we could backport them to futures streams, makes sense 🙏

iTranscend: great catch. I'll relocate the tests asap

lexnv:
> @lexnv What was the motivation behind the removal of SubstreamSet in the first place?

The core logic of litep2p has moved away from `SubstreamSet` toward `FuturesStream`.
However, we do have a testing component `CustomProtocol` which still uses that. 

Offhand, I believe we can adjust the logic to achieve a similar output:


We'd need to keep polling the futuresStream as before:
```
event = self.substreams.next()
```

iTranscend:
Hey @lexnv & @dmitry-markin, sorry for the radio silence,

I finally got around to resolving this. I went with lexnv's suggestion of maintaining a mapping of peerIDs to channels for forwarding commands to Substream futures within the FuturesStream.

Please take a look when you're chanced.

PR #465, opened 2025-11-05
fix: multistream-select negotiation on outbound substream over webrtc

### LINKED ISSUE #464: webrtc: multistream-select negotiation not fully spec compliant
I'm testing interoperability between a smoldot dialer running in wasm/Chrome and a litep2p listener. Right now this is limited to connection establishment, noise handshake and ping behaviour in either direction. The only part that doesn't work is multistream-select negotiation on the outbound substream from litep2p to smoldot.

Due to the libp2p-webrtc specific protobuf framing and how network data is surfaced by str0m, there is [a custom code path](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/transport/webrtc/connection.rs#L374) for handling multistream-select negotiation on webrtc connections. That code path does not use [`multistream_select::protocol::MessageIO`](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/protocol.rs#L260) by going through [`listener_select_proto()`](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/listener_select.rs#L57) and therefore does not handle length prefixes in multistream-select messages correctly.

When smoldot is made aware of a newly opened inbound substream by calling [`MultiStream::add_substream()`](https://github.com/smol-dot/smoldot/blob/33d34fb8733c7d8a7e5878f8658bc7a0d0383bae/lib/src/libp2p/connection/established/multi_stream.rs#L192), it [sends the multistream-select header right away](https://github.com/smol-dot/smoldot/blob/33d34fb8733c7d8a7e5878f8658bc7a0d0383bae/lib/src/libp2p/connection/multistream_select.rs#L209), without waiting for any data to arrive from the dialer. The resulting message only contains the header, [with a length prefix](https://github.com/smol-dot/smoldot/blob/33d34fb8733c7d8a7e5878f8658bc7a0d0383bae/lib/src/libp2p/connection/multistream_select.rs#L398).

Once the protocol request message has been received from the dialer and the API user has accepted the protocol, [the confirmation message is sent](https://github.com/smol-dot/smoldot/blob/33d34fb8733c7d8a7e5878f8658bc7a0d0383bae/lib/src/libp2p/connection/multistream_select.rs#L133).

The `Vec<u8>` [received from str0m in the `ChannelData` event](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/transport/webrtc/connection.rs#L732) still contains the length prefix, which causes the relevant comparisons in `multistream_select::protocol::Message::decode()` to fail [for the header](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/protocol.rs#L176) and the [protocol confirmation](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/protocol.rs#L190).

When litep2p is the listener, the negotiation succeeds, because <strike>litep2p sends both header and protocol confirmation in one message, which `multistream_select::protocol::Message::encode()` treats as response to an `ls` command and [adds length prefixes](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/protocol.rs#L153)</strike> [`webrtc_encode_multistream_message()`](https://github.com/paritytech/litep2p/blob/1e3fa87bc2dd143f8a4390582a4539e10a17eae0/src/multistream_select/protocol.rs#L235) adds length prefixes.

haikoschol:
This PR changes how multistream-select messages received on an opening outbound substream are handled on webrtc connections. For details please see the issue description.

The PR builds on top of #441 for the str0m update. But looking at it again I realized that it also made changes to multistream-select handling. I'll look into what those are and how they relate to my changes. In any case, I believe my changes bring the implementation closer to compliance with the multistream-select spec and enable interoperability with smoldot.

fixes #464

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -1447,67 +1447,70 @@ mod tests {
         (dial_address, connection_id)
     }
 
-    #[cfg(feature = "websocket")]
-    struct MockTransport {
-        rx: tokio::sync::mpsc::Receiver<TransportEvent>,
-    }
-
-    #[cfg(feature = "websocket")]
-    impl MockTransport {
-        fn new(rx: tokio::sync::mpsc::Receiver<TransportEvent>) -> Self {
-            Self { rx }
-        }
-    }
-
-    #[cfg(feature = "websocket")]
-    impl Transport for MockTransport {
-        fn dial(&mut self, _connection_id: ConnectionId, _address: Multiaddr) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn accept(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn accept_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn reject_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn reject(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn open(
-            &mut self,
-            _connection_id: ConnectionId,
-            _addresses: Vec<Multiaddr>,
-        ) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn negotiate(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
-            Ok(())
-        }
-
-        fn cancel(&mut self, _connection_id: ConnectionId) {}
-    }
-
-    #[cfg(feature = "websocket")]
-    impl Stream for MockTransport {
-        type Item = TransportEvent;
-        fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-            self.rx.poll_recv(cx)
-        }
-    }
-
     #[tokio::test]
     #[cfg(feature = "websocket")]
     #[cfg(feature = "quic")]
     async fn transport_events() {
+        struct MockTransport {
+            rx: tokio::sync::mpsc::Receiver<TransportEvent>,
+        }
+
+        impl MockTransport {
+            fn new(rx: tokio::sync::mpsc::Receiver<TransportEvent>) -> Self {
+                Self { rx }
+            }
+        }
+
+        impl Transport for MockTransport {
+            fn dial(
+                &mut self,
+                _connection_id: ConnectionId,
+                _address: Multiaddr,
+            ) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn accept(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn accept_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn reject_pending(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn reject(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn open(
+                &mut self,
+                _connection_id: ConnectionId,
+                _addresses: Vec<Multiaddr>,
+            ) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn negotiate(&mut self, _connection_id: ConnectionId) -> crate::Result<()> {
+                Ok(())
+            }
+
+            fn cancel(&mut self, _connection_id: ConnectionId) {}
+        }
+
+        impl Stream for MockTransport {
+            type Item = TransportEvent;
+            fn poll_next(
+                mut self: Pin<&mut Self>,
+                cx: &mut Context<'_>,
+            ) -> Poll<Option<Self::Item>> {
+                self.rx.poll_recv(cx)
+            }
+        }
+
         let mut transports = TransportContext::new();
 
         let (tx_tcp, rx) = tokio::sync::mpsc::channel(8);
--- a/src/transport/tcp/connection.rs
+++ b/src/transport/tcp/connection.rs
@@ -123,6 +123,15 @@ pub struct NegotiatedConnection {
     substream_open_timeout: Duration,
 }
 
+impl std::fmt::Debug for NegotiatedConnection {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("NegotiatedConnection")
+            .field("peer", &self.peer)
+            .field("endpoint", &self.endpoint)
+            .finish()
+    }
+}
+
 impl NegotiatedConnection {
     /// Get `ConnectionId` of the negotiated connection.
     pub fn connection_id(&self) -> ConnectionId {
--- a/src/transport/tcp/mod.rs
+++ b/src/transport/tcp/mod.rs
@@ -22,7 +22,6 @@
 //! TCP transport.
 
 use crate::{
-    config::Role,
     error::{DialError, Error},
     transport::{
         common::listener::{DialAddresses, GetSocketAddr, SocketListener, TcpAddress},
@@ -78,9 +77,8 @@ struct PendingInboundConnection {
 enum RawConnectionResult {
     /// The first successful connection.
     Connected {
-        connection_id: ConnectionId,
-        address: Multiaddr,
-        stream: TcpStream,
+        negotiated: NegotiatedConnection,
+        errors: Vec<(Multiaddr, DialError)>,
     },
 
     /// All connection attempts failed.
@@ -121,7 +119,7 @@ pub(crate) struct TcpTransport {
     pending_raw_connections: FuturesStream<BoxFuture<'static, RawConnectionResult>>,
 
     /// Opened raw connection, waiting for approval/rejection from `TransportManager`.
-    opened_raw: HashMap<ConnectionId, (TcpStream, Multiaddr)>,
+    opened: HashMap<ConnectionId, NegotiatedConnection>,
 
     /// Cancel raw connections futures.
     ///
@@ -300,7 +298,7 @@ impl TransportBuilder for TcpTransport {
                 config,
                 context,
                 dial_addresses,
-                opened_raw: HashMap::new(),
+                opened: HashMap::new(),
                 pending_open: HashMap::new(),
                 pending_dials: HashMap::new(),
                 pending_inbound_connections: HashMap::new(),
@@ -360,6 +358,37 @@ impl Transport for TcpTransport {
         Ok(())
     }
 
+    fn accept_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
+        let pending = self.pending_inbound_connections.remove(&connection_id).ok_or_else(|| {
+            tracing::error!(
+                target: LOG_TARGET,
+                ?connection_id,
+                "Cannot accept non existent pending connection",
+            );
+
+            Error::ConnectionDoesntExist(connection_id)
+        })?;
+
+        self.on_inbound_connection(connection_id, pending.connection, pending.address);
+
+        Ok(())
+    }
+
+    fn reject_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
+        self.pending_inbound_connections.remove(&connection_id).map_or_else(
+            || {
+                tracing::error!(
+                    target: LOG_TARGET,
+                    ?connection_id,
+                    "Cannot reject non existent pending connection",
+                );
+
+                Err(Error::ConnectionDoesntExist(connection_id))
+            },
+            |_| Ok(()),
+        )
+    }
+
     fn accept(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
         let context = self
             .pending_open
@@ -393,37 +422,6 @@ impl Transport for TcpTransport {
         Ok(())
     }
 
-    fn accept_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
-        let pending = self.pending_inbound_connections.remove(&connection_id).ok_or_else(|| {
-            tracing::error!(
-                target: LOG_TARGET,
-                ?connection_id,
-                "Cannot accept non existent pending connection",
-            );
-
-            Error::ConnectionDoesntExist(connection_id)
-        })?;
-
-        self.on_inbound_connection(connection_id, pending.connection, pending.address);
-
-        Ok(())
-    }
-
-    fn reject_pending(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
-        self.pending_inbound_connections.remove(&connection_id).map_or_else(
-            || {
-                tracing::error!(
-                    target: LOG_TARGET,
-                    ?connection_id,
-                    "Cannot reject non existent pending connection",
-                );
-
-                Err(Error::ConnectionDoesntExist(connection_id))
-            },
-            |_| Ok(()),
-        )
-    }
-
     fn reject(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
         self.pending_open
             .remove(&connection_id)
@@ -439,13 +437,18 @@ impl Transport for TcpTransport {
         let mut futures: FuturesUnordered<_> = addresses
             .into_iter()
             .map(|address| {
-                let dial_addresses = self.dial_addresses.clone();
+                let yamux_config = self.config.yamux_config.clone();
+                let max_read_ahead_factor = self.config.noise_read_ahead_frame_count;
+                let max_write_buffer_size = self.config.noise_write_buffer_size;
                 let connection_open_timeout = self.config.connection_open_timeout;
+                let substream_open_timeout = self.config.substream_open_timeout;
+                let dial_addresses = self.dial_addresses.clone();
+                let keypair = self.context.keypair.clone();
                 let nodelay = self.config.nodelay;
                 let resolver = self.resolver.clone();
 
                 async move {
-                    TcpTransport::dial_peer(
+                    let (address, stream) = TcpTransport::dial_peer(
                         address.clone(),
                         dial_addresses,
                         connection_open_timeout,
@@ -453,7 +456,26 @@ impl Transport for TcpTransport {
                         resolver,
                     )
                     .await
-                    .map_err(|error| (address, error))
+                    .map_err(|error| (address, error))?;
+
+                    let open_address = address.clone();
+                    let (socket_address, peer) = TcpAddress::multiaddr_to_socket_address(&address)
+                        .map_err(|error| (address, error.into()))?;
+
+                    TcpConnection::open_connection(
+                        connection_id,
+                        keypair,
+                        stream,
+                        socket_address,
+                        peer,
+                        yamux_config,
+                        max_read_ahead_factor,
+                        max_write_buffer_size,
+                        connection_open_timeout,
+                        substream_open_timeout,
+                    )
+                    .await
+                    .map_err(|error| (open_address, error.into()))
                 }
             })
             .collect();
@@ -463,12 +485,7 @@ impl Transport for TcpTransport {
             let mut errors = Vec::with_capacity(num_addresses);
             while let Some(result) = futures.next().await {
                 match result {
-                    Ok((address, stream)) =>
-                        return RawConnectionResult::Connected {
-                            connection_id,
-                            address,
-                            stream,
-                        },
+                    Ok(negotiated) => return RawConnectionResult::Connected { negotiated, errors },
                     Err(error) => {
                         tracing::debug!(
                             target: LOG_TARGET,
@@ -496,52 +513,12 @@ impl Transport for TcpTransport {
     }
 
     fn negotiate(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
-        let (stream, address) = self
-            .opened_raw
+        let negotiated = self
+            .opened
             .remove(&connection_id)
             .ok_or(Error::ConnectionDoesntExist(connection_id))?;
 
-        let (socket_address, peer) = TcpAddress::multiaddr_to_socket_address(&address)?;
-        let yamux_config = self.config.yamux_config.clone();
-        let max_read_ahead_factor = self.config.noise_read_ahead_frame_count;
-        let max_write_buffer_size = self.config.noise_write_buffer_size;
-        let connection_open_timeout = self.config.connection_open_timeout;
-        let substream_open_timeout = self.config.substream_open_timeout;
-        let keypair = self.context.keypair.clone();
-
-        tracing::trace!(
-            target: LOG_TARGET,
-            ?peer,
-            ?connection_id,
-            ?address,
-            "negotiate connection",
-        );
-
-        self.pending_dials.insert(connection_id, address);
-        self.pending_connections.push(Box::pin(async move {
-            match tokio::time::timeout(connection_open_timeout, async move {
-                TcpConnection::negotiate_connection(
-                    stream,
-                    peer,
-                    connection_id,
-                    keypair,
-                    Role::Dialer,
-                    socket_address,
-                    yamux_config,
-                    max_read_ahead_factor,
-                    max_write_buffer_size,
-                    substream_open_timeout,
-                )
-                .await
-                .map_err(|error| (connection_id, error.into()))
-            })
-            .await
-            {
-                Err(_) => Err((connection_id, DialError::Timeout)),
-                Ok(Err(error)) => Err(error),
-                Ok(Ok(connection)) => Ok(connection),
-            }
-        }));
+        self.pending_connections.push(Box::pin(async move { Ok(negotiated) }));
 
         Ok(())
     }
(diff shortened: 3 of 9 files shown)


src/transport/webrtc/connection.rs
```diff
@@ -631,6 +642,8 @@ impl WebRtcConnection {
             protocol: protocol.to_string(),
         });
 
+        // self.rtc.channel(channel_id).unwrap().set_buffered_amount_low_threshold(1024);
```
haikoschol: This is [an unrelated issue](https://github.com/paritytech/litep2p/pull/441#discussion_r2464955120).

src/transport/webrtc/connection.rs
```diff
@@ -787,17 +814,15 @@ impl WebRtcConnection {
                 },
                 event = self.handles.next() => match event {
                     None => unreachable!(),
-                    Some((channel_id, None | Some(SubstreamEvent::Close))) => {
+                    Some((_, None)) => {}
```
haikoschol: This is a superficial fix for another unrelated issue which I haven't investigated yet. Without this change, a "channel closed" event is fired repeatedly after a while. I believe this is caused by `SubstreamHandle::tx` being dropped, presumably when the substream(/channel?) is closed. The fact that the event continues firing suggests that `SubstreamHandle::rx` continues to be polled. Maybe the handle simply needs to be removed from `self.handles`?

I've also seen this happen when briefly testing with a rust-libp2p dialer, so this is not related to smoldot interop.

src/transport/webrtc/connection.rs
```diff
@@ -787,17 +814,15 @@ impl WebRtcConnection {
...
                         tracing::trace!(
                             target: LOG_TARGET,
                             peer = ?self.peer,
                             ?channel_id,
                             "channel closed",
                         );
-
-                        self.rtc.direct_api().close_data_channel(channel_id);
                         self.channels.insert(channel_id, ChannelState::Closing);
-                        self.handles.remove(&channel_id);
```
lexnv, line 800: Wouldn't this lead to memory leaks due to keeping the channel ids around? 🤔

haikoschol: You are right. I don't remember what exactly I had in mind with that change. Probably expecting that `Event::ChannelClose` would be emitted eventually and `Connection::on_channel_closed()` called, which does the full cleanup. But that's not exactly how it works.

The way it was before is incorrect as well, but that's out of scope for this PR since it is related with signaling of substream closure via the WebRTC protobuf frame.
Anyway, I'll revert this change and we'll revisit it in a follow-up PR.

src/transport/webrtc/connection.rs
haikoschol: I've removed this line since it is unrelated to this PR and should be address later when working on [backpressure](https://github.com/paritytech/litep2p/issues/45).

src/transport/webrtc/connection.rs
haikoschol: I just tested again without this change and now I don't get repeating "channel closed" messages in the logs. Presumably this is due to a change somewhere else that I can't pin down right now. In any case, it was unrelated to multistream-select negotiation.

src/multistream_select/dialer_select.rs
```diff
@@ -41,6 +41,7 @@ use std::{
     pin::Pin,
     task::{Context, Poll},
 };
+use crate::error::SubstreamError;
```
lexnv: nit: We could group it with the previous crate import

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
...
-                NegotiationError::Failed,
-            ));
+        let header = Protocol::try_from(&b"/multistream/1.0.0"[..])
+            .expect("valid multitstream-select header");
+
+        // All multistream-select messages are length-prefixed. Since this code path is not using
+        // multistream_select::protocol::MessageIO, we need to decode and remove the length here.
+        let mut remaining: &[u8] = &payload;
+        let (len, tail) = unsigned_varint::decode::usize(remaining).
+            map_err(|_| error::NegotiationError::ParseError(ParseError::InvalidData))?;
```
lexnv: nit: Let's log a degub about about this error, decoding usize should pass in most of the cases

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
...
+
+        let payload = tail[..len].to_vec();
+
```
lexnv: nit: It would be good if we can greak the code such that we provide better trace logging:

```rust
let message = Message::decode(payload.into());

tracing::trace!(
    target: LOG_TARGET,
    ?message,
    "Decoded message while registering response",
);

let mut protocols = match message {
    ..
};

```

Let me know if this makes sense or if you think it will polute too much the logs

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
...
 
+        remaining = &tail[len..];
+
+        loop {
+            if remaining.is_empty() {
+                break;
+            }
+
+            let (len, tail) = unsigned_varint::decode::usize(remaining).
+                map_err(|_| error::NegotiationError::ParseError(ParseError::InvalidData))?;
```
lexnv: Does this mean we are expecting multiple `[ len_prefix ++ Message ]` to arrive from the `register_response` of the WebRTC?

Who would normally exhibit this behavior (libp2p or smoldot)?

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
...
+
+            if len > tail.len() {
+                break;
+            }
+
+            let payload = tail[..len].to_vec();
+
+            match Message::decode(payload.into()) {
```
lexnv: nit: Lets log in a smilar way the message here, the `_` is too loose and it might help us debug connectivity issues

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
...
+                    _ => Err(error::NegotiationError::StateMismatch),
+                }
+            }
+            Ok(Message::ListProtocols) => return Err(error::NegotiationError::StateMismatch),
+            Err(_) => return Err(error::NegotiationError::ParseError(ParseError::InvalidData)),
         };
 
+        remaining = &tail[len..];
+
+        loop {
```
lexnv: nit: We can move this to a dedicated fn `drain_trailing_protocols`. Then, we can add tests for this behavior to enure we are not missing other frames and we receive expected errors.

Is this spec compliant? It seems like some hoops we have to jump through to make either libp2p/smoldot happy

src/multistream_select/dialer_select.rs
```diff
@@ -357,23 +358,67 @@ impl WebRtcDialerState {
         &mut self,
         payload: Vec<u8>,
     ) -> Result<HandshakeResult, crate::error::NegotiationError> {
-        let Message::Protocols(protocols) =
-            Message::decode(payload.into()).map_err(|_| ParseError::InvalidData)?
-        else {
-            return Err(crate::error::NegotiationError::MultistreamSelectError(
-                NegotiationError::Failed,
-            ));
+        let header = Protocol::try_from(&b"/multistream/1.0.0"[..])
```
lexnv: nit: Could add a const about this similar to `MSG_MULTISTREAM_1_0`

src/multistream_select/dialer_select.rs
```diff
@@ -755,23 +802,18 @@ mod tests {
...
-        };
+        let mut bytes = BytesMut::with_capacity(32);
+        bytes.put_u8(MSG_MULTISTREAM_1_0.len() as u8);
+        let _ = Message::Header(HeaderLine::V1).encode(&mut bytes).unwrap();
+
+        let proto = Protocol::try_from(&b"/13371338/proto/1"[..]).expect("valid protocol name");
+        bytes.put_u8((proto.as_ref().len() + 1) as u8); // + 1 for \n
+        let _ = Message::Protocol(proto).encode(&mut bytes).unwrap();
+
+        let expected_message = bytes.freeze().to_vec();
```
lexnv, line 806: dq: This slightly changes the meaning of the messages.

Before we were expecting: 

```
[Protocols(/multistream/1.0.0), Protocol(/13371338/proto/1)]
```

Now we are expecting:

```
// \n added
[Header(/multistream/1.0.0 /\n),

// \n counted but not checked
Protocol(/13371338/proto/1 [])
```

Hmm, one of those representation can't be spec compliant right? 

Also are we ignoring malformated messages:
- we decode the frame size correctly
- but we don't check then if the last character is `\n` or not sufficeint bytes provided?

lexnv (commented):
Great job so far! Feels like we are getting closer to figuring out the multistream select 🙏 

My questions revolve around the spec correctness of the implementation, or if we basically implemented a tiny subset in the past? And mainly, which implementation is causing us to extensively parse the frames (libp2p or smoldot or both?) 🤔

src/multistream_select/dialer_select.rs
haikoschol: I think this is fine, especially as a trace.

src/multistream_select/dialer_select.rs
haikoschol: I've extracted the loop to a function, but I don't know what you mean with "not missing other frames". The code is covered by some of the existing tests (`negotiate_main_protocol()` and `negotiate_fallback_protocol()` IIRC).

The reason for adding this loop is that I removed a line in `webrtc_encode_multistream_message()` that [adds a newline at the end](https://github.com/paritytech/litep2p/pull/465/files#diff-d546bf98df892f7b9165971703a8e8c521913a7c5c4c8ae9c1c501949ddfe263L253). What isn't spec compliant is adding this newline, because that is only present in an "ls response" (even with the newline it is not a valid "ls response" because it lacks a length-prefix for the total message size at the start). It worked before because smoldot and rust-libp2p tolerate this extra newline.

lexnv: Got it! Makes perfect sense! Thanks @haikoschol again for handling this 🙏

src/multistream_select/dialer_select.rs
```diff
@@ -358,20 +357,32 @@ impl WebRtcDialerState {
         &mut self,
         payload: Vec<u8>,
     ) -> Result<HandshakeResult, crate::error::NegotiationError> {
-        let header = Protocol::try_from(&b"/multistream/1.0.0"[..])
```
lexnv: Nit: Before we were constructing the protocol from `b"/multistream/1.0.0"`. Now we are using `pub const MSG_MULTISTREAM_1_0: &[u8] = b"/multistream/1.0.0\n";` which adds the extra `\n`.

This should be fine as long as we don't add an extra `\n` at the end: https://github.com/paritytech/litep2p/pull/465/files#diff-d546bf98df892f7b9165971703a8e8c521913a7c5c4c8ae9c1c501949ddfe263L253?

haikoschol: Yes, the line that adds the `\n` is problematic for reasons I've tried to explain [here](https://github.com/paritytech/litep2p/pull/465#discussion_r2523177188).

The result we want from `webrtc_encode_multistream_message()` is a single WebRTC protobuf frame that contains one or more multistream messages.

`Message::encode()` adds a newline when encoding `Message::Protocol`, so if we add another one in `webrtc_encode_multistream_message()`, the result is not spec-compliant.

src/multistream_select/dialer_select.rs
haikoschol: Right now the test ensures that the wire format that `WebRtcDialerState::propose()` produces is what we expect. Before it actually ensured that whatever `propose()` produces can be decoded with `Message::decode()`. So I did change the semantics of the test after all. But I would argue that the new meaning is more useful. We could also add decoding and assert we get the expected strings back, but due to the changes made, this can no longer be done by just calling `Message::decode()`.

src/multistream_select/dialer_select.rs
```diff
@@ -405,12 +441,62 @@ impl WebRtcDialerState {
...
+                tracing::debug!(
+                    target: LOG_TARGET,
+                    ?error,
+                    message = ?remaining,
+                    "Failed to decode length-prefix in multistream message");
+                error::NegotiationError::ParseError(ParseError::InvalidData)
+            })?;
+
+        if len > tail.len() {
+            break;
```
lexnv: nit: I think this would miss a few protocols, maybe return an error and add a debug log?

haikoschol: I just tried this and it caused the test [`header_line_missing()`](https://github.com/ChainSafe/litep2p/blob/9621e3ccda915882f0dc93b7c2ea46c80c0fea7a/src/multistream_select/dialer_select.rs#L902) to fail. This is because that test used `Message::Protocols` to encode two protocol names (without a header) to use as a negotiation response. This isn't spec compliant in two ways that are irrelevant to what is being tested, so I changed the test to send one protocol name.

src/multistream_select/dialer_select.rs
```diff
@@ -357,24 +357,57 @@ impl WebRtcDialerState {
...
+                    target: LOG_TARGET,
+                    ?error,
+                    message = ?payload,
+                    "Failed to decode length-prefix in multistream message");
+            error::NegotiationError::ParseError(ParseError::InvalidData)
+        })?;
+
+        let payload = tail[..len].to_vec();
+
+        let message = Message::decode(payload.into());
```
dmitry-markin: nit: to not allocate & copy the data we can wrap the original `Vec<u8>` data into `Bytes` directly and then slice this new `Bytes` object into other `Bytes` chunks to pass to `Message::decode` here and later in `drain_trailing_protocols`:

```suggestion
        let len_size = remaining.len() - tail.len();
        let bytes = Bytes::from(payload);
        let payload = bytes.slice(len_size..len_size + len);
        let remaining = bytes.slice(len_size + len..);
        
        let message = Message::decode(payload);
```

src/multistream_select/dialer_select.rs
```diff
@@ -405,10 +438,66 @@ impl WebRtcDialerState {
     }
 }
 
+fn drain_trailing_protocols(
+    tail: &[u8],
+    len: usize,
+) -> Result<Vec<Protocol>, error::NegotiationError> {
+    let mut protocols = vec![];
+    let mut remaining = &tail[len..];
```
dmitry-markin:
```suggestion
fn drain_trailing_protocols(
    mut remaining: Bytes,
) -> Result<Vec<Protocol>, error::NegotiationError> {
    let mut protocols = vec![];
```

src/multistream_select/dialer_select.rs
```diff
@@ -357,24 +357,57 @@ impl WebRtcDialerState {
...
+                    HandshakeState::WaitingProtocol => Err(
+                        error::NegotiationError::MultistreamSelectError(NegotiationError::Failed),
+                    ),
+                    _ => Err(error::NegotiationError::StateMismatch),
+                },
+            Ok(Message::ListProtocols) => return Err(error::NegotiationError::StateMismatch),
+            Err(_) => return Err(error::NegotiationError::ParseError(ParseError::InvalidData)),
         };
 
+        match drain_trailing_protocols(tail, len) {
```
dmitry-markin:
```suggestion
        match drain_trailing_protocols(remaining) {
```

src/multistream_select/dialer_select.rs
```diff
@@ -405,10 +438,66 @@ impl WebRtcDialerState {
...
+                actual_length = tail.len(),
+                "Truncated multistream message",
+            );
+
+            return Err(error::NegotiationError::ParseError(ParseError::InvalidData));
+        }
+
+        let payload = tail[..len].to_vec();
+
+        match Message::decode(payload.into()) {
```
dmitry-markin:
```suggestion
        let len_size = remaining.len() - tail.len();
        let payload = remaining.slice(len_size..len_size + len);

        match Message::decode(payload) {
```

src/multistream_select/dialer_select.rs
```diff
@@ -405,10 +438,66 @@ impl WebRtcDialerState {
...
+                    ?error,
+                    message = ?tail[..len],
+                    "Failed to decode multistream message",
+                );
+                return Err(error::NegotiationError::ParseError(ParseError::InvalidData));
+            }
+            _ => return Err(error::NegotiationError::StateMismatch),
+        }
+
+        remaining = &tail[len..];
```
dmitry-markin:
```suggestion
        remaining = remaining.slice(len_size + len..);
```

src/multistream_select/dialer_select.rs
```diff
@@ -405,10 +438,66 @@ impl WebRtcDialerState {
...
+) -> Result<Vec<Protocol>, error::NegotiationError> {
+    let mut protocols = vec![];
+    let mut remaining = &tail[len..];
+
+    loop {
+        if remaining.is_empty() {
+            break;
+        }
+
+        let (len, tail) = unsigned_varint::decode::usize(remaining).map_err(|error| {
```
dmitry-markin:
```suggestion
        let (len, tail) = unsigned_varint::decode::usize(&remaining).map_err(|error| {
```

src/multistream_select/listener_select.rs
```diff
@@ -351,6 +351,14 @@ pub fn webrtc_listener_negotiate<'a>(
     supported_protocols: &'a mut impl Iterator<Item = &'a ProtocolName>,
     payload: Bytes,
 ) -> crate::Result<ListenerSelectResult> {
+    let payload = if payload.len() > 2 && payload[0..payload.len() - 2] != b"\n\n"[..] {
```
dmitry-markin: Is there a way around this without allocation? As the comment in `Message::decode` says, the well-formed message is terminated with a newline. Also, why are we checking for two newlines, but are adding one?

src/transport/webrtc/connection.rs
```diff
@@ -742,6 +743,20 @@ impl WebRtcConnection {
 
                         continue;
                     }
+                    Event::ChannelBufferedAmountLow(channel_id) => {
```
dmitry-markin: Are you sure it is safe to rely on this event for closing the channel? Its original purpose seems to suggest to buffer more data, and not all the data might be drained yet.

dmitry-markin (approved):
Looks good code-wise, but I haven't checked the conformance to the spec.

The only concern is the use of `ChannelBufferedAmountLow` event.

src/transport/webrtc/connection.rs
haikoschol: This change was inherited from [the PR/branch I based this on](https://github.com/paritytech/litep2p/pull/441) and is unrelated to the multistream-select negotiation fixes. I'm happy to remove it if you prefer, but note that the channel is only closed when it was already in `ChannelState::Closing`.

How we use `Event::ChannelBufferedAmountLow` (probably in the context of backpressure) and how channels are closed (when correctly handling `FIN` flags, etc) needs to be revisited in any case.

src/multistream_select/listener_select.rs
haikoschol: It can be avoided by passing a `BytesMut` to the function. While looking into this, I noticed that there are (failing) tests in this module that I didn't run when working on this PR and then I noticed that the tests are not run in GH Actions either. Is this intentional? (cc @lexnv)

haikoschol: > Also, why are we checking for two newlines, but are adding one?

This is quite messy because we are misusing `Message::Protocols` to parse what is actually two distinct multistream-select messages that are often sent inside one `WebRtcMessage`. I wanted to keep the changes in this PR minimal, knowing that we need to rewrite the listener negotiation anyway to support the "back-and-forth" (as you pointed out in #75).

haikoschol: There is a way to avoid appending a newline and using `Message::Protocols` to decode, by reusing `drain_trailing_protocols()`. I've implemented this in [`c0a44c0 `](https://github.com/ChainSafe/litep2p/commit/c0a44c089c3698242eb752bef6c0f5eea872f328).

I have not retested this with libp2p and smoldot. Will do that tomorrow.

haikoschol: > I have not retested this with libp2p and smoldot. Will do that tomorrow.

### litep2p-perf with a libp2p dialer

works (d/l up to 512KiB, u/l up to 16MiB)

### litep2p-perf with [a smoldot dialer](https://github.com/ChainSafe/smoldot/commits/haiko-webrtc-perf/)

litep2p crashes apparently during the noise handshake:

```
2025-11-27T12:45:32.307986Z DEBUG litep2p::webrtc: received non-stun message source=127.0.0.1:63971

thread 'main' (15158228) panicked at /Users/haiko/code/polkadot/litep2p/src/transport/webrtc/opening.rs:432:70:
to succeed: ParseError(InvalidData)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-11-27T12:45:32.308961Z  WARN litep2p::transport-service: transport service closed
2025-11-27T12:45:32.309016Z  WARN litep2p::transport-service: transport service closed
2025-11-27T12:45:32.309022Z  WARN litep2p::transport-service: transport service closed
```

This makes absolutely no sense to me yet. I know that I've tested this before and I haven't made any changes to the smoldot side since. I've tried a few previous commits for litep2p but always had the same result. Need to keep investigating...

### [webrtc-poc](https://github.com/haikoschol/webrtc-poc/blob/haiko-testing/litep2p-server/src/main.rs) with a libp2p dialer

what works:

- noise handshake
- substream opening and multistream-select negotiation from libp2p to litep2p
- ping from libp2p to litep2p

what doesn't work:

- substream opening and multistream-select negotiation from litep2p to libp2p

I thought I had tested this before, but I'm not sure. Maybe I just focused on smoldot. My previous comment [here](https://github.com/paritytech/litep2p/pull/465#issuecomment-3525617599) doesn't mention this pairing either. I've also tried previous commits on this PR branch without success. Need to keep investigating here as well.

### [webrtc-poc](https://github.com/haikoschol/webrtc-poc/blob/haiko-testing/litep2p-server/src/main.rs) with [a smoldot dialer](https://github.com/ChainSafe/smoldot/commits/haiko-libp2p-webrtc)

works

src/transport/webrtc/connection.rs
dmitry-markin: Let's remove it and revisit buffering, closing and backpressure in a separate PR.

src/multistream_select/listener_select.rs
haikoschol: > ### [webrtc-poc](https://github.com/haikoschol/webrtc-poc/blob/haiko-testing/litep2p-server/src/main.rs) with a libp2p dialer
> what works:
> 
> * noise handshake
> * substream opening and multistream-select negotiation from libp2p to litep2p
> * ping from libp2p to litep2p
> 
> what doesn't work:
> 
> * substream opening and multistream-select negotiation from litep2p to libp2p

I figured out what is going on here and opened #494. The relevant insight in the context of this review is that the faulty behavior is not caused by a regression introduced in this PR.

haikoschol: > ### litep2p-perf with [a smoldot dialer](https://github.com/ChainSafe/smoldot/commits/haiko-webrtc-perf/)
> litep2p crashes apparently during the noise handshake:
> 
> ```
> 2025-11-27T12:45:32.307986Z DEBUG litep2p::webrtc: received non-stun message source=127.0.0.1:63971
> 
> thread 'main' (15158228) panicked at /Users/haiko/code/polkadot/litep2p/src/transport/webrtc/opening.rs:432:70:
> to succeed: ParseError(InvalidData)
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> 2025-11-27T12:45:32.308961Z  WARN litep2p::transport-service: transport service closed
> 2025-11-27T12:45:32.309016Z  WARN litep2p::transport-service: transport service closed
> 2025-11-27T12:45:32.309022Z  WARN litep2p::transport-service: transport service closed
> ```
> 
> This makes absolutely no sense to me yet. I know that I've tested this before and I haven't made any changes to the smoldot side since.

I've looked into this issue again and figured out what is going on. I hadn't actually tested this specific combination before (litep2p listener, smoldot dialer, perf behavior). The combinations I tested were:

* libp2p listener, smoldot dialer, perf behavior
* litep2p listener, smoldot dialer, ping in both directions

I have two branches in the ChainSafe smoldot fork; one for the perf behavior ([`haiko-webrtc-perf`](https://github.com/ChainSafe/smoldot/commits/haiko-webrtc-perf/)) and one for ping ([`haiko-libp2p-webrtc`](https://github.com/ChainSafe/smoldot/commits/haiko-libp2p-webrtc)). When I tested the second combination listed above (and found the issue this PR is addressing) I ran into this crash in the noise handshake on the branch `haiko-libp2p-webrtc` and fixed it (or rather avoided it) in [this commit](https://github.com/ChainSafe/smoldot/commit/10db21bc7d1ad07d8e578e637caf3d1ab10520e9). Making the same change on the other branch also avoids this crash.

What happens is that without this fix, smoldot sends a two byte long message (i.e. an incomplete WebRTC protobuf frame), which fails to decode [here](https://github.com/paritytech/litep2p/blob/925caa9dac387fff250ec1bfcf170b8cefd57be7/src/transport/webrtc/opening.rs#L256) and panics due to an `.expect()` [here](https://github.com/paritytech/litep2p/blob/925caa9dac387fff250ec1bfcf170b8cefd57be7/src/transport/webrtc/opening.rs#L432).

In short, this crash is not a regression caused by the changes in this PR.

haikoschol: @dmitry-markin can you please go through my comments above and resolve this thread if you are happy with the change I made in [c0a44c0](https://github.com/ChainSafe/litep2p/commit/c0a44c089c3698242eb752bef6c0f5eea872f328)?

src/multistream_select/dialer_select.rs
```diff
@@ -357,24 +358,59 @@ impl WebRtcDialerState {
...
         payload: Vec<u8>,
     ) -> Result<HandshakeResult, crate::error::NegotiationError> {
-        let Message::Protocols(protocols) =
-            Message::decode(payload.into()).map_err(|_| ParseError::InvalidData)?
-        else {
-            return Err(crate::error::NegotiationError::MultistreamSelectError(
-                NegotiationError::Failed,
-            ));
+        // All multistream-select messages are length-prefixed. Since this code path is not using
+        // multistream_select::protocol::MessageIO, we need to decode and remove the length here.
```
lexnv, line 362:
```suggestion
        // multistream_select::protocol::MessageIO. We need to decode and remove the length here.
```

lexnv: Lets do this or similar, the CI passed but it doesn't allow us to merge your PR, let's retriggerit

haikoschol: I just merged `master` to poke it.

Issue #471, opened 2025-11-12
kad: Allow connecting to more than one DHT network simultaneously


dmitry-markin:
This is needed for complete IPFS support in Substrate, because we need to participate in both Polkadot & IPFS DHTs.

Note that this is not the same as supporting additional/fallback DHT protocol names, because the latter leads to the DHT networks becoming intertwined.

PR #473, opened 2025-11-13
kad: Allow connecting to more than one DHT network

dmitry-markin:
Needed for IPFS support in substrate, because we need to connect to polkadot & IPFS DHTs simultaneously.

Close https://github.com/paritytech/litep2p/issues/471.

### DIFF
--- a/src/protocol/transport_service.rs
+++ b/src/protocol/transport_service.rs
@@ -564,7 +564,11 @@ impl Stream for TransportService {
         while let Poll::Ready(event) = self.rx.poll_recv(cx) {
             match event {
                 None => {
-                    tracing::warn!(target: LOG_TARGET, "transport service closed");
+                    tracing::warn!(
+                        target: LOG_TARGET,
+                        protocol = ?protocol_name,
+                        "transport service closed"
+                    );
                     return Poll::Ready(None);
                 }
                 Some(InnerTransportEvent::ConnectionEstablished {


dmitry-markin:
I would like to release this as v0.12.1 (because the public API hasn't changed) and include into litep2p upgrade PR https://github.com/paritytech/polkadot-sdk/pull/9685 in polkadot-sdk.

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -240,7 +240,9 @@ impl Kademlia {
                 self.routing_table.on_connection_established(Key::from(peer), endpoint);
 
                 let Some(actions) = self.pending_dials.remove(&peer) else {
-                    entry.insert(PeerContext::new());
+                    // Note that we do not add peer entry if we don't have any pending actions.
+                    // This is done to not populate `self.peers` with peers that don't support
```
lexnv, line 244: dq: This won't affect networks with a single DHT started? Maybe we could add a trace here in case some issues popup in the future?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -343,6 +345,7 @@ impl Kademlia {
         let pending_action = &mut self
             .peers
             .get_mut(&peer)
+            // If we opened an outbound substream, we must have pending actions for the peer.
```
lexnv, line 348: Couldn't we get in a race case here between the following timelines?

- T0: Pending to open outbound substream
- T1: Outbound opened and queue for reporting
- T2: disconnect_peer - the peer is disconnected and reported
- T3: Outbound opened reproted now

src/protocol/libp2p/kademlia/mod.rs
lexnv: Thinking outloud: If that's the case we could maybe add a new builder method on the KadConfig to signal we run in multi-dht-worlds? And return none only on multi-dht?

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: This might happen when there was an error reading inbound request at the same time as we sent an outbound request. The worst that can happen is we won't process pending actions for a peer, but this is not much different to an error during outbound request.

As a side note, the PR doesn't change the way this race can happen.

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin: This won't affect a single DHT network, as the entry is always inserted anyway when substream is opened. Strictly speaking, it doesn't make sense to consider transport-level connected peers as connected, because they might not speak the Kademlia protocol (even in a single DHT case). We are interested in substreams over a specific protocol.

dmitry-markin: > Thinking outloud: If that's the case we could maybe add a new builder method on the KadConfig to signal we run in multi-dht-worlds? And return none only on multi-dht?

The logic shouldn't be different for a single DHT versus multi-DHT cases.

PR #478, opened 2025-11-21
address: Add failed to dial addresses bucket

lexnv:
This PR ensures that litep2p keeps track of addresses that failed to dial in the past, up to 64 addresses.

This ensures resources are preserved and failed addresses are no longer dialed.

Litep2p will no longer dial addresses that are marked as unrecoverable:
- failures of ` DialError::AddressError`
- failures of `NegotiationError::PeerIdMismatch`, which means the address corresponds to a different peer than the expected one

### DIFF
--- a/src/transport/manager/address.rs
+++ b/src/transport/manager/address.rs
@@ -536,7 +536,7 @@ mod tests {
     fn evict_on_capacity() {
         let mut store = AddressStore {
             addresses: HashMap::new(),
-            known_bad_addresses: HashSet::new(),
+            known_bad_addresses: HashMap::new(),
             max_capacity: 2,
         };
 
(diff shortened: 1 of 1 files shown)


src/transport/manager/address.rs
```diff
@@ -222,6 +232,11 @@ impl AddressStore {
     /// If the address is not in the store, it will be inserted.
     /// Otherwise, the score and connection ID will be updated.
     pub fn insert(&mut self, record: AddressRecord) {
+        if record.score == scores::ADDRESS_FAILURE {
```
dmitry-markin:
```suggestion
        if record.score <= scores::UNRECOVERABLE_FAILURES {
```
This way it will look less like some magic number using.

src/transport/manager/address.rs
```diff
@@ -28,6 +32,9 @@ use std::collections::{hash_map::Entry, HashMap};
 /// Maximum number of addresses tracked for a peer.
 const MAX_ADDRESSES: usize = 64;
 
+/// Duration for which a bad address is kept in the known bad list.
+const BAD_ADDRESS_EXPIRATION: Duration = Duration::from_secs(15 * 60); // 15 Minutes
```
dmitry-markin, line 36:
```suggestion
const BAD_ADDRESS_EXPIRATION: Duration = Duration::from_secs(5 * 60); // 5 Minutes
```
What about 5 mins?

src/transport/manager/address.rs
```diff
@@ -218,22 +231,22 @@ impl AddressStore {
...
             return;
         }
 
-        // The eviction algorithm favours addresses with higher scores.
-        //
-        // This algorithm has the following implications:
-        //  - it keeps the best addresses in the store.
-        //  - if the store is at capacity, the worst address will be evicted.
-        //  - an address that is not dialed yet (with score zero) will be preferred over an address
-        //  that already failed (with negative score).
```
dmitry-markin, line 236: Have you deleted the comments on purpose?

src/transport/manager/address.rs
```diff
@@ -242,17 +255,49 @@ impl AddressStore {
...
-        // Insert the record.
         self.addresses.insert(record.address.clone(), record);
     }
 
+    /// Mark an address as unrecoverable.
+    ///
+    /// This will remove the address from the store and add it to the known bad addresses
+    /// with a timestamp.
+    fn unrecoverable_address(&mut self, address: Multiaddr) {
+        self.addresses.remove(&address);
```
dmitry-markin, line 272: This will lead to permanently banning the address even after the ban expires, better to exclude the address when getting it in `addresses()`.

src/transport/manager/address.rs
dmitry-markin, line 305: We should check the banned addresses here and unban the ones with the ban expired & exclude the ones with the active ban.

src/transport/manager/address.rs
```diff
@@ -200,14 +210,17 @@ impl AddressStore {
...
         }
     }
 
     /// Get the score for a given error.
     pub fn error_score(error: &DialError) -> i32 {
         match error {
-            DialError::AddressError(_) => scores::ADDRESS_FAILURE,
+            DialError::AddressError(_) => scores::UNRECOVERABLE_FAILURES,
+            DialError::NegotiationError(NegotiationError::PeerIdMismatch(_, _)) =>
+                scores::UNRECOVERABLE_FAILURES,
```
dmitry-markin, line 223: In this case we should ban the address, but not set its score to  `i32::MIN` to allow connecting once the ban expires. I don't think we can do this using the magic number, better to return a dedicated type, or at least (i32, bool).

PR #480, opened 2025-11-24
transport/manager: Enhance local address checks

lexnv:
This PR enhances the local address checks that the transport manager performs before adding new addresses to the local address store.

This makes the manager more robust wrt:
-  the local node is listening on any interfaces (0.0.0.0:port)
- a `127.0.0.1:port` address is resolved to the same listen address


src/transport/manager/handle.rs
```diff
@@ -164,14 +165,66 @@ impl TransportManagerHandle {
...
             .iter()
             .take_while(|protocol| !std::matches!(protocol, Protocol::P2p(_)))
             .collect();
 
-        self.listen_addresses.read().contains(&address)
+        // Check for the exact match.
+        let listen_addresses = self.listen_addresses.read();
+        if listen_addresses.contains(&address) {
+            return true;
+        }
```
dmitry-markin, line 197: This should handle everything, because `self.listen_addresses` are expanded to include addresses of all local interfaces when creating a listener socket:
https://github.com/paritytech/litep2p/blob/3de3785f5308e646c5b5c3203d423d9601d170c1/src/transport/common/listener.rs#L272-L306

lexnv: Hmm, if that's the case, we shouldn't see logs like:

```rust
litep2p::transport-manager: [Parachain] add known addresses peer=PeerId("A")

peer_addresses={
"/ip4/127.0.0.1/tcp/30333/p2p/...",
```

This gets printed after the addresses are already checked 🤔 

https://github.com/paritytech/litep2p/blob/5e4e6bffa2a033bd572818010c9e4726036f7016/src/transport/manager/handle.rs#L247-L249

https://github.com/paritytech/litep2p/blob/5e4e6bffa2a033bd572818010c9e4726036f7016/src/transport/manager/handle.rs#L275-L280

Maybe we are getting on the else branch of `if address.ip().is_unspecified()`:

https://github.com/paritytech/litep2p/blob/3de3785f5308e646c5b5c3203d423d9601d170c1/src/transport/common/listener.rs#L307-L309

This might mean that there's a node CLI config which won't call into `Interface::show()` and therefore accept localhost addresses wrongly. However I started my node without modifying any defaults wrt listen addresses, so this might be coming in part from substrate. 

Still would be good to close the gap on the manager handler as well, wdyt? 🙏

dmitry-markin: They can differ in `/ws/` component. The local node is likely not a validator and listens on `.../ws` address, while the added address doesn't have a `/ws` component.

src/transport/manager/handle.rs
```diff
@@ -164,14 +165,66 @@ impl TransportManagerHandle {
         }
     }
 
+    /// Helper to extract IP and Port from a Multiaddr
+    fn extract_ip_port(&self, addr: &Multiaddr) -> Option<(IpAddr, u16)> {
```
dmitry-markin:
```suggestion
    fn extract_ip_port(addr: &Multiaddr) -> Option<(IpAddr, u16)> {
```

PR #482, opened 2025-11-24
Allow all the Bitswap CIDs (v1) to pass regardless of used hash

bkontur:
Relates to the: https://github.com/paritytech/polkadot-sdk/issues/10301

Fully tested with https://github.com/paritytech/polkadot-bulletin-chain/pull/100

## Rationale

The only weak point is that `BitswapRequestHandler` expects a Blake2b hash (due to `BlockT`) here: [https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/network/src/bitswap/mod.rs#L208-L209](https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/network/src/bitswap/mod.rs#L208-L209). However, this is not a real problem, because we initially create a default 32-byte hash and then replace it with the 32-byte hash from the CID. Since we only claim to support 32-byte hashes, it just works. 😊

## Solution

[Proposed by](https://github.com/paritytech/litep2p/pull/482#pullrequestreview-3504227563) @dmitry-markin - remove CID filtering from litep2p and do filtering in PolkadotSDK - [see more](https://github.com/paritytech/litep2p/pull/482#issuecomment-3574854766).

## Follow-up

As a follow-up, I would refactor `fn indexed_transaction` / `fn has_indexed_transaction` here: [https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/api/src/client.rs#L154-L163](https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/api/src/client.rs#L154-L163), and make the following change:
```diff
+        type ContentHash = [u8; 32];

	/// Get single indexed transaction by content hash.
	///
	/// Note that this will only fetch transactions
	/// that are indexed by the runtime with `storage_index_transaction`.
-	fn indexed_transaction(&self, hash: Block::Hash) -> sp_blockchain::Result<Option<Vec<u8>>>;
+	fn indexed_transaction(&self, hash: ContentHash) -> sp_blockchain::Result<Option<Vec<u8>>>;

	/// Check if transaction index exists.
-	fn has_indexed_transaction(&self, hash: Block::Hash) -> sp_blockchain::Result<bool> {
-	fn has_indexed_transaction(&self, hash: ContentHash) -> sp_blockchain::Result<bool> {
```

Ultimately, we index transactions by their content hash, which is a `[u8; 32]` here:
[https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/primitives/io/src/lib.rs#L1494](https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/primitives/io/src/lib.rs#L1494),
and it is generated as a `[u8; 32]` here:
[https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/frame/transaction-storage/src/lib.rs#L239-L242](https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/frame/transaction-storage/src/lib.rs#L239-L242).

### DIFF
--- a/src/protocol/libp2p/bitswap/mod.rs
+++ b/src/protocol/libp2p/bitswap/mod.rs
@@ -29,7 +29,6 @@ use crate::{
 };
 
 use cid::Version;
-pub use multihash::Code as MultihashCode;
 use prost::Message;
 use tokio::sync::mpsc::{Receiver, Sender};
 use tokio_stream::{StreamExt, StreamMap};


src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +164,55 @@ impl Bitswap {
                 let want_type = match entry.want_type {
                     0 => WantType::Block,
                     1 => WantType::Have,
-                    _ => return None,
+                    _ => {
+                        tracing::warn!(
```
lexnv: nit: Since these warnings can be quite easily user-generated, and not necesarily something we support for our bitswap logic, I would downgrade them to `debug` to avoid poluting the logs

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +164,55 @@ impl Bitswap {
...
+                        return None;
+                    }
                 };
 
-                (cid.version() == cid::Version::V1
-                    && cid.hash().code() == u64::from(Code::Blake2b256)
-                    && cid.hash().size() == 32)
-                    .then_some((cid, want_type))
+                // Check supported CID versions.
+                match cid.version() {
```
lexnv: nit: I find `if let` more readable, but feel free to ignore

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +164,55 @@ impl Bitswap {
...
+                        "Unsupported multihash size: {size} for cid: {cid}, supports only 32!"
+                    );
+                    return None;
+                }
+
+                // Check supported multihash a.k.a. hashing algorithm.
+                let code = cid.hash().code();
+                if code != u64::from(Code::Blake2b256)
+                    && code != u64::from(Code::Sha2_256)
+                    && code != u64::from(Code::Keccak256)
```
lexnv: Would it be better to extend this functionality and make it configurable via `bitswap::Config`?

Then, we can add a group of supported hash codes:

```rust

// In config:
Config {
    ...
    supported_hash_codes: HashSet<u64>,
}
impl Default for Config {
    supported_hash_codes: [u64::from(Code::Blake2b256)].iter().collect(),
}

/// Builder to set supported hash codes for bitswap requests.
///
/// Defaults to supporting only `Code::Blake2b256`.
fn with_supported_hash_codes(&self, codes: impl Iter<Item = Code>) {
    self.supported_hash_codes.extend(codes.map(u64::from));
}

// In bitswap:
if self.supported_hash_codes.contains(..)

Comment (bkontur) [reply src/protocol/libp2p/bitswap/mod.rs]: > Would it be better to extend this functionality and make it configurable via `bitswap::Config`?

@lexnv Yes, yes, that would be much better. And where do I set those hard-coded values then?
Here?
https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/network/src/litep2p/shim/bitswap.rs#L47
```
let (config, handle) = Config::new()
.with_supported_hash_codes(blake)
.with_supported_hash_codes(sha2)
.with_supported_hash_codes(keccak)
.build();
```

Comment (lexnv) [reply src/protocol/libp2p/bitswap/mod.rs]: Exactly 👍

File: Cargo.toml [hunk 4]
```diff
@@ -25,7 +25,7 @@ indexmap = { version = "2.9.0", features = ["std"] }
 libc = "0.2.158"
 mockall = "0.13.1"
 multiaddr = "0.17.0"
-multihash = { version = "0.17.0", default-features = false, features = ["std", "multihash-impl", "identity", "sha2", "blake2b"] }
+multihash = { version = "0.17.0", default-features = false, features = ["std", "multihash-impl", "identity", "sha2", "sha3", "blake2b"] }
```
dr333ws: @bkontur I also see "Keccak256" in src/protocol/libp2p/bitswap/mod.rs among other hashing algorithm. 
It needs to be imported, isn't it?

src/protocol/libp2p/bitswap/mod.rs
dr333ws: done

src/protocol/libp2p/bitswap/mod.rs
dr333ws: changed

src/protocol/libp2p/bitswap/mod.rs
dr333ws: Thanks for the suggestions, folks!
I implemented what you suggested ~but with one minor details:~
- ~`default_hash_codes` are blake, sha2 and Keccak256 -- to maintain backward compatibility with the present code~

@lexnv , PTAL at the new code

Cargo.toml
bkontur: I just added  Code::Keccak256 and it is feature gated:
```
    #[cfg(feature = "sha3")]
    #[mh(code = 0x1b, hasher = crate::Keccak256)]
    Keccak256,
```

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -61,8 +64,35 @@ impl Config {
                 event_tx,
                 protocol: ProtocolName::from(PROTOCOL_NAME),
                 codec: ProtocolCodec::UnsignedVarint(Some(MAX_PAYLOAD_SIZE)),
+                supported_hash_codes: default_hash_codes(),
```
bkontur: We want to keep there `Blake2b256` by default - backwards compatibility

```suggestion
                supported_hash_codes: std::collections::HashSet::from([u64::from(multihash::Code::Blake2b256)]);
```

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -61,8 +64,35 @@ impl Config {
...
+
+/// Default set of supported hash codes.
+fn default_hash_codes() -> std::collections::HashSet<u64> {
+    [
+        u64::from(multihash::Code::Blake2b256),
+        u64::from(multihash::Code::Sha2_256),
+        u64::from(multihash::Code::Keccak256),
+    ]
+    .into_iter()
+    .collect()
```
bkontur: And then these three values will be provided in the Polkadot-SDK as mentioned here: https://github.com/paritytech/litep2p/pull/482#discussion_r2556775137

```suggestion
```

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -61,8 +64,35 @@ impl Config {
...
+    /// Set supported multihash codes for bitswap CID validation.
+    /// # Example
+    ///
+    /// ```rust
+    /// let (config, handle) = Config::new()
+    ///     .with_supported_hash_codes([Code::Blake2b256, Code::Sha2_256]);
+    /// ```
+    pub fn with_supported_hash_codes(
+        mut self,
+        codes: impl IntoIterator<Item = multihash::Code>,
```
bkontur: @lexnv hmm, this way, we would need to add `multihash` to the `sc-network`, what about just re-exporting here in this file:
```
pub use multihash::Code;
or
pub use multihash::Code as MultihashCode;
```

src/protocol/libp2p/bitswap/mod.rs
bkontur: well, if anybody adds new CID version2, the `match` would catch it and makes us handle it, but let's see :)

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +167,48 @@ impl Bitswap {
...
+                    }
                 };
 
-                (cid.version() == cid::Version::V1
-                    && cid.hash().code() == u64::from(Code::Blake2b256)
-                    && cid.hash().size() == 32)
-                    .then_some((cid, want_type))
+                // Check supported CID versions.
+                if let Version::V0 = cid.version() {
+                    tracing::warn!(
```
bkontur:
```suggestion
                    tracing::trace!(
```

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +167,48 @@ impl Bitswap {
...
+                        "Unsupported CID version {:?} for cid=: {cid}",
+                        cid.version()
+                    );
+                    return None;
+                }
+
+                // Check supported multihash length (only 32).
+                let size = cid.hash().size();
+                if size != 32 {
+                    tracing::warn!(
```
bkontur:
```suggestion
                    tracing::debug!(
```

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +167,48 @@ impl Bitswap {
...
+                        target: LOG_TARGET,
+                        "Unsupported multihash size: {size} for cid: {cid}, supports only 32!"
+                    );
+                    return None;
+                }
+
+                // Check supported multihash a.k.a. hashing algorithm.
+                let code = cid.hash().code();
+                if !self.supported_hash_codes.contains(&code) {
+                    tracing::warn!(
```
bkontur:
```suggestion
                    tracing::debug!(
```

src/protocol/libp2p/bitswap/mod.rs
lexnv: The code is not compiling, let me know when that's fixed 🙏

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -47,6 +47,9 @@ pub struct Config {
 
     /// RX channel for receiving commands from the user.
     pub(super) cmd_rx: Receiver<BitswapCommand>,
+
+    /// Supported multihash codes for CID validation.
+    pub(super) supported_hash_codes: std::collections::HashSet<u64>,
```
lexnv: Let's import this 🙏

src/protocol/libp2p/bitswap/config.rs
lexnv: `pub use multihash::Code as MultihashCode;` sounds good to me

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -61,8 +64,24 @@ impl Config {
...
     }
+
+    /// Set supported multihash codes for bitswap CID validation.
+    /// # Example
+    ///
+    /// ```rust
+    /// let (config, handle) = Config::new()
+    ///     .with_supported_hash_codes([Code::Blake2b256, Code::Sha2_256]);
+    /// ```
+    pub fn with_supported_hash_codes(
```
lexnv: nit: Let's mention we are operating with `Blake2b256` by default

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -61,8 +64,24 @@ impl Config {
...
+                supported_hash_codes: std::collections::HashSet::from([u64::from(multihash::Code::Blake2b256)]);
             },
             BitswapHandle::new(event_rx, cmd_tx),
         )
     }
+
+    /// Set supported multihash codes for bitswap CID validation.
+    /// # Example
+    ///
+    /// ```rust
```
lexnv: This will probably compile the example, maybe we want to fix the doc example with some imports behind `#` so they don't get rendered in docsrs 🙏

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -121,6 +123,7 @@ impl Bitswap {
             service,
             cmd_rx: config.cmd_rx,
             event_tx: config.event_tx,
+            supported_hash_codes: config.supported_hash_codes,
```
lexnv: Nit: let's add a debug in the `fn new` to know the supported hashes, just in case we start hitting `Unsupported multihash code: {code} for cid: {cid}`; or add there the full supported list (since it shouldn't be that big)

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +167,48 @@ impl Bitswap {
...
+                    );
+                    return None;
+                }
+
+                // Check supported multihash a.k.a. hashing algorithm.
+                let code = cid.hash().code();
+                if !self.supported_hash_codes.contains(&code) {
+                    tracing::debug!(
+                        target: LOG_TARGET,
+                        "Unsupported multihash code: {code} for cid: {cid}"
```
lexnv:
```suggestion
                        code,
                        cid,
                        expected = ?self.supported_hash_codes,
                        "Unsupported multihash"
```

.gitignore
```diff
@@ -1 +1,2 @@
 /target
+.idea
```
lexnv: nit: This is unrelated to the PR

Cargo.toml
lexnv: @dmitry-markin would love to get your thoughts on this 🙏 

Should we consider adding a new feature flag to optionally enable `multihash/sha3` such that we dont pollute polkadot-sdk with features we don't need?

bkontur:
> IMO we can simplify the PR substantially.
> 
> I don't think we should complicate the litep2p Bitswap configuration with a multihash code filter. It would be more straightforward to pass CIDs as is to the client code, that knows anyway what codes/codecs it supports. The client code can then decide whether to respond with a `DontHave` message straight away, or silently drop such request.


I like this also, so basically, you are saying, we just remove the CID filter from litep2p and we do this filtering in the PolkadotSDK part, right?

Is anybody else using litep2p bitswap handler? Just asking, that we allow all CIDs now :)

bkontur:
@x3c41a ok, so:
- [ ] let's revert all these changes here
- [ ] the only change here would be that we remove original CID filtering, we allow to pass all the CIDs
- [ ] when merged, release litep2p

then open a new PolkadotSDK PR:
- [ ] add `supported_hash_codes` to the `BitswapServer` and in the `new()` just allow all three
- [ ] basically, move this CID filtering somewhere here : https://github.com/paritytech/polkadot-sdk/blob/e588acf5e12b14c68331d2e08afe7c12d6671df9/substrate/client/network/src/litep2p/shim/bitswap.rs#L63-L64
```
.into_iter()
.filter(|(cid, want_type)| filter_cid(..)) // <- better name
.map(|(cid, want_type)| {
```

- [ ] bump PolkadotSDK with released litep2p

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -167,10 +167,7 @@ impl Bitswap {
                     _ => return None,
```
bkontur, line 166:
```suggestion
                    _ => {
                        tracing::debug!(
                            target: LOG_TARGET,
                            "Unhandled `WantList` type: {}",
                            entry.want_type
                        );
                        return None;
                    }
```

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -164,13 +164,17 @@ impl Bitswap {
...
                     1 => WantType::Have,
-                    _ => return None,
+                    _ => {
+                        tracing::debug!(
+                            target: LOG_TARGET,
+                            "Unhandled `WantList` type: {}",
+                            entry.want_type
+                        );
+                        return None;
+                    }
```
dmitry-markin:
```suggestion
                    _ => return None,
```
There are no other want types as per [Bitswap spec](https://specs.ipfs.tech/bitswap-protocol/), and we better not log anything if the garbage is coming from the network. We can filter such requests though to not pass to the client code.

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -29,7 +29,7 @@ use crate::{
 };
 
 use cid::Version;
-use multihash::Code;
+pub use multihash::Code as MultihashCode;
```
dmitry-markin:
```suggestion
```

Issue #477, opened 2025-11-19
webrtc: Very low bandwidth on substreams


haikoschol:
https://github.com/lexnv/litep2p-perf/pull/2 adds support for measuring bandwidth on WebRTC connections using rust-libp2p and litep2p. Both receiving data and sending data on a WebRTC substream is very slow in litep2p (see the table in the [`README`](https://github.com/timwu20/litep2p-perf/blob/d47573b4360efa82371613d4b4b0cddc1a71e092/README.md)).

When receiving data, bandwidth seems to drop by 10x when going from 128KiB to 256 KiB and continues to degrade from there.

When sending data, bandwidth is much lower compared to all other tested pairings, but seems to actually increase up to 512KiB. Sending 1.0MiB however seems to never finish.

lexnv:
Thanks @haikoschol, the 128MiB must be a separate issue we are hitting 🙏 

```
2025-11-25T07:36:34.178232Z DEBUG str0m::ice_::agent: Remove failed pair: CandidatePair(0-0 (host-host) prio=9151313343271665150 state=Waiting attempts=0 unanswered=0 remote=1 nom=Success)
```

This line must be the culprit, I'm not sure what the meaning is behind`state=Waiting.` Thinking out loud, a few places I would look into:
- Maybe we reached a timeout somewhere, and the state was never advanced to `Processing`? How long does it take to hit this line since the beginning of the log file?
- Maybe the socket itself has underlying issues due to our buffering or IO error?
- while at it, it might be worth enabling trace levels on the str0m crate if not already done

I would still create a dedicated PR for the `tx.reserve` context waker fix and look into this one afterwards

PR #483, opened 2025-11-25
webrtc: Avoid future being dropped when channel is full

haikoschol:
This PR addresses part of #477. Specifically the issue described in [this comment](https://github.com/paritytech/litep2p/issues/477#issuecomment-3570583758).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.12.3] - 2025-12-16
+
+This release improves the robustness of the multistream-select negotiation over WebRTC transport and fixes inbound bandwidth metering on substreams. It also enhances the dialing success rate by improving the transport dialing logic. Additionally, it re-exports CID's multihash to facilitate the construction of CID V1.
+
+### Changed
+
+- transports: Improves the robustness and success rate of connection dialing  ([#495](https://github.com/paritytech/litep2p/pull/495))
+- types: Re-export cid's multihash to construct CID V1  ([#491](https://github.com/paritytech/litep2p/pull/491))
+
+### Fixed
+
+- fix: multistream-select negotiation on outbound substream over webrtc  ([#465](https://github.com/paritytech/litep2p/pull/465))
+- substream: Fix inbound bandwidth metering  ([#499](https://github.com/paritytech/litep2p/pull/499))
 
 ## [0.12.2] - 2025-11-28
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1837,7 +1837,7 @@ checksum = "6373607a59f0be73a39b6fe456b8192fcc3585f602af20751600e974dd455e77"
 
 [[package]]
 name = "litep2p"
-version = "0.12.2"
+version = "0.12.3"
 dependencies = [
  "async-trait",
  "bs58 0.5.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "litep2p"
 description = "Peer-to-peer networking library"
 repository = "https://github.com/paritytech/litep2p"
 license = "MIT"
-version = "0.12.2"
+version = "0.12.3"
 edition = "2021"
 
 # cargo-machete does not detect serde_millis usage, so we ignore the warning


src/transport/webrtc/substream.rs
```diff
@@ -190,29 +197,40 @@ impl tokio::io::AsyncRead for Substream {
...
+        mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
         buf: &[u8],
     ) -> Poll<Result<usize, std::io::Error>> {
         if let State::SendClosed = *self.state.lock() {
             return Poll::Ready(Err(std::io::ErrorKind::BrokenPipe.into()));
         }
 
-        // TODO: try to coalesce multiple calls to `poll_write()` into single `Event::Message`
+        loop {
```
haikoschol: This loop is part of the code generated by Gemini, but I think it will only ever be executed once.

haikoschol:
> Draft for two reasons:
> 
> 1. So I can try avoiding the `BoxFuture` and associated heap allocation

We can avoid the future that yields a permit entirely by calling `self.tx.try_reserve()`. I've tested this change on a branch that included my changes from #465 successfully for up- & download of 32 MiB. Beyond that, the connection dies. That is caused by a separate issue described in [this comment](https://github.com/paritytech/litep2p/issues/477#issuecomment-3575306819).

src/transport/webrtc/substream.rs
```diff
@@ -198,17 +198,20 @@ impl tokio::io::AsyncWrite for Substream {
...
-
-        let permit = match futures::ready!(future.poll(cx)) {
-            Err(_) => return Poll::Ready(Err(std::io::ErrorKind::BrokenPipe.into())),
+        let permit = match self.tx.try_reserve() {
             Ok(permit) => permit,
+            Err(err) =>
+                return match err {
+                    TrySendError::Full(_) => {
+                        cx.waker().wake_by_ref();
+                        Poll::Pending
```
lexnv: This will effectively enter into a busy loop until there's capacity in the buffer:
- we call `poll_write` and the channel is full
- we wake up the waker if there's no capacity
- the runtime calls `poll_write` again, but without any context of the `tx` being full or not
- we enter the same code path until `tx` has available capacity, wake the waker and return Pending

haikoschol: Ah, that makes sense. I've added a shared `AtomicWaker` to let the consumer do the wakeup, but I'm not sure this is a solid solution. Let me know whether you prefer this or going back to the `BoxFuture`.

src/transport/webrtc/substream.rs
```diff
@@ -144,7 +154,9 @@ impl Stream for SubstreamHandle {
     type Item = Event;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.rx.poll_recv(cx)
+        let item = self.rx.poll_recv(cx);
+        self.write_waker.wake();
```
lexnv: How could we tell from the `poll_next` rx channel that the `poll_write` tx channel has free capacity?

Isn't this introducing a slightly different variation of the same class of issue (CPU looping)?

What would happen if we save the waker in `poll_write`, but the channel has capacity (meaning we start sending right away)? Wouldn't we wake the context needlessly on every poll_next?

haikoschol: > How could we tell from the `poll_next` rx channel that the `poll_write` tx channel has free capacity?

`SubstreamHandle::rx` is [the other side](https://github.com/paritytech/litep2p/blob/791279681de32c64daeb39b90ba1e2351b5cfe27/src/transport/webrtc/substream.rs#L83) of `Substream::tx`.

When a protocol wants to send data, it calls `Substream::poll_write()`, which writes the data to `Substream::tx`. `WebRtcConnection::run()` [polls the associated `SubstreamHandle`](https://github.com/paritytech/litep2p/blob/630aa379919a47ddea50dfd44b507086bdd93c6b/src/transport/webrtc/connection.rs#L788), which pulls the data from `SubstreamHandle::rx`.

BTW, When I double-checked this, I noticed that the doc comments on `SubstreamHandle::tx` and `SubstreamHandle::rx` are incorrect. On `rx` it says "RX channel for receiving messages from `peer`." It probably should say something like "RX channel for receiving outbound messages from the associated `Substream` instance."

> Isn't this introducing a slightly different variation of the same class of issue (CPU looping)?

I don't think so but don't know how to verify that. I'm out of my depth here on the inner workings of Rust async and tokio, so I'm relying on AI. 😐

> What would happen if we save the waker in `poll_write`, but the channel has capacity (meaning we start sending right away)? Wouldn't we wake the context needlessly on every poll_next?

Yes, there would be spurious wakes, although I'm not sure they would happen on *every* `poll_next()`. According to GPT 5.1:

> In almost all realistic workloads, the cost of occasional spurious wakeups is much lower than the cost of introducing a heap allocation via `reserve_owned()` + `BoxFuture`.

haikoschol: I've added a test in [`b88b11e`](https://github.com/paritytech/litep2p/pull/483/commits/b88b11e19dc194559e745eb707ec036bd475f061) that is supposed to verify that the `AtomicWaker` approach works.

lexnv: Yep it makese sense 🙏 

After digging a bit into this issue, I believe we could use tokio-util/[PollSender](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.PollSender.html) here or similar. It will basically allocate the future once, and reutilize the memory allocated for subsequent polls

src/transport/webrtc/substream.rs
```diff
@@ -473,4 +473,47 @@ mod tests {
         )
         .await;
     }
+
+    #[tokio::test]
+    async fn backpressure_released_wakes_blocked_writer() {
```
lexnv, line 462: Nice! 🚀

src/transport/webrtc/substream.rs
haikoschol: I've replaced the `AtomicWaker` with `PollSender`. Unit tests pass. I'll manually test with litep2p-perf now.

haikoschol: > I'll manually test with litep2p-perf now.

works

src/transport/webrtc/substream.rs
```diff
@@ -144,7 +147,9 @@ impl Stream for SubstreamHandle {
     type Item = Event;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        self.rx.poll_recv(cx)
+        let item = self.rx.poll_recv(cx);
+        cx.waker().wake_by_ref();
```
lexnv: This should no longer be needed

haikoschol: Missed that one, thanks!

src/transport/webrtc/substream.rs
```diff
@@ -190,49 +193,45 @@ impl tokio::io::AsyncRead for Substream {
...
-        futures::pin_mut!(future);
-
-        let permit = match futures::ready!(future.poll(cx)) {
+        match futures::ready!(self.tx.poll_reserve(cx)) {
+            Ok(()) => {}
             Err(_) => return Poll::Ready(Err(std::io::ErrorKind::BrokenPipe.into())),
-            Ok(permit) => permit,
         };
 
+        let num_bytes = std::cmp::min(MAX_FRAME_SIZE, buf.len());
```
dmitry-markin, line 209: This means we silently drop everything above `MAX_FRAME_SIZE`. It's better to drop the entire frame here and generate a warning. I.e., it' doesn't make sense sending partial frame if part of the data is going to be discarded, but we need to notify the developer of the transport misuse.

src/transport/webrtc/substream.rs
dmitry-markin:
```suggestion
/// Substream handle that is given to the WebRTC transport backend.
```

src/transport/webrtc/substream.rs
haikoschol: Agreed. I just changed it to return an error and noticed that [some of the tests actually expect to be able to write more than `MAX_FRAME_SIZE`](https://github.com/paritytech/litep2p/blob/785a89fec191e0239ac107e4a9028e5a42ca12b1/src/transport/webrtc/substream.rs#L264). Not sure what the idea behind that was, just wanted to point it out.

haikoschol: @dmitry-markin is [`4d8cfb3`](https://github.com/paritytech/litep2p/pull/483/changes/4d8cfb31bc2675238b944273a9ed930ad7183e02) what you had in mind?

dmitry-markin: > [some of the tests actually expect to be able to write more than `MAX_FRAME_SIZE`](https://github.com/paritytech/litep2p/blob/785a89fec191e0239ac107e4a9028e5a42ca12b1/src/transport/webrtc/substream.rs#L264)

Then I got it wrong and we must be able to write arbitrary-sized data to WebRTC streams, not discarding or trimming anything. It's up to WebRTC transport implementation to handle framing internally and respect `MAX_FRAME_SIZE`. The test should work as it is, please revert the changes.

haikoschol: Since this is out of scope for this PR, how about I create an issue, reference it in the code and we can merge this?

haikoschol: > Since this is out of scope for this PR, how about I create an issue, reference it in the code and we can merge this?

#502

dmitry-markin: Did the test work before your PR? Then we shouldn't create an issue for it, but instead not break it in the PR.

haikoschol: Yes, the tests passed before (and now again after the revert).

src/transport/webrtc/substream.rs
```diff
@@ -190,49 +194,53 @@ impl tokio::io::AsyncRead for Substream {
...
-        let permit = match futures::ready!(future.poll(cx)) {
+        match futures::ready!(self.tx.poll_reserve(cx)) {
+            Ok(()) => {}
             Err(_) => return Poll::Ready(Err(std::io::ErrorKind::BrokenPipe.into())),
-            Ok(permit) => permit,
         };
 
-        let frame = buf[..num_bytes].to_vec();
-        permit.send(Event::Message(frame));
+        if buf.len() > MAX_FRAME_SIZE {
```
lexnv: nit: Should this be checked before `self.tx.poll_reserve(cx)`, such we don't waste resources and return earlier before fetching a permit?

haikoschol: Sorry, forgot to push the commit that reverted this check. Did this now and added a TODO comment with a link to #502.

dmitry-markin: Why do you need https://github.com/paritytech/litep2p/issues/502 at all?

haikoschol: > Why do you need #502 at all?

To make sure it won't be forgotten about. What do you propose instead?

dmitry-markin: Isn't it already working, as the passing test suggests?

haikoschol: > Isn't it already working, as the passing test suggests?

Sorry, I didn't read the test carefully and just assumed it didn't check that all the data made it to the other side of the channel between `Substream` and `SubstreamHandle`. But it does check that and it passes. I also just noticed that the test calls `AsyncWriteExt::write_all()` but the silent truncation you flagged is in `Substream::poll_write()`. So that's why the test passes I guess.

My knowledge of tokio async stuff is pretty limited so please just tell me what you want me to do to so we can merge this PR. 😅

src/transport/webrtc/substream.rs
lexnv: Ideally, we could handle the packets in this PR. As opposed to fixing one issue and introducing another one to tackle later on, we'll probably forget about it.

We just need a tiny loop here:

```rust
while written < buf.len() {

tx.poll_reserve(cx)
let frame_size = std::cmp::min(remaining, MAX_FRAME_SIZE);
send_item(frame[...len].into())

}
```

dmitry-markin: @lexnv @haikoschol It's fine for `AsyncWrite::poll_write` to write just part of the data, this is the way non-blocking IO works and the reason why it returns the number of bytes written.

No need to do anything further.

src/transport/webrtc/substream.rs
```diff
@@ -190,49 +193,46 @@ impl tokio::io::AsyncRead for Substream {
...
-        futures::pin_mut!(future);
-
-        let permit = match futures::ready!(future.poll(cx)) {
+        match futures::ready!(self.tx.poll_reserve(cx)) {
+            Ok(()) => {}
             Err(_) => return Poll::Ready(Err(std::io::ErrorKind::BrokenPipe.into())),
-            Ok(permit) => permit,
         };
 
+        // TODO: https://github.com/paritytech/litep2p/issues/502 Don't silently truncate frames.
```
dmitry-markin:
```suggestion
```

dmitry-markin: @haikoschol Please remove this TODO comment and we will merge the PR.

src/transport/webrtc/substream.rs
haikoschol: > Ideally, we could handle the packets in this PR. As opposed to fixing one issue and introducing another one to 
tackle later on, we'll probably forget about it.

We didn't introduce it, [it worked like this before](https://github.com/paritytech/litep2p/blob/785a89fec191e0239ac107e4a9028e5a42ca12b1/src/transport/webrtc/substream.rs#L203).

> We just need a tiny loop here:
> 
> ```rust
> while written < buf.len() {
> 
> tx.poll_reserve(cx)
> let frame_size = std::cmp::min(remaining, MAX_FRAME_SIZE);
> send_item(frame[...len].into())
> 
> }
> ```

Will this behave as intended when the channel is full? Are there any other edge cases? And wouldn't we just reimplement `AsyncWriteExt::write_all()`?

After Dmitry prompted me to check the test again I noticed that [it calls `write_all()` with 2 * `MAX_FRAME_SIZE` + 1](https://github.com/paritytech/litep2p/blob/785a89fec191e0239ac107e4a9028e5a42ca12b1/src/transport/webrtc/substream.rs#L264) and then checks that three messages come out the other end of the channel in `SubstreamHandle`. So it is possible already to send large amounts of data and have them split into multiple frames.

dmitry-markin: > This means we silently drop everything above `MAX_FRAME_SIZE`. It's better to drop the entire frame here and generate a warning. I.e., it' doesn't make sense sending partial frame if part of the data is going to be discarded, but we need to notify the developer of the transport misuse.

I apologize for causing confusion with my original comment.

haikoschol: > I apologize for causing confusion with my original comment.

No worries, I've removed the TODO comment.

lexnv:
Nice job here everyone! Good to have this merged 🙏

PR #495, opened 2025-12-04
transports: Improves the robustness and success rate of connection dialing

lexnv:
This PR improves the robustness and success rate of connection dialing by grouping peer negotiation (multistream-select) into the "open" (socket-to-socket connectivity) stages of the TCP and Websocket transports.
This avoids premature cancellation of potentially successful dials.

The litep2p establishes connection in two sequential stages:
- Step 1: socket to socket connection (raw TCP connection)
- Step 2: Peer negotiation (multistream-select protocol)

Previously, the transport manager would cancel all pending dials immediately after the first successful completion of Step 1.

Timeline:
- T0: Transport Manager dials peer on TCP Address A and Websocket Address B
- T1: Address A established raw socket connection (Step 1). The Transport Manager cancels immediately the dial to Address B
- T2: Address A proceeds to the peer negotiation (Step 2).
- T3: Negotiation on address A fails 

Because at T1 the manager cancels all other attempts, the dialing is reported as failed without trying to dial more than one address. 

To improve the robustness of the dialing process, Step 2 is merged with Step 1.

### Implementation Details

Any installed protocol (TCP / Websocket) must comply to the following interface:

https://github.com/paritytech/litep2p/blob/f1b4e0b9ac60a5cc1a3cf1fde6aa1defeb2168b7/src/transport/mod.rs#L192-L219

TCP/Websocket protocols now implement:

- `fn open` integrates step 1 and step 2 (with peer negotiation)
- TCPAddress A is reported to the manager only after the negotiation phase
- `fn negotiate` forwards the already negotiated connection to the manager immediately (no-op since it was already handled)
- WebSocket Address B is correctly dropped since we have a connection on the TCP socket

Closes: https://github.com/paritytech/litep2p/issues/232

### DIFF
--- a/src/transport/tcp/mod.rs
+++ b/src/transport/tcp/mod.rs
@@ -119,7 +119,7 @@ pub(crate) struct TcpTransport {
     pending_raw_connections: FuturesStream<BoxFuture<'static, RawConnectionResult>>,
 
     /// Opened raw connection, waiting for approval/rejection from `TransportManager`.
-    opened_raw: HashMap<ConnectionId, NegotiatedConnection>,
+    opened: HashMap<ConnectionId, NegotiatedConnection>,
 
     /// Cancel raw connections futures.
     ///
@@ -298,7 +298,7 @@ impl TransportBuilder for TcpTransport {
                 config,
                 context,
                 dial_addresses,
-                opened_raw: HashMap::new(),
+                opened: HashMap::new(),
                 pending_open: HashMap::new(),
                 pending_dials: HashMap::new(),
                 pending_inbound_connections: HashMap::new(),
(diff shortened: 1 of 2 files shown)


src/transport/tcp/mod.rs
```diff
@@ -496,52 +513,12 @@ impl Transport for TcpTransport {
     }
 
     fn negotiate(&mut self, connection_id: ConnectionId) -> crate::Result<()> {
```
dmitry-markin, line 515: May be we can delete `negotiate` fn from the trait completely, highlighting the fact that negotiation should happen as part of `open()`?

And likely rename `self.opened_raw` -> `self.opened`.

PR #501, opened 2025-12-12
Add Bitswap client

dmitry-markin:
Implement Bitswap client. This is needed to benchmark data retrieval from transaction storage.

Additionally, outbound substreams are now reused for subsequent messages (there is a limit of one outbound substream per peer).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.12.3] - 2025-12-16
+
+This release improves the robustness of the multistream-select negotiation over WebRTC transport and fixes inbound bandwidth metering on substreams. It also enhances the dialing success rate by improving the transport dialing logic. Additionally, it re-exports CID's multihash to facilitate the construction of CID V1.
+
+### Changed
+
+- transports: Improves the robustness and success rate of connection dialing  ([#495](https://github.com/paritytech/litep2p/pull/495))
+- types: Re-export cid's multihash to construct CID V1  ([#491](https://github.com/paritytech/litep2p/pull/491))
+
+### Fixed
+
+- fix: multistream-select negotiation on outbound substream over webrtc  ([#465](https://github.com/paritytech/litep2p/pull/465))
+- substream: Fix inbound bandwidth metering  ([#499](https://github.com/paritytech/litep2p/pull/499))
 
 ## [0.12.2] - 2025-11-28
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1027,15 +1027,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fxhash"
-version = "0.2.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c31b6d751ae2c7f11320402d34e41349dd1016f8d5d45e48c4312bc8625af50c"
-dependencies = [
- "byteorder",
-]
-
-[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1855,7 +1846,7 @@ checksum = "6373607a59f0be73a39b6fe456b8192fcc3585f602af20751600e974dd455e77"
 
 [[package]]
 name = "litep2p"
-version = "0.12.2"
+version = "0.12.3"
 dependencies = [
  "async-trait",
  "bs58 0.5.1",
@@ -2806,7 +2797,7 @@ dependencies = [
  "pin-project-lite 0.2.16",
  "quinn-proto",
  "quinn-udp",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "rustls 0.20.9",
  "thiserror 1.0.69",
  "tokio",
@@ -2823,7 +2814,7 @@ dependencies = [
  "bytes",
  "rand 0.8.5",
  "ring 0.16.20",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "rustls 0.20.9",
  "slab",
  "thiserror 1.0.69",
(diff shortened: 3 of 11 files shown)


src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -91,6 +97,33 @@ impl Prefix {
         res.extend_from_slice(multihash_len);
         res
     }
+
+    /// Parse byte representation of prefix.
+    pub fn from_bytes(prefix_bytes: &[u8]) -> Option<Prefix> {
+        let (version, rest) = unsigned_varint::decode::u64(prefix_bytes).ok()?;
+        let (codec, rest) = unsigned_varint::decode::u64(rest).ok()?;
+        let (multihash_type, rest) = unsigned_varint::decode::u64(rest).ok()?;
+        let (multihash_len, _rest) = unsigned_varint::decode::u64(rest).ok()?;
```
lexnv: nit: this means we ignore the unconsumed bytes, would that be expected?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -147,92 +188,227 @@ impl Bitswap {
...
+                Ok(()) => {
+                    // Store the peer to open a substream once it is connected.
+                    self.pending_dials.insert(peer);
                 }
-                ResponseType::Presence { cid, presence } => {
-                    response.block_presences.push(schema::bitswap::BlockPresence {
-                        cid: cid.to_bytes(),
-                        r#type: presence as i32,
-                    });
+                Err(error) => {
```
lexnv, line 389: nit: We might have an obscure race that have seen in other components/kademlia:
- We are disconnected at T0 on like 351 while `self.service.open_substream(peer)`
- However we had a pending conntions initiated by some other protocol
- By the time we `self.service.dial(&peer)` the peer might have connected
- Maybe its worth for `Err(crate::Error::AlreadyConnected)` to rerequet `self.service.open_substream(peer)`

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -287,3 +469,70 @@ impl Bitswap {
...
+                    send_dont_have: false,
+                })
+                .collect(),
+            full: false,
+        }),
+        ..Default::default()
+    };
+
+    let message = request.encode_to_vec().into();
+    if let Ok(Ok(())) = tokio::time::timeout(WRITE_TIMEOUT, substream.send_framed(message)).await {
```
lexnv, line 530: nit: this looks a bit odd?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -147,92 +188,227 @@ impl Bitswap {
...
+            }
+        }
 
-        substream.close().await;
+        // Store pending actions for later and open substream if not requested already.
+        let pending_actions = self.pending_outbound.entry(peer).or_default();
+        let no_pending_substream = pending_actions.is_empty();
+        pending_actions.push(SubstreamAction::SendRequest(cids));
+
+        if no_pending_substream {
```
lexnv: dq: If the `send_request` above fails, then we drop the substream via `entry.remove()`
- However here we are deterining to open substreams only if we had prior actions in `pending_outbound`
- If we do not, we will not open a new substream and leave the current cids as pending

Should we alreays request a new substream on `entry.remove()` code paths?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -147,92 +188,227 @@ impl Bitswap {
...
+    /// Handle opened outbound substream.
     async fn on_outbound_substream(
         &mut self,
         peer: PeerId,
         substream_id: SubstreamId,
         mut substream: Substream,
     ) {
-        let Some(entries) = self.pending_outbound.remove(&substream_id) else {
+        let Some(actions) = self.pending_outbound.remove(&peer) else {
             tracing::warn!(target: LOG_TARGET, ?peer, ?substream_id, "pending outbound entry doesn't exist");
```
lexnv, line 308: dq: Should we check the `pending_dials` here? Thinking of cases where we might receive this before `on_connection_established` but looks unlikely

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -287,3 +469,70 @@ impl Bitswap {
         }
     }
 }
+
```
lexnv, line 510: dq: Unrelated to these changes, we have an `self.inbound.remove(&peer)` above. However, there's a high likelyhood that if the inbound is broken, the outboud is as well. Should cleanup the outbound and force a reconnection?

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: This is a good question. Do you think we should check that `_rest.len() == 0` ?

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: `send_framed()` returns `Result` and `tokio::time::timeout` wraps it in its own `Result`, so we need to check that both succeeded.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: Don't have a strong opinion here, but if prefix is extended in the future in a backward-compatible fashion by appending more fields, allowing `_rest.len() != 0` here is kind of "future-compatible".

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: Thanks, completely forgot about it! Will handle this case.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: The logic here is if there are pending actions, we already requested opening a substream (or dialing a peer then opening a substream), so we shouldn't request opening the second substream.

I will add a comment explaining this.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: There may be legitimate cases where inbound fails (i.e., some implementations can close the outbound after sending just one request, like litep2p did before this PR), but outbound is fine.

The current code will fail fast if writing to the outbound fails and request a reconnection, so it shouldn't be a problem anyway.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: `TransportManager` should deliver events in order, so this should not be possible.

PR #509, opened 2026-01-12
Avoid panics returning error instead

### LINKED ISSUE #350: webrtc: Improve robustness by avoiding panics
https://github.com/paritytech/litep2p/blob/e0b7118ea2ca3e9f9beb9605f649d7e9dcdd24c4/src/transport/webrtc/opening.rs#L413-L414

dimartiro:
# Description

Improve robustness by replacing `.expect()` calls with proper error handling in `opening.rs`.

Instead of panicking when `on_noise_channel_open()` or `on_noise_channel_data()` fail, the errors are now logged and the connection is gracefully closed by returning `WebRtcEvent::ConnectionClosed`

Closes #350
Closes https://github.com/ChainSafe/gossamer-parity/issues/77

### DIFF
--- a/src/protocol/libp2p/bitswap/config.rs
+++ b/src/protocol/libp2p/bitswap/config.rs
@@ -30,8 +30,9 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
 /// IPFS Bitswap protocol name as a string.
 pub const PROTOCOL_NAME: &str = "/ipfs/bitswap/1.2.0";
 
-/// Maximum Size for `/ipfs/bitswap/1.2.0` payloads.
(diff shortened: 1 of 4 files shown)


dimartiro:
> lgtm. @dimartiro where these the only places in the webrtc implementation that use `expect()`, `unwrap()`, etc? I understood the issue more general than just being about `opening.rs` and IIRC there are more places that need proper error handling. Having said that, having many small PRs is probably a good approach either way.

Thanks for the review!
So far, these are the ones annotated with the issue in the comments. I'm not too familiar with the code yet, but I'll take a look and check if there are more places where this change might apply

lexnv (approved):
Looks good to me! I would keep the issues open until we sort out all the places we might panic 🙏

PR #510, opened 2026-01-12
notification: Replace async block with poll_recv

lexnv:
The `Connection::poll_next` implementation needlessly created an `async block` just to drop it when returning `Poll::Pending`. Instead, this PR polls the async_rx and sync_rx receivers manually. Considering that all substrate implementations use `sync_rx`, this takes priority (instead of the previous `tokio::select!`, which polled fairly).

Discovered during investigation of:
- https://github.com/paritytech/polkadot-sdk/issues/10425

cc @paritytech/networking

### DIFF
--- a/src/transport/webrtc/opening.rs
+++ b/src/transport/webrtc/opening.rs
@@ -409,8 +409,15 @@ impl OpeningWebRtcConnection {
                             continue;
                         }
 
-                        // TODO: https://github.com/paritytech/litep2p/issues/350 no expect
-                        self.on_noise_channel_open().expect("to succeed");
+                        if let Err(error) = self.on_noise_channel_open() {
+                            tracing::debug!(
+                                target: LOG_TARGET,
+                                connection_id = ?self.connection_id,
+                                ?error,
+                                "noise channel open failed",
+                            );
+                            return WebRtcEvent::ConnectionClosed;
+                        }
                     }
                     Event::ChannelData(data) => {
                         tracing::trace!(
@@ -428,8 +435,18 @@ impl OpeningWebRtcConnection {
                             continue;
                         }
 
-                        // TODO: https://github.com/paritytech/litep2p/issues/350 no expect
-                        return self.on_noise_channel_data(data.data).expect("to succeed");
+                        match self.on_noise_channel_data(data.data) {
+                            Ok(event) => return event,
+                            Err(error) => {
+                                tracing::debug!(
+                                    target: LOG_TARGET,
+                                    connection_id = ?self.connection_id,
+                                    ?error,
+                                    "noise channel data handling failed",
+                                );
+                                return WebRtcEvent::ConnectionClosed;
+                            }
+                        }
                     }
                     Event::ChannelClose(channel_id) => {
                         tracing::debug!(target: LOG_TARGET, ?channel_id, "channel closed");


src/protocol/notification/connection.rs
```diff
@@ -203,24 +203,21 @@ impl Stream for Connection {
...
-                        tokio::select! {
-                            notification = this.async_rx.recv() => notification,
-                            notification = this.sync_rx.recv() => notification,
-                        }
-                    };
-                    futures::pin_mut!(future);
-
-                    match future.poll_unpin(cx) {
-                        Poll::Pending => None,
+                None => match this.async_rx.poll_recv(cx) {
```
skunert: Tokio select chooses randomly which future gets polled first, now async_rx is always polled first, potentially sync_rx never polled. Is fairness necessary here or this is intentional?

lexnv: Yep that's a fair point :D In substrate, we are using exclusively `synx_rx`. IIRC, there's no usage of `async_rx` atm. Will change the order, thanks

src/protocol/notification/connection.rs
```diff
@@ -203,24 +203,21 @@ impl Stream for Connection {
         loop {
             let notification = match this.next_notification.take() {
                 Some(notification) => Some(notification),
-                None => {
-                    let future = async {
-                        tokio::select! {
-                            notification = this.async_rx.recv() => notification,
```
bkchr, line 209: https://docs.rs/tokio/1.49.0/tokio/sync/mpsc/struct.Receiver.html#cancel-safety

But this is cancel safe. So, I don't get your argument? 

The wake up call should still be registered and the entire future be called when there was an event?

lexnv: Yep, the receivers should be cancel safe. The issue here is with the `let future = async { }`. The context waker is registered by the inner `recv` calls inside the temporary future. The future would later on be dropped if `future.poll_unpin` returns `Poll::Pending`. Then, when the `sync_rx` got a new notification, it would wake the waker corresponding to the dropped future, causing the poll_next to stall.

bkchr: > Then, when the `sync_rx` got a new notification, it would wake the waker corresponding to the dropped future, causing the poll_next to stall.

But the waker is for the entire task and not just the future. So, the waker just wakes up the entire task and not some particular future.

lexnv: I've dug a bit into tokio to figure this out, indeed I'm mistaken with the "stalled connection" because I assumed `recv` worked similarly to `reserve` (the initial issue we noticed in webrtc):
- The tokio's bounded [Receiver](https://github.com/tokio-rs/tokio/blob/240cc44da87ac30188e946ed1003b0e67f64aad8/tokio/src/sync/mpsc/bounded.rs#L106-L109) uses a waiting list of "context waker" via a wrapper over [Semaphore](https://github.com/tokio-rs/tokio/blob/240cc44da87ac30188e946ed1003b0e67f64aad8/tokio/src/sync/batch_semaphore.rs#L35-L41) implementation

- I assumed that `sync_rx.recv()` would call into the semaphore `acquire` or similar to place the context waker into the linked list (obtaining an [Acquire](https://github.com/tokio-rs/tokio/blob/240cc44da87ac30188e946ed1003b0e67f64aad8/tokio/src/sync/batch_semaphore.rs#L71-L76))
  - Because the `sync_rx.recv()` future would get dropped immediately, the waker would be removed on [Drop](https://github.com/tokio-rs/tokio/blob/240cc44da87ac30188e946ed1003b0e67f64aad8/tokio/src/sync/batch_semaphore.rs#L694-L702) from the linked list
  - When the notification is received, there would be no registered waker in the list
 

However, the semaphore is only used for capacity. When we call into recv, the Receiver stores the waker into a separate [variable](https://github.com/tokio-rs/tokio/blob/240cc44da87ac30188e946ed1003b0e67f64aad8/tokio/src/sync/mpsc/chan.rs#L56-L57):

```
    /// Receiver waker. Notified when a value is pushed into the channel.
    rx_waker: CachePadded<AtomicWaker>,


fn recv
   self.inner.rx_waker.register_by_ref(cx.waker());
```

So regardless if the temporary `future` gets dropped, we'll still wake the proper waker under the hood. This PR justs turns into a tiny optimization to not create and drop a dedicated `async block` :D

src/protocol/notification/connection.rs
dmitry-markin: Thinking out loud, if the mostly loaded one is `sync_rx`, we should poll first `async_rx`, so that we fall through to polling of `sync_rx` as well. In the opposite case we might end up polling only the `sync_rx` which is always loaded and starve `async_rx`.

May be we can keep `tokio::select!` that does the polling randomization internally?

lexnv: That makes sense, we can close this for now, no longer needed if we keep the `select!` 🙏

PR #511, opened 2026-01-14
substream: Decrement the bytes counter to avoid excessive flushing

lexnv:
The substream implementation utilises a `pending_out_bytes` counter to enforce backpressure via flushing once the buffer fills up to 64KiB.

- The pending out bytes counter is properly incremented when data is received
- However, the variable was never decremented when writing data into the underlying socket
- Once enough data is received, this led to the `poll_ready` to make a syscall to flush the socket at every call

This implementation decrements the counter when data is moved to the socket. This should optimize the `poll_ready` fn for long lived substreams. 

While at it, the `poll_ready` implementation calls into the `Sink::poll_flush` which also drains the pending frames.

### DIFF
--- a/src/substream/mod.rs
+++ b/src/substream/mod.rs
@@ -548,7 +548,9 @@ impl Stream for Substream {
                                 return Poll::Ready(None);
                             }
 
-                            if nread == payload_size {
+                            this.offset = this.offset.saturating_add(nread);
+
+                            if this.offset == payload_size {
                                 let mut payload = std::mem::replace(
                                     &mut this.read_buffer,
                                     BytesMut::zeroed(payload_size),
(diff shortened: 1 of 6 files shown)


src/substream/mod.rs
```diff
@@ -677,7 +677,15 @@ impl Sink<Bytes> for Substream {
...
-            return poll_flush!(&mut self.substream, cx).map_err(From::from);
+            // This attempts to empty 'pending_out_frames' into the socket.
+            match futures::Sink::poll_flush(self.as_mut(), cx) {
+                Poll::Ready(Ok(())) => {}
+                Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+                Poll::Pending => {
+                    // Still flushing. We cannot accept new data yet.
+                    return Poll::Pending;
+                }
+            }
```
dmitry-markin, line 688: How is this different from `poll_flush!`?

dmitry-markin (approved):

src/substream/mod.rs
lexnv: The `poll_flush!` should call into the socket / TCP under the hood via `AsyncWrite::poll_flush`.

By calling the `futures::Sink::poll_flush` we ensure our impl is always called and try to write any pending frames before calling `poll_flush`

Issue #476, opened 2025-11-19
webrtc: Gracefully close substreams


haikoschol:
The libp2p WebRTC transport uses [a protobuf message containing the payload and flags](https://github.com/libp2p/specs/blob/master/webrtc/README.md#multiplexing) for signaling (half-)closing and resets of substreams.

The litep2p implementation does not set these flags when a substream is closed.

Strictly speaking, to be fully spec compliant, closing a substream should perform the following steps:

1. Stop accepting new data to send out.
1. Ensure that all pending data has been sent (flush).
1. Ideally, in the last protobuf message used for 2., set the `FIN` flag. Otherwise send a protobuf message with an empty payload.
1. Wait for the remote to respond with a `FIN_ACK`.
1. Close the underlying WebRTC data channel ([`DirectApi::close_data_channel()`](https://docs.rs/str0m/0.11.1/str0m/change/struct.DirectApi.html#method.close_data_channel))
1. Drop all resources associated with the substream.

PR #513, opened 2026-01-14
webrtc: Support FIN/FIN_ACK handshake for substream shutdown

timwu20:
## Summary

Implements proper FIN/FIN_ACK handshake for WebRTC substream shutdown according to the libp2p WebRTC-direct specification, with timeout handling to prevent indefinite waiting.

Closes https://github.com/paritytech/litep2p/issues/476

### 1. **Proper FIN/FIN_ACK Handshake**
- `poll_shutdown` now waits for FIN_ACK acknowledgment before completing
- Added waker mechanism to wake shutdown task when FIN_ACK is received
- Ensures graceful shutdown with delivery guarantees

### 2. **Timeout Protection**
- Added 5-second timeout for FIN_ACK (matching go-libp2p's behavior)
- Prevents indefinite waiting if remote never sends FIN_ACK
- Automatically completes shutdown after timeout to avoid resource leaks

### 3. **Shutdown Requirements Fulfilled**
- ✅ Stops accepting new writes during shutdown (transitions through `Closing` state)
- ✅ Flushes pending data before sending FIN
- ✅ Sends FIN flag with empty payload
- ✅ Waits for FIN_ACK acknowledgment from remote
- ✅ Triggers data channel closure and resource cleanup

### 4. **Code Refactoring**
- Unified `Message` event variant to include optional flag parameter (instead of separate `MessageWithFlags` variant)
- Unified encode functions into single function with optional flag parameter
- Fixed flag handling bug: changed from bitwise operations to equality checks
- Renamed plural "flags" to singular "flag" throughout codebase for consistency with protobuf schema

## Implementation Details

The shutdown process now follows this state machine:
1. `Open` → `Closing` (blocks new writes)
2. Flush pending data
3. Send FIN flag
4. `Closing` → `FinSent`
5. Wait for FIN_ACK or timeout (5 seconds)
6. `FinSent` → `FinAcked`
7. Drop substream → channel closes → data channel cleanup

The timeout task is spawned immediately when entering `FinSent` state and wakes the shutdown future after 5 seconds if no FIN_ACK is received.

## Testing

- Added comprehensive tests for FIN/FIN_ACK handshake
- Added test for timeout behavior when FIN_ACK is not received

## Compatibility

- Follows [libp2p WebRTC specification](https://github.com/libp2p/specs/blob/master/webrtc/README.md)
- 5-second timeout matches go-libp2p's implementation

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -1297,8 +1297,12 @@ impl TransportManager {
                                 }
                             }
                         }
-                        TransportEvent::ConnectionOpened { connection_id, address } => {
+                        TransportEvent::ConnectionOpened { connection_id, address, errors } => {
                             self.opening_errors.remove(&connection_id);
+
+                            for (addr, error) in &errors {
+                                self.update_address_on_dial_failure(addr.clone(), error);
+                            }
 
                             if let Err(error) = self.on_connection_opened(transport, connection_id, address) {
                                 tracing::debug!(
--- a/src/transport/mod.rs
+++ b/src/transport/mod.rs
@@ -141,6 +141,9 @@ pub(crate) enum TransportEvent {
 
         /// Address that was dialed.
         address: Multiaddr,
+
+        /// Errors from unsuccessful dial attempts.
+        errors: Vec<(Multiaddr, DialError)>,
     },
 
     /// Connection closed to remote peer.
--- a/src/transport/quic/mod.rs
+++ b/src/transport/quic/mod.rs
@@ -78,6 +78,7 @@ enum RawConnectionResult {
         connection_id: ConnectionId,
         address: Multiaddr,
         stream: NegotiatedConnection,
+        errors: Vec<(Multiaddr, DialError)>,
     },
 
     /// All connection attempts failed.
@@ -437,6 +438,7 @@ impl Transport for QuicTransport {
                             connection_id,
                             address,
                             stream,
+                            errors,
                         },
                     Err(error) => {
                         tracing::debug!(
@@ -514,6 +516,7 @@ impl Stream for QuicTransport {
                     connection_id,
                     address,
                     stream,
+                    errors,
                 } => {
                     let Some(handle) = self.cancel_futures.remove(&connection_id) else {
                         tracing::warn!(
@@ -531,6 +534,7 @@ impl Stream for QuicTransport {
                         return Poll::Ready(Some(TransportEvent::ConnectionOpened {
                             connection_id,
                             address,
+                            errors,
                         }));
                     }
                 }
--- a/src/transport/tcp/mod.rs
+++ b/src/transport/tcp/mod.rs
@@ -605,6 +605,7 @@ impl Stream for TcpTransport {
                         return Poll::Ready(Some(TransportEvent::ConnectionOpened {
                             connection_id,
                             address,
+                            errors,
                         }));
                     }
                 }
--- a/src/transport/websocket/mod.rs
+++ b/src/transport/websocket/mod.rs
@@ -650,6 +650,7 @@ impl Stream for WebSocketTransport {
                         return Poll::Ready(Some(TransportEvent::ConnectionOpened {
                             connection_id,
                             address,
+                            errors,
                         }));
                     }
                 }
--- a/tests/connection/failed_addresses_on_success.rs
+++ b/tests/connection/failed_addresses_on_success.rs
@@ -1,0 +1,193 @@
+// Copyright 2024 litep2p developers
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+//! Tests for verifying that failed addresses are reported back on a successful connection.
+//!
+//! This addresses the issue where the TCP transport (and others) attempts to dial up to 8
+//! addresses simultaneously. When one connection succeeds while others fail, the errors
+//! should be propagated back to the transport manager so that failed addresses can have
+//! their scores updated.
+//!
+//! See: https://github.com/paritytech/litep2p/issues/526
+
+use litep2p::{
+    config::ConfigBuilder, crypto::ed25519::Keypair, protocol::libp2p::ping::Config as PingConfig,
+    transport::tcp::config::Config as TcpConfig, Litep2p, Litep2pEvent,
+};
+
+use futures::StreamExt;
+use multiaddr::{Multiaddr, Protocol};
+use multihash::Multihash;
+use std::time::Duration;
+use tokio::net::TcpListener;
+
+#[tokio::test]
+async fn failed_addresses_reported_on_successful_connection() {
+    let _ = tracing_subscriber::fmt()
+        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .try_init();
+
+    // Test for verifying that failed addresses are reported back on a successful connection.
+    // - https://github.com/paritytech/litep2p/issues/526
+    let (ping_config2, _ping_event_stream2) = PingConfig::default();
+    let config2 = ConfigBuilder::new()
+        .with_keypair(Keypair::generate())
+        .with_tcp(TcpConfig {
+            listen_addresses: vec!["/ip4/127.0.0.1/tcp/0".parse().unwrap()],
+            ..Default::default()
+        })
+        .with_libp2p_ping(ping_config2)
+        .build();
+
+    let mut litep2p2 = Litep2p::new(config2).unwrap();
+    let peer2 = *litep2p2.local_peer_id();
+    let valid_address = litep2p2.listen_addresses().next().unwrap().clone();
+
+    tracing::info!(?peer2, ?valid_address, "target litep2p created");
+
+    // Create 7 TCP listeners that will accept connections but delay (simulating slow addresses)
+    // These will cause the connection attempts to timeout or fail during negotiation
+    let mut failing_listeners = Vec::new();
+    let mut failing_addresses = Vec::new();
+
+    for i in 0..7 {
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+        let address = Multiaddr::empty()
+            .with(Protocol::from(addr.ip()))
+            .with(Protocol::Tcp(addr.port()))
+            .with(Protocol::P2p(
+                Multihash::from_bytes(&peer2.to_bytes()).unwrap(),
+            ));
+
+        tracing::info!(index = i, ?address, "created failing listener");
+
+        failing_addresses.push(address);
+        failing_listeners.push(listener);
+    }
+
+    // Spawn tasks to handle the failing listeners - accept connection then delay 5 seconds
+    // This simulates addresses that are reachable but slow to complete the handshake
+    for (i, listener) in failing_listeners.into_iter().enumerate() {
+        tokio::spawn(async move {
+            loop {
+                match listener.accept().await {
+                    Ok((socket, addr)) => {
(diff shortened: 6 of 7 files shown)


src/transport/webrtc/util.rs
```diff
@@ -29,35 +29,16 @@ pub struct WebRtcMessage {
     /// Payload.
     pub payload: Option<Vec<u8>>,
 
-    // Flags.
-    pub flags: Option<i32>,
+    /// Flag.
+    pub flag: Option<i32>,
```
lexnv: nit: Maybe we could enforce here `Option<Flag>`?

src/transport/webrtc/substream.rs
```diff
@@ -74,6 +91,12 @@ pub struct Substream {
 
     /// RX channel for receiving messages from `peer`.
     rx: Receiver<Event>,
+
+    /// Waker to notify when shutdown completes (FIN_ACK received).
+    shutdown_waker: Arc<Mutex<Option<Waker>>>,
```
lexnv: Could an AtomicWaker be used instead?

src/transport/webrtc/substream.rs
```diff
@@ -32,32 +32,49 @@ use tokio_util::sync::PollSender;
...
     Open,
 
     /// Remote is no longer interested in receiving anything.
     SendClosed,
+
+    /// Shutdown initiated, flushing pending data before sending FIN.
+    Closing,
+
+    /// We sent FIN, waiting for FIN_ACK.
+    FinSent,
```
lexnv, line 72: How do we handle receiving a FIN packet while already in the FinSent state? In other words, due to net delays remote has decided to send a `FIN` and we receive it immediately after sending our own packet

src/transport/webrtc/substream.rs
```diff
@@ -223,13 +317,86 @@ impl tokio::io::AsyncWrite for Substream {
...
+            Ok(()) => {
+                // Transition to FinSent after successfully sending FIN
+                // Record timestamp for timeout tracking
+                *self.state.lock() = State::FinSent;
+                self.fin_sent_at = Some(Instant::now());
+                *self.shutdown_waker.lock() = Some(cx.waker().clone());
+
+                // Spawn timeout task to wake us after FIN_ACK_TIMEOUT
+                let waker = cx.waker().clone();
+                tokio::spawn(async move {
```
lexnv: Could the spanw be avoided here? 

Would it be possible for `poll_shutdown` to be called w different waker instances?
- in this case we'll spawn a timeout which after 5s wakes the wrong waker
- the poll contract states we should preserve the latest waker received

Maybe change this to an optional pinned tokio::time::sleep? Then we'll wake the same function and wake the appropriate waker 

nit: maybe use pin_project to avoid the boxing? Or similar strategy to how PollSender manages mem with memcpy under the hood

src/transport/webrtc/util.rs
```diff
@@ -29,35 +29,16 @@ pub struct WebRtcMessage {
...
-        out_buf.into()
-    }
-
-    /// Encode WebRTC message with flags.
-    #[allow(unused)]
-    pub fn encode_with_flags(payload: Vec<u8>, flags: i32) -> Vec<u8> {
-        let protobuf_payload = schema::webrtc::Message {
-            message: (!payload.is_empty()).then_some(payload),
-            flag: Some(flags),
+            flag,
```
lexnv: The code below is a bit unoptimized:
- one allocation for vec::with_capacity (protobuf enc)
- one allocation for bytes mut for variant-prefix

Could we write something like:

```rust
    let proto_len = protobuf_payload.encoded_len();
    let varint_len = unsigned_varint::encode::usize_buffer().len()

    // single alloc
    let mut out_buf = Vec::with_capacity(varint_len + proto_len);

    let mut varint_buf = unsigned_varint::encode::usize_buffer();
    let varint_bytes = unsigned_varint::encode::usize(proto_len, &mut varint_buf);
    out_buf.extend_from_slice(varint_bytes);

    // Encode protobuf directly into final buffer
    protobuf_payload.encode(&mut out_buf).expect(..suffucient)
```
?

src/transport/webrtc/util.rs
```diff
@@ -84,7 +65,7 @@ impl WebRtcMessage {
         match schema::webrtc::Message::decode(result) {
             Ok(message) => Ok(Self {
                 payload: message.message,
-                flags: message.flag,
+                flag: message.flag,
```
lexnv: nit: BytesMut::from(payload) will copy the entire payload, instead could we:
- let (len, remaining) = unsigned_varint::decode::usize(payload) 
- then decode the mssage via let message = schema::webrtc::Message::decode(&remaining[..len])?

src/transport/webrtc/substream.rs
```diff
@@ -106,36 +134,75 @@ pub struct SubstreamHandle {
...
+                    // Wake up any task waiting on shutdown
+                    if let Some(waker) = self.shutdown_waker.lock().take() {
+                        waker.wake();
+                    }
+                }
+                return Ok(());
+            }
+
+            if flag == Flag::StopSending as i32 {
                 *self.state.lock() = State::SendClosed;
```
lexnv: Could we receive the flag `StopSending` in the following race-case?

- T0: poll_write is called with Open
- T1: poll_reserve(cx) returns Pending -- waker registered for channel capacity
- T2: remote sends us the STOP_SENDING flag

Because we are not wakeing up any waker here, the `poll_write` will block indefinetely until we get channel capacity

src/transport/webrtc/substream.rs
```diff
@@ -32,32 +32,49 @@ use tokio_util::sync::PollSender;
...
+#[derive(Debug, Clone, Copy)]
 enum State {
     /// Substream is fully open.
     Open,
 
     /// Remote is no longer interested in receiving anything.
     SendClosed,
+
+    /// Shutdown initiated, flushing pending data before sending FIN.
+    Closing,
```
lexnv, line 69: Should we couple any `State` transition with an associated waker? 

In other words, who is going to wake up the counter part of `poll_shutdown` or `poll_write` if we are transitioning the state from `on_message` or individual poll functions?

src/transport/webrtc/substream.rs
```diff
@@ -106,36 +134,75 @@ pub struct SubstreamHandle {
...
+                return Ok(());
+            }
+
+            if flag == Flag::StopSending as i32 {
                 *self.state.lock() = State::SendClosed;
+                return Ok(());
             }
 
-            if flags & 2 == Flag::ResetStream as i32 {
+            if flag == Flag::ResetStream as i32 {
```
lexnv: Similarly to the state message, we are not waking other poll_ functions to announce the state transition. Shouldn't we do that to avoid blocking / delys in processing?

src/transport/webrtc/substream.rs
```diff
@@ -32,32 +32,49 @@ use tokio_util::sync::PollSender;
...
 
 /// Substream stream.
+#[derive(Debug, Clone, Copy)]
 enum State {
     /// Substream is fully open.
     Open,
 
     /// Remote is no longer interested in receiving anything.
     SendClosed,
+
```
lexnv, line 67: We could also replace the state with an AtomicU8 and have the implementation look like:

```rust
struct SharedState {
   // lock-free / more efficient 
   // BITS 0 to 3 
   // 0: open 1 sendClosed 2: closing 3: FinSent (our states) 4: FinAck
    
   // BIT 4: remote recv closed via FIN recv
   // BIT 5: STOP_SENDING remote send closed
   state: AtomicU8, 
   
   // Waker for tasks waiting on state changes (ie poll write blocked on state, but poll shutdown has FIN_ACK) -- triggered via on_message for STOP_SENDING, FIN_ACK, RESET_STREAM etc
   state_waker: AtomicWaker,
}

File: src/transport/webrtc/util.rs (hunk 1 above)
Comment (timwu20) [reply src/transport/webrtc/util.rs]: Updated in https://github.com/paritytech/litep2p/pull/513/commits/c7f4ea8618fc8b9788e36001a719af914ad083a5

File: src/transport/webrtc/substream.rs (hunk 3 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs:72]: This is expected behaviour since we are waiting on remote to send FIN_ACK before closing our write side of the channel.  When we receive a FIN from the remote we send FIN_ACK back and close the read side of the channel.  It's just in practice that we don't support this "half-closed" read/write side of the transport since when we call `shutdown` on the substream, we are closing both sides.  So if we have sent the FIN flag to the remote, we called `shutdown` on the substream.

File: src/transport/webrtc/substream.rs (hunk 4 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs]: Given that I've changed it to use `AtomicWaker` in https://github.com/paritytech/litep2p/pull/513/commits/42de0ab6c944e1ca3b995602d0d98797e25ce101, which updates to the latest worker it should now work if `poll_shutdown` is called multiple times.  However, it's quite unlikely that the `poll_shutdown` would be called multiple times since the `Substream` is not polled in different contexts.  

I've avoided the spawn by using a `Option<Pin<Box<tokio::time::Sleep>>>` in https://github.com/paritytech/litep2p/pull/513/commits/2ec58e4c08cae1a7ee0f31eb15e2b1134e6a13ea, but I was unable to use `pin_project` because it would make `Substream` `!Unpin`, which is required in the rest of the codebase.  The tradeoff being we are doing a small heap allocation for the timeout when shutting down.

File: src/transport/webrtc/util.rs (hunk 5 above)
Comment (timwu20) [reply src/transport/webrtc/util.rs]: I made the changes in https://github.com/paritytech/litep2p/pull/513/commits/a4effe22d4ddaa81f93bb85e59fb9eb365a06485.  I also optimized the `decode` function as well.

File: src/transport/webrtc/util.rs (hunk 6 above)
Comment (timwu20) [reply src/transport/webrtc/util.rs]: I implemented the suggestion in a4effe2.

File: src/transport/webrtc/substream.rs (hunk 7 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs]: Good catch! This should be fixed now in https://github.com/paritytech/litep2p/pull/513/commits/73f76ece464d5133caf62e15265025b7ae2975e1.  I introduced a `write_waker` to `Substream` and `SubstreamHandle` to re-poll the waker in the event that a `STOP_SENDING` flag is received.

File: src/transport/webrtc/substream.rs (hunk 10 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs:67]: Given that state transitions are relatively rare in this protocol, the performance gain would not be that pronounced.  I also think it's easier to understand semantically using the enum. I think It's more of a nice-to-have optimization, rather than a must have.  

In terms of the single `state_waker` `AtomicWaker`, given that the `Substream` must be owned and polled by exactly one task, I think it could be merged, but I think it makes it easier to understand semantically about which waker we are registering.  I'd rather not introduce the burden of figuring out if we should register over an already registered `AtomicWorker` based on state transition.

Does that make sense @lexnv?

File: src/transport/webrtc/substream.rs (hunk 8 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs:69]: This should be addressed now in https://github.com/paritytech/litep2p/pull/513/commits/73f76ece464d5133caf62e15265025b7ae2975e1.

File: src/transport/webrtc/substream.rs (hunk 9 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs]: We are waking the `write_waker` if we receive this flag in https://github.com/paritytech/litep2p/pull/513/commits/73f76ece464d5133caf62e15265025b7ae2975e1.

File: src/transport/webrtc/substream.rs (hunk 2 above)
Comment (timwu20) [reply src/transport/webrtc/substream.rs]: Updated in https://github.com/paritytech/litep2p/pull/513/commits/42de0ab6c944e1ca3b995602d0d98797e25ce101.

File: src/transport/webrtc/substream.rs [hunk 11]
```diff
@@ -106,36 +145,112 @@ pub struct SubstreamHandle {
...
+                        payload: vec![],
+                        flag: Some(Flag::FinAck),
+                    }) {
+                        tracing::warn!(
+                            target: "litep2p::webrtc::substream",
+                            ?e,
+                            "failed to send FIN_ACK, remote will timeout"
+                        );
+                    }
+                    return Ok(());
```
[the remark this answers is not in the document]
timwu20: The code already returns early after detecting a duplicate FIN. The `swap` returns the previous value. If `read_closed` was already true, we return immediately at line 199 without sending FIN_ACK.

src/transport/webrtc/substream.rs
```diff
@@ -223,13 +376,105 @@ impl tokio::io::AsyncWrite for Substream {
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Result<(), std::io::Error>> {
+        // State machine for proper shutdown:
```
lexnv, line 374: In `poll_flush` we return immediately, IIUC we should send the FIN packet only after all pending data is delivered. Do we handle this gracefully or can deliver FIN before buffered data reaches the remote?

src/transport/webrtc/substream.rs
```diff
@@ -33,31 +33,51 @@ use std::{
...
 const MAX_FRAME_SIZE: usize = 16384;
 
+/// Timeout for waiting on FIN_ACK after sending FIN.
+/// Matches go-libp2p's 5 second stream close timeout.
+#[cfg(not(test))]
+const FIN_ACK_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Shorter timeout for tests.
+#[cfg(test)]
+const FIN_ACK_TIMEOUT: Duration = Duration::from_secs(2);
```
lexnv: We can remove this and rely on 5s as production

src/transport/webrtc/substream.rs
timwu20: Yes with the `poll_reserve` call on line 446, we won't send the FIN via `self.tx` unless the channel has space.  After the FIN flag is sent, we won't drop the connection until a FIN_ACK is received or the timeout expires.

src/transport/webrtc/substream.rs
timwu20: Updated in https://github.com/paritytech/litep2p/pull/513/commits/184971c9cd9df718e8c9b81bdf8b414eb32c3f0c.

Issue #514, opened 2026-01-16
[Bitswap] Handle big data


bkontur:
I am uploading cca 33 MiB file splitted to the 1MiB chunks, means 34 chunks stored separately.
Uploading one more file with DAG content for all chunks -> rootCID.
I can retrieve all the chunks one by one from IPFS (ipfs.get)
 

The alice.log looks good, we received requests for DAG rootCid and all 34 chunk cids.
All the cids were found and returned by BitswapHandle.

**But the problem is somwhere between Kubo and litep2p `substream.send_framed(message)`, the message in this case contains all the 34 chunks, so it has cca 32 MiB.**

Here: https://github.com/paritytech/litep2p/compare/bko-bitswap?expand=1 I did some debugging and fixing, but I am missing something.
I also read about Bitswap impls using pendingBytes for kind of backpreassure and limiting. But don't want to leave false clues.

# Reproduce 
```
# checkout this branch:
https://github.com/paritytech/polkadot-bulletin-chain/pull/184

cargo install just --locked
cd examples

# Start the all env (zombie, kubo, ...) with logs
ZOMBIENET_BINARY=~/local_bridge_testing/bin/zombienet just setup-services "/tmp/bulletin-tests-run-SHvwb/test" "bulletin-polkadot-runtime"

# Trigger the test
npm run papi:generate
node store_big_data.js

# Wait a little bit for storing 32 MiB
# fetch download -> http 504 error

⬇️ Downloading the full content (no chunking) by cid from url:  http://127.0.0.1:8080/ipfs/bafykbzaceagtohskeansk72cjlebkypzplg3ew5sjuzjarljcpfr4rzbr6zlo
❌ Error: Error: HTTP error 504

# Explore Kubo logs
tail -f /tmp/bulletin-tests-run-SHvwb/test/ipfs-node.log

# Explore node logs with some litep2p logging
tail -f /tmp/bulletin-tests-run-SHvwb/test/alice.log
```

## One attempt with chunking on the litep2p level

I added some debug and logs: https://github.com/paritytech/litep2p/compare/bko-bitswap?expand=1
Patched bulletin Cargo.toml
```
litep2p = "0.12.3"  # keep the same version

[patch.crates-io]
litep2p = { path = "../../cargo-remote-builds-litep2p/6886191744186279728" }
```

And with this chunking I can see s**ome cids** are received by Kubo but the stream is probably closed and no file is downloaded.

```
[...]

PR #516, opened 2026-01-16
bitswap: Split block responses into batches under 2 MiB

dmitry-markin:
Split blocks in a Bitswap response into batches under 2 MiB so the maximum substream message size is respected and we don't lose the blocks.

Closes https://github.com/paritytech/litep2p/issues/514.

Follow-ups:
- [x] Add tests
- [ ] Release litep2p with the fix & upgrade litep2p in polkadot-sdk

### DIFF
--- a/src/crypto/noise/mod.rs
+++ b/src/crypto/noise/mod.rs
@@ -797,7 +797,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
             Role::Dialer => {
                 // write initial message
                 let first_message = noise.first_message(Role::Dialer)?;
-                let _ = io.write(&first_message).await?;
+                io.write_all(&first_message).await?;
                 io.flush().await?;
 
                 // read back response which contains the remote peer id
@@ -812,7 +812,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
 
                 // send the final message which contains local peer id
                 let second_message = noise.second_message()?;
-                let _ = io.write(&second_message).await?;
+                io.write_all(&second_message).await?;
                 io.flush().await?;
 
                 payload
@@ -823,7 +823,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
 
                 // send local peer id.
                 let second_message = noise.second_message()?;
-                let _ = io.write(&second_message).await?;
+                io.write_all(&second_message).await?;
                 io.flush().await?;
 
                 // read remote's second message which contains their peer id


src/protocol/libp2p/bitswap/config.rs
```diff
@@ -30,9 +30,13 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
...
-/// Maximum Size for `/ipfs/bitswap/1.2.0` substream payload. Note this is bigger than 2 MiB max
-/// block size to account for protobuf message overhead.
-const MAX_PAYLOAD_SIZE: usize = 2_100_000;
+/// Maximum size for `/ipfs/bitswap/1.2.0` substream message. Includes enough room for protobuf
+/// overhead. Enforced on the transport level.
+pub const MAX_MESSAGE_SIZE: usize = 4_194_304;
+
+/// Maximum batch size of all blocks in a single Bitswap message combined. Enforced on the
+/// application protocol level.
+pub const MAX_BATCH_SIZE: usize = 2_097_152;
```
lexnv: nit: `2 * 1024 * 1024`

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -535,16 +536,102 @@ async fn send_request(substream: &mut Substream, cids: Vec<(Cid, WantType)>) ->
...
+            }
+        } else {
+            // This should never happen in practice, but log a warning if the presence message
+            // exceeded [`config::MAX_MESSAGE_SIZE`].
+            tracing::warn!(
+                target: LOG_TARGET,
+                size = message.len(),
+                max_size = config::MAX_MESSAGE_SIZE,
+                "outgoing Bitswap presence message exceeded max size",
+            );
```
lexnv, line 565: nit: this is similar to the block below, maybe we can group them to avoid duplications?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -535,16 +536,102 @@ async fn send_request(substream: &mut Substream, cids: Vec<(Cid, WantType)>) ->
...
+    while let Some(batch) = extract_next_batch(&mut blocks, config::MAX_BATCH_SIZE) {
+        if let Some((message, block_count)) = blocks_message(batch) {
+            if message.len() <= config::MAX_MESSAGE_SIZE {
+                tracing::trace!(
+                    target: LOG_TARGET,
+                    block_count,
+                    "sending Bitswap blocks message",
+                );
+                match tokio::time::timeout(WRITE_TIMEOUT, substream.send_framed(message)).await {
+                    Err(_) => return Err(Error::Timeout),
```
lexnv, line 587: dq: This will drop any remaining block because we consider the connection as unhealthy?

However, this might be expected under heavy load that the protocol handle will not be able to keep up with messages. Maybe we can propagate this to higher levels or is the timeout error sufficient to retry later on?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -535,16 +536,102 @@ async fn send_request(substream: &mut Substream, cids: Vec<(Cid, WantType)>) ->
...
+                    Ok(Err(e)) => return Err(Error::SubstreamError(e)),
+                    Ok(Ok(())) => {}
+                }
+            } else {
+                // This should never happen in practice, but log a warning if the blocks message
+                // exceeded [`config::MAX_MESSAGE_SIZE`].
+                tracing::warn!(
+                    target: LOG_TARGET,
+                    size = message.len(),
```
lexnv, line 596: nit: Could the CID be useful here for debugging?

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: This is a good question I don't have the right an answer to. Indeed the entire response will be dropped. I am inclined to think that if the protocol handle is not able to keep up with messages, we shouldn't retry sending. Instead it will be up to the Bitswap client code to handle the timeout and repeat the query, may be querying another peer.

It would be good to investigate how Kubo handles this and if this will automatically work.

dmitry-markin: https://github.com/paritytech/litep2p/issues/520

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: May be not here, but in `extract_next_batch` we should log it of course. Missed it.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: I can't think of a good way of de-duplicating this, the data and the log messages are different.

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -30,9 +30,13 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
 /// IPFS Bitswap protocol name as a string.
 pub const PROTOCOL_NAME: &str = "/ipfs/bitswap/1.2.0";
 
-/// Maximum Size for `/ipfs/bitswap/1.2.0` substream payload. Note this is bigger than 2 MiB max
-/// block size to account for protobuf message overhead.
-const MAX_PAYLOAD_SIZE: usize = 2_100_000;
+/// Maximum size for `/ipfs/bitswap/1.2.0` substream message. Includes enough room for protobuf
+/// overhead. Enforced on the transport level.
+pub const MAX_MESSAGE_SIZE: usize = 4_194_304;
```
bkontur: nit: the same here 4 * 1024 * 1024

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -553,24 +640,179 @@ async fn send_response(substream: &mut Substream, entries: Vec<ResponseType>) ->
...
+    max_size: usize,
+) -> Option<Drain<'a, (Cid, Vec<u8>)>> {
+    // Get rid of oversized blocks to not stall the processing by not being able to queue them.
+    loop {
+        if let Some(block) = blocks.front() {
+            if block.1.len() > max_size {
+                tracing::warn!(
+                    target: LOG_TARGET,
+                    size = block.1.len(),
+                    max_size,
```
bkontur: max_size is not that clear, I suggest to rename it to `max_block_size`

```suggestion
    max_block_size: usize,
) -> Option<Drain<'a, (Cid, Vec<u8>)>> {
    // Get rid of oversized blocks to not stall the processing by not being able to queue them.
    loop {
        if let Some(block) = blocks.front() {
            if block.1.len() > max_block_size {
                tracing::warn!(
                    target: LOG_TARGET,
                    size = block.1.len(),
                    max_block_size,
```

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -553,24 +640,180 @@ async fn send_response(substream: &mut Substream, entries: Vec<ResponseType>) ->
...
-                });
+                }
+            })
+            .collect(),
+        ..Default::default()
+    };
+
+    let count = message.payload.len();
+
+    (count > 0).then(|| (message.encode_to_vec().into(), count))
```
bkontur, line 654: How does Bitswap work, shouldn't we send also empty Message? Doesn't it cause disconnect or something?

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -535,16 +536,102 @@ async fn send_request(substream: &mut Substream, cids: Vec<(Cid, WantType)>) ->
...
+        .into_iter()
+        .filter_map(|entry| match entry {
+            ResponseType::Block { cid, block } => Some((cid, block)),
+            ResponseType::Presence { .. } => None,
+        })
+        .collect::<VecDeque<_>>();
+
+    while let Some(batch) = extract_next_batch(&mut blocks, config::MAX_BATCH_SIZE) {
+        if let Some((message, block_count)) = blocks_message(batch) {
+            if message.len() <= config::MAX_MESSAGE_SIZE {
```
bkontur, line 580: Basically, this IF does not make sense, if we ensure that MAX_BATCH_SIZE < MAX_MESSAGE_SIZE

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -553,24 +640,180 @@ async fn send_response(substream: &mut Substream, entries: Vec<ResponseType>) ->
...
+    let count = message.payload.len();
+
+    (count > 0).then(|| (message.encode_to_vec().into(), count))
+}
+
+/// Extract a batch of blocks of no more than `max_size` from `blocks`.
+/// Returns `None` if no more blocks are left.
+fn extract_next_batch<'a>(
+    blocks: &'a mut VecDeque<(Cid, Vec<u8>)>,
+    max_size: usize,
```
bkontur: ok, not max_block_size, but `max_batch_size`?

```suggestion
    max_batch_size: usize,
```

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -30,9 +30,13 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
 /// IPFS Bitswap protocol name as a string.
 pub const PROTOCOL_NAME: &str = "/ipfs/bitswap/1.2.0";
 
-/// Maximum Size for `/ipfs/bitswap/1.2.0` substream payload. Note this is bigger than 2 MiB max
-/// block size to account for protobuf message overhead.
-const MAX_PAYLOAD_SIZE: usize = 2_100_000;
```
bkontur, line 35: @rosarp probably this constant was the reason, why 2 MiB didn't work :)

src/protocol/libp2p/bitswap/config.rs
```diff
@@ -30,9 +30,13 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
...
+/// Maximum size for `/ipfs/bitswap/1.2.0` substream message. Includes enough room for protobuf
+/// overhead. Enforced on the transport level.
+pub const MAX_MESSAGE_SIZE: usize = 4 * 1024 * 1024;
+
+/// Maximum batch size of all blocks in a single Bitswap message combined. Enforced on the
+/// application protocol level.
+pub const MAX_BATCH_SIZE: usize = 2 * 1024 * 1024;
```
bkontur, line 39: I am just thinking, what about adding both to the Config:
```
pub struct Config {
    /// Protocol name.
    pub(crate) protocol: ProtocolName,
    /// Protocol codec.
    pub(crate) codec: ProtocolCodec,
    /// TX channel for sending events to the user protocol.
    pub(super) event_tx: Sender<BitswapEvent>,
    /// RX channel for receiving commands from the user.
    pub(super) cmd_rx: Receiver<BitswapCommand>,
    
    pub max_message_size: usize,
    pub max_batch_size: usize,
}
```
And use those constants for default value?

src/protocol/libp2p/bitswap/mod.rs
bkontur: well, I agree with Alex, this could be maybe de-duplicated. In both cases, you are building `let message = schema::bitswap::Message`, so when building it we could check MAX_MESSAGE_SIZE or MAX_BATCH_SIZE.

Basically, we could reuse `extract_next_batch(&mut blocks, config::MAX_BATCH_SIZE)` for both - maybe with closure which would return length (for presence - 2 bytes?, for block real length)

src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -553,24 +640,180 @@ async fn send_response(substream: &mut Substream, entries: Vec<ResponseType>) ->
...
-        Err(_) => Err(Error::Timeout),
-        Ok(Err(e)) => Err(Error::SubstreamError(e)),
-        Ok(Ok(())) => Ok(()),
+    // Determine how many blocks we can batch. Note that we can always batch at least one
+    // block due to check above.
+    let mut total_size = 0;
+    let mut block_count = 0;
+
+    for b in blocks.iter() {
+        let next_block_size = b.1.len();
```
bkontur, line 689: Does this batching also count with `prefix` and struct bytes overhead (2-3 bytes)?
```
                schema::bitswap::Block {
                    prefix,
                    data: block,
                });
```

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: If we passed an empty iterator (e.g., no block presence at all), it doesn't make sense to send empty message.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: It is highly unlikely in practice, but due to protobuf overhead message size can be > `MAX_BATCH_SIZE`.

src/protocol/libp2p/bitswap/config.rs
dmitry-markin: In the past it was exactly 2 MiB, but due to prefix and protobuf overhead the message didn't fit.

src/protocol/libp2p/bitswap/config.rs
dmitry-markin: 2 MiB are defined in the protocol spec, so i don't think we should add the possibility to "tune" it.

src/protocol/libp2p/bitswap/mod.rs
dmitry-markin: No, it's only about total block size. As per spec, 2 MiB blocks must go through, so the actual message size will be higher.

PR #518, opened 2026-01-19
crypto/noise: Improve stability of websockets by fixing AsyncWrite implementation

lexnv:
The PR fixes a `AsyncWrite::poll_write` implementation of the crypto/noise sockets that causes panics in rust-yamux and leads to unnecessary connection closure and instability:
- T0: poll_write is called with buffer of len 512 bytes
  - the implementation encrypt data and buffers it
  - because the io socket is not ready, poll pending is returned.
- T1: tokio_tungstenite (or other component) decides that a new payload must be sent (usually a PONG frame) of len 12 bytes
  - because the inner buffers contain the previous message, upon flushing the size of the older message is returned (ie 512)
  - rust-yamux uses the 512 bytes to index a buffer of 12 bytes (as expected by the second poll_write with buffer 12)
  - This caused frequent panics on the websocket implementation, which is currently addressed as abrupt connection termination
  
This PR fixes the `AsyncWrite` contract violation by effectively decoupling the encryption from the writing steps.
- the inner buffers are drained as much as possible (until the socket returns poll pending)
- The provided message is encrypted into the inner buffer if it has capacity and the number of bytes is returned immediately
- a subsequent poll_write or poll_flush or poll_close will propagate the encrypted buffered data to the underlying socket

Previous fixes:
- https://github.com/libp2p/rust-yamux/pull/202
- https://github.com/libp2p/rust-yamux/pull/211

The fixes are still needed since the tokio-tungstenite (websocket crate) was not properly scoped and may exhibit similar behavior. 

cc @paritytech/networking

### DIFF
--- a/src/crypto/noise/mod.rs
+++ b/src/crypto/noise/mod.rs
@@ -852,7 +852,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
             Role::Dialer => {
                 // write initial message
                 let first_message = noise.first_message(Role::Dialer)?;
-                let _ = io.write(&first_message).await?;
+                io.write_all(&first_message).await?;
                 io.flush().await?;
 
                 // read back response which contains the remote peer id
@@ -867,7 +867,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
 
                 // send the final message which contains local peer id
                 let second_message = noise.second_message()?;
-                let _ = io.write(&second_message).await?;
+                io.write_all(&second_message).await?;
                 io.flush().await?;
 
                 payload
@@ -878,7 +878,7 @@ pub async fn handshake<S: AsyncRead + AsyncWrite + Unpin>(
 
                 // send local peer id.
                 let second_message = noise.second_message()?;
-                let _ = io.write(&second_message).await?;
+                io.write_all(&second_message).await?;
                 io.flush().await?;
 
                 // read remote's second message which contains their peer id
--- a/src/protocol/libp2p/bitswap/config.rs
+++ b/src/protocol/libp2p/bitswap/config.rs
@@ -30,9 +30,13 @@ use tokio::sync::mpsc::{channel, Receiver, Sender};
 /// IPFS Bitswap protocol name as a string.
 pub const PROTOCOL_NAME: &str = "/ipfs/bitswap/1.2.0";
 
-/// Maximum Size for `/ipfs/bitswap/1.2.0` substream payload. Note this is bigger than 2 MiB max
-/// block size to account for protobuf message overhead.
-const MAX_PAYLOAD_SIZE: usize = 2_100_000;
+/// Maximum size for `/ipfs/bitswap/1.2.0` substream message. Includes enough room for protobuf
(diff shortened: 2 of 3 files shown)


src/crypto/noise/mod.rs
```diff
@@ -651,90 +646,152 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncWrite for NoiseSocket<S> {
...
+
+        // Flush underlying socket
+        Pin::new(&mut this.io).poll_flush(cx)
     }
 
     fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        // Ensure buffer is flushed before closing
+        if let Poll::Pending = self.as_mut().poll_flush(cx) {
+            return Poll::Pending;
+        }
```
dmitry-markin: nit:

```suggestion
        futures::ready!(self.as_mut().poll_flush(cx));
```

src/crypto/noise/mod.rs
```diff
@@ -651,90 +646,152 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncWrite for NoiseSocket<S> {
...
-                                    continue;
-                                }
-                            }
+        // If we are still in Writing state, we must check if we have space.
+        let mut buffer_offset = match this.write_state {
+            WriteState::Idle => 0,
+            WriteState::Writing { encrypted_len, .. } => encrypted_len,
+        };
+        // If buffer is too full to fit even one frame overhead, we must wait.
+        if buffer_offset + MAX_NOISE_MSG_LEN + 2 > this.encrypt_buffer.len() {
```
dmitry-markin: Why is the check below `if buffer_offset + chunk.len() + overhead > this.encrypt_buffer.len() {` not enough? This `MAX_NOISE_MSG_LEN` seem to be a lot of room — do we really need it?

src/crypto/noise/mod.rs
```diff
@@ -651,90 +646,152 @@ impl<S: AsyncRead + AsyncWrite + Unpin> AsyncWrite for NoiseSocket<S> {
...
+            // This can happen once we have pending data and no space to buffer new data.
+            // Therefore, a call to poll_write returned Pending in the first step. This
+            // just forwards the Pending to the caller until we have space again.
+            return Poll::Pending;
```
dmitry-markin: It looks like we haven't registered the waker to wake up the task once we have space in the buffer (and `poll_write` above might not wake as as well if it returned `Ready`). May be we don't need this `if` block at all to not return `Poll::Pending`?

dmitry-markin (approved):
Cool finding and good job here!

src/crypto/noise/mod.rs
lexnv: It makes sense, especially since we do the check below, have removed this 🙏

src/crypto/noise/mod.rs
lexnv: Yep, if the `poll_write` returns Poll::Ready, then we must rely on the fact that:
- one chunk+overhead has sufficient space to fill into the `encrypt_buffer`

PR #519, opened 2026-01-19
crypto/noise: Fix connection negotiation logic on partial writes

lexnv:
This PR replaces the `io.write` with `io.write_all` calls to ensure the negotiation messages are fully propagated.

- There was a bug in the `fn handshake` implementation, which might propagate only a part of the provided negotiation message (ie on partial writes the `write` call would return immediately)
- This would then be rejected by the other side, causing unnecessary negotiation errors while opening the connection
- The downstream effect of this is that higher level protocols (like notification protocols) would backoff the peer for ~5s.


dmitry-markin (approved):
Nice catch!

PR #525, opened 2026-01-27
transport: Make accept async to close the gap on service races

lexnv:
This PR ensures that protocols know about an incoming connection established before the litep2p users.

There is a race condition in the litep2p that is caused by the following:
- connection established are reported immediately to the users (the ones that poll the litep2p object to advacne the network)
- then the notification / request response protocols are informed about the connection

Under heavy load this can produce a gap, where the higher levels know about the connection, but protocols cannot open a new substream with the said connection.

 To close the gap, this PR modified the `accept` function to first report the connection to the protocols before the higher levels are inforemd.

### DIFF
--- a/src/transport/manager/mod.rs
+++ b/src/transport/manager/mod.rs
@@ -32,7 +32,7 @@ use crate::{
             peer_state::{ConnectionRecord, PeerState, StateDialResult},
             types::PeerContext,
         },
-        Endpoint, Transport, TransportEvent, MAX_PARALLEL_DIALS,
+        Endpoint, Transport, TransportEvent,
     },
     types::{protocol::ProtocolName, ConnectionId},
     BandwidthSink, PeerId,
@@ -205,9 +205,6 @@ pub struct TransportManager {
     /// Bandwidth sink.
     bandwidth_sink: BandwidthSink,
 
-    /// Maximum parallel dial attempts per peer.
-    max_parallel_dials: usize,
-
     /// Installed protocols.
     protocols: HashMap<ProtocolName, ProtocolContext>,
 
@@ -268,9 +265,6 @@ pub struct TransportManagerBuilder {
     /// Bandwidth sink.
     bandwidth_sink: Option<BandwidthSink>,
 
-    /// Maximum parallel dial attempts per peer.
-    max_parallel_dials: usize,
-
     /// Connection limits config.
     connection_limits_config: limits::ConnectionLimitsConfig,
 }
@@ -288,7 +282,6 @@ impl TransportManagerBuilder {
             keypair: None,
             supported_transports: HashSet::new(),
             bandwidth_sink: None,
-            max_parallel_dials: MAX_PARALLEL_DIALS,
             connection_limits_config: limits::ConnectionLimitsConfig::default(),
         }
     }
@@ -311,12 +304,6 @@ impl TransportManagerBuilder {
     /// Set the bandwidth sink
     pub fn with_bandwidth_sink(mut self, bandwidth_sink: BandwidthSink) -> Self {
         self.bandwidth_sink = Some(bandwidth_sink);
-        self
-    }
-
-    /// Set the maximum parallel dials per peer
-    pub fn with_max_parallel_dials(mut self, max_parrallel_dials: usize) -> Self {
-        self.max_parallel_dials = max_parrallel_dials;
         self
     }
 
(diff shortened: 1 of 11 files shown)


src/transport/webrtc/connection.rs
```diff
@@ -667,19 +667,25 @@ impl WebRtcConnection {
...
+    //     );
+
+    //     let _ = self
+    //         .protocol_set
+    //         .report_connection_established(self.peer, self.endpoint.clone())
+    //         .await;
+
+    //     self.run_event_loop().await;
+    // }
+
```
bkchr:
```suggestion
```

src/transport/manager/mod.rs
```diff
@@ -1090,6 +1096,32 @@ impl TransportManager {
     /// Poll next event from [`crate::transport::manager::TransportManager`].
     pub async fn next(&mut self) -> Option<TransportEvent> {
         loop {
+            // First, check if we have a pending accept future to poll
+            if let Some((peer, endpoint, mut future)) = self.pending_accept.take() {
```
bkchr: So we "stop" here everything when one connection is waiting for its accept to finish?

lexnv: Yep, this is not ok, especially under load. The PR is entirely vibed as PoC for @tdimitrov 🙏 

I'll need to rewrite this with some FuturesUnordered and see ask Tsvetomir for another testing round 🙏

src/transport/manager/mod.rs
```diff
@@ -252,6 +252,10 @@ pub struct TransportManager {
 
     /// Opening connections errors.
     opening_errors: HashMap<ConnectionId, Vec<(Multiaddr, DialError)>>,
+
+    /// Pending accept futures with associated connection information.
+    /// The manager will emit a `TransportEvent::ConnectionEstablished` when complete.
```
bkchr:
```suggestion
```

src/transport/manager/mod.rs
```diff
@@ -1091,6 +1096,33 @@ impl TransportManager {
     pub async fn next(&mut self) -> Option<TransportEvent> {
         loop {
             tokio::select! {
+                result = self.pending_accept.next(), if !self.pending_accept.is_empty() => {
+                    let Some((peer, endpoint, result)) = result else {
+                        continue;
+                    };
```
bkchr:
```suggestion
                (peer, endpoint, result) = self.pending_accept.select_next_some(), if !self.pending_accept.is_empty() => {
```

bkchr (approved):

src/transport/manager/mod.rs
```diff
@@ -1270,16 +1297,30 @@ impl TransportManager {
...
+                                        Ok(future) => {
+                                            // A ConnectionEstablished is propagated to the user once
+                                            // all protocols have been notified.
+                                            self.pending_accept.push(Box::pin(async move {
+                                                let result = future.await;
+                                                (peer, endpoint, result)
+                                            }));
+                                        }
+                                        Err(error) => {
+                                            tracing::debug!(
```
skunert, line 1312: Don't we need to clean up some data in this case? Accept will not run at all when we arrive here, but we already modified some state in `on_connection_established`

src/transport/manager/mod.rs
```diff
@@ -1091,6 +1095,29 @@ impl TransportManager {
...
+                                target: LOG_TARGET,
+                                ?peer,
+                                ?endpoint,
+                                "connection accepted and protocols notified",
+                            );
+
+                            return Some(TransportEvent::ConnectionEstablished { peer, endpoint });
+                        }
+                        Err(error) => {
+                            tracing::error!(
```
skunert, line 1102: Same as the other comment, if we fail here we need to clean up some data like this here: https://github.com/paritytech/litep2p/blob/fe1a44821c35d2db2391fbd6743bdf2f9478e9d4/src/transport/manager/mod.rs#L899-L899

src/transport/quic/mod.rs
```diff
@@ -307,35 +307,48 @@ impl Transport for QuicTransport {
...
-            )
-            .start()
-            .await;
-        }));
+        let peer = connection.peer;
+        let endpoint_clone = endpoint.clone();
+
+        Ok(Box::pin(async move {
+            // First, notify all protocols about the connection establishment
+            protocol_set.report_connection_established(peer, endpoint_clone).await?;
```
dmitry-markin, line 334: `report_connection_established` internally sends an event to protocols. Can it be a problem if the protocol doesn't read the event fast enough?

May be we need to modify `InnerTransportEvent::ConnectionEstablished` to include oneshot channel and collect ACKs from the protocols?

lexnv: Yep this could be a great followup. Would like to keep this separate as we'll need to revalidate if this adds any delays to connection / deadlocks if one single protocol is under pressure during high loads. 🙏

PR #542, opened 2026-02-22
support for dialing DNS addresses with QUIC

### LINKED ISSUE #406: Add support for dialing DNS addresses with QUIC
Currently, QUIC only supports dialing IPv4 & IPv6 addresses, and not DNS addresses like TCP & WebSocket transports. We should add support for dialing DNS addresses with QUIC as well.

### Context

A multiaddress is translated into socket address in `QuickTransport::dial` with the use of `QuicListener::get_socket_address`: https://github.com/paritytech/litep2p/blob/954d01b3afea6ee73973603a517739f3166a5603/src/transport/quic/mod.rs#L253

It is later matched for IPv4 & IPv6 in the [`QuicListener::get_socket_address`](https://github.com/paritytech/litep2p/blob/954d01b3afea6ee73973603a517739f3166a5603/src/transport/quic/listener.rs#L104) implementation, generating an error if it's a DNS address.

DNS addresses should be also handled and resolved using litep2p resolver passed to `QuicTransport::new` during the initialization, like it's done in TCP & WebSocket transports.

dharjeezy:
Extends QUIC transport to resolve DNS multiaddresses (/dns/, /dns4/, /dns6/) before connecting, similarly with TCP and WebSocket transports. closes #406


src/transport/quic/mod.rs
```diff
@@ -129,6 +133,73 @@ pub(crate) struct QuicTransport {
...
+    ) -> Result<(AddressType, Option<PeerId>), AddressError> {
+        tracing::trace!(target: LOG_TARGET, ?address, "parse quic multi address");
+
+        let mut iter = address.iter();
+
+        let address_type = match iter.next() {
+            Some(Protocol::Ip4(addr)) => match iter.next() {
+                Some(Protocol::Udp(port)) =>
+                    AddressType::Socket(SocketAddr::new(IpAddr::V4(addr), port)),
+                _ => return Err(AddressError::InvalidProtocol),
```
lexnv: nit: The error cases are not expected to happen in production environemnts. Could we capture more details into the logs, similar to how we parse in tcp/wss:

https://github.com/paritytech/litep2p/blob/dee98766d053a5454cc9218247d292483d0ab415/src/transport/common/listener.rs#L364-L371

src/transport/quic/mod.rs
```diff
@@ -129,6 +133,73 @@ pub(crate) struct QuicTransport {
...
+            _ => return Err(AddressError::InvalidProtocol),
+        };
+
+        match iter.next() {
+            Some(Protocol::QuicV1) => {}
+            _ => return Err(AddressError::InvalidProtocol),
+        }
+
+        let maybe_peer = match iter.next() {
+            Some(Protocol::P2p(multihash)) => Some(PeerId::from_multihash(multihash)?),
```
lexnv:
```suggestion
            Some(Protocol::P2p(multihash)) => Some(PeerId::from_multihash(multihash).map_err(AddressError::InvalidPeerId)?),
```

src/transport/quic/mod.rs
```diff
@@ -129,6 +133,73 @@ pub(crate) struct QuicTransport {
...
+            None => None,
+            _ => return Err(AddressError::PeerIdMissing),
```
lexnv: It is ok to have the peer ID missing, we are expecting the final protocol of the multiaddr to either have a `/p2p` prefix or not.

In this case, we have detected here a different protocol:


```suggestion
      protocol => {
            tracing::error!(
                target: LOG_TARGET,
                ?protocol,
                "invalid protocol, expected `P2p` or `None`"
            );
            return Err(AddressError::InvalidProtocol);
        }
```

src/transport/quic/mod.rs
```diff
@@ -129,6 +133,73 @@ pub(crate) struct QuicTransport {
...
 impl QuicTransport {
+    /// Parse a QUIC multiaddress, supporting both IP and DNS addresses.
+    ///
+    /// Accepted formats:
+    /// - `/ip4/<addr>/udp/<port>/quic-v1[/p2p/<peer-id>]`
+    /// - `/ip6/<addr>/udp/<port>/quic-v1[/p2p/<peer-id>]`
+    /// - `/dns/<host>/udp/<port>/quic-v1[/p2p/<peer-id>]`
+    /// - `/dns4/<host>/udp/<port>/quic-v1[/p2p/<peer-id>]`
+    /// - `/dns6/<host>/udp/<port>/quic-v1[/p2p/<peer-id>]`
+    fn parse_quic_address(
```
lexnv: We could extend the `QuicListener::get_socket_address` function to support DNS as well

Currently, that function is no longer used in favour of this one.

lexnv:
Left some tiny comments, overall looks like going in the right direction! Thanks for contributing 🙏

src/transport/quic/listener.rs
```diff
@@ -60,15 +61,25 @@ impl QuicListener {
         let mut listen_addresses = Vec::new();
 
         for address in addresses.into_iter() {
-            let (listen_address, _) = Self::get_socket_address(&address)?;
+            let listen_address = match Self::get_socket_address(&address)?.0 {
```
lexnv: nit: If one of the addresses is malformed, we'll stop listening on any address. Instead, could we collect valid addresses with a `filter_map` similar to tcp?

lexnv (approved):
LGTM!

lexnv:
@dharjeezy it looks like quic tests are hanging, could you have a look? Thanks! 🙏 

```
test transport::quic::tests::test_quinn has been running for over 60 seconds
```

bkchr (approved):

PR #554, opened 2026-03-03
Make webrtc-specific multistream-select negotiation spec compliant

### LINKED ISSUE #75: `multistream-select` implementation used by WebRTC doesn't conform to the spec

haikoschol:
This PR brings the WebRTC-specific multistream-select implementation in line with the [spec](https://github.com/libp2p/specs/blob/master/connections/README.md#multistream-select). Before this change, the header, the desired protocol and all fallback protocols were sent in a single multistream-select message. The spec mandates that only the desired protocol is to be sent instead. If the peer responds with a `na`, indicating it does not support this protocol, another multistream-select message should be sent, containing the first fallback protocol. This is to be repeated until the peer accepts a protocol or the dialer runs out of fallbacks.

This fix is required for interop with smoldot. Without this fix, the following happens:

- The dialer proposes a protocol with fallbacks in a single message (e.g. `/multistream/1.0.0\n/c7fd75141ca230c62e99af66d68f9d1f84b0821bcc1e17d2be92aaf5602d1c0e/block-announces/1\n/dot/block-announces/1\n`).
- Smoldot reads from its input buffer until it sees `/c7fd75141ca230c62e99af66d68f9d1f84b0821bcc1e17d2be92aaf5602d1c0e/block-announces/1\n` and sends back a confirmation.
- The fallback protocol remains in the input buffer and is misinterpreted by Smoldot as the block announce handshake. Smoldot does not verify the handshake on inbound substreams.
- The dialer sends the actual block announce handshake (70 byte payload).
- Smoldot misinterprets this as the first block announce message and fails to decode it and logs this error to the browser console: `DecodeBlockAnnounceError(Verify)`.

closes paritytech/litep2p#75

### DIFF
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -339,9 +339,9 @@ impl WebRtcDialerState {
         protocol: ProtocolName,
         fallback_names: Vec<ProtocolName>,
     ) -> crate::Result<(Self, Vec<u8>)> {
-        let message = webrtc_encode_multistream_message(std::iter::once(Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        )))?
+        ))?
         .freeze()
         .to_vec();
 
@@ -368,9 +368,9 @@ impl WebRtcDialerState {
         self.protocol = next;
         self.state = HandshakeState::WaitingResponse;
 
-        let message = webrtc_encode_multistream_message(std::iter::once(Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        )))?
+        ))?
         .freeze()
         .to_vec();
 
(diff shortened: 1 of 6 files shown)


haikoschol:
The [last commit](https://github.com/paritytech/litep2p/pull/554/changes/7473d5fe77afbf0d98522f28dd887ffef8d47cad) ended up being a big larger than expected. Originally I tried to get rid of `drain_trailing_protocols` because it was based on invalid assumptions and had a confusing name from its original usage.

But then I realized that the current multistream-select implementation still does not cover the valid scenario of receiving the header and the first proposed protocol in two separate webrtc protobuf messages. While implementing support for this I also made some changes that reduce the number of allocations when decoding multistream-select messages.

I think the result is overall better than before though. I've made sure the tests still make sense and tested it manually with smoldot.

Still, these changes are not necessary, more of a nice-to-have. If you prefer to keep the PR as small as possible, I can revert the commit.

lexnv:
Closing this in favor of:
- https://github.com/paritytech/litep2p/pull/573

PR #562, opened 2026-04-03
fix: recover outbound opens from stale primary handles

peter941221:
# PR Body - litep2p transport fix + force-close attribution for `polkadot-sdk#11540`

## Summary

This PR does two narrow things that came out of the `paritytech/polkadot-sdk#11540` investigation:

1. Fix outbound substream opening when the cached primary handle is stale but a live secondary handle still exists.
2. Add protocol-aware logging around notification-channel clogging / `ForceClose` so the next field run can answer which protocol is actually triggering the teardown path.

## Transport fix

In `TransportService`, outbound open previously always preferred the cached primary connection.

That can leave the transport stuck in a bad state when:

- the primary handle is already stale
- a live secondary connection still exists
- or `try_get_permit()` succeeds on primary but the send path immediately fails with `ConnectionClosed`

This PR changes that path so that:

- a live secondary is promoted before opening the outbound substream if the primary handle is stale
- if both cached handles are stale, `open_substream()` returns `ConnectionClosed` but keeps the peer context until the real `ConnectionClosed` propagation arrives
- if the primary send path fails with `ConnectionClosed`, the transport retries once through fresh connection selection so a live secondary can recover the open

## Force-close attribution logs

The latest `#11540` thread also surfaced a different live-runtime question: when notification backpressure triggers `ForceClose`, which protocol is actually clogging the queue?

To make that observable, this PR carries `protocol_name` and `sync_channel_size` into `NotificationHandle` and logs:

- when a clogged sync queue first queues `ForceClose`
- when queuing `ForceClose` fails
- when the handle remains clogged but a close is already queued
- when the clogged state is cleared
- when `NotificationProtocol` processes the resulting `ForceClose` command

This is instrumentation only; it does not change the current close semantics.

## Tests

Local `litep2p`:

```bash
cargo test protocol::transport_service::tests:: -- --nocapture
cargo test protocol::notification::tests:: -- --nocapture
```

Path-patched `polkadot-sdk` validation in WSL:

```bash
SKIP_WASM_BUILD=1 cargo test --locked -p sc-network notification_protocol_ -- --nocapture
SKIP_WASM_BUILD=1 cargo test --locked -p sc-network reserved_peer_close_triggers_immediate_reopen -- --nocapture
SKIP_WASM_BUILD=1 cargo test --locked -p sc-network closing_primary_with_only_open_desired_secondary_closes_instead_of_replacing -- --nocapture
SKIP_WASM_BUILD=1 cargo test --locked -p sc-network notification_sink_replacement_stays_internal_and_keeps_peer_live -- --nocapture
SKIP_WASM_BUILD=1 cargo test --locked -p sc-network inflight_secondary_validation_closes_peer_until_replacement_really_opens -- --nocapture
```

## Scope note

I am not claiming this is the single complete explanation for every disconnect reported in `paritytech/polkadot-sdk#11540`.

My current reading is narrower:

- this PR fixes one transport-state correctness hole around stale primary / live secondary handling
- and it makes the notification-triggered `ForceClose` path much easier to attribute in the next field run

That felt like the safest honest slice to upstream first.

### DIFF
--- a/src/protocol/notification/config.rs
+++ b/src/protocol/notification/config.rs
@@ -93,8 +93,14 @@ impl Config {
         let (notif_tx, notif_rx) = channel(DEFAULT_CHANNEL_SIZE);
         let (command_tx, command_rx) = channel(DEFAULT_CHANNEL_SIZE);
         let handshake = Arc::new(RwLock::new(handshake));
-        let handle =
-            NotificationHandle::new(event_rx, notif_rx, command_tx, Arc::clone(&handshake));
+        let handle = NotificationHandle::new(
+            protocol_name.clone(),
+            sync_channel_size,
+            event_rx,
+            notif_rx,
+            command_tx,
+            Arc::clone(&handshake),
+        );
 
         (
             Self {
--- a/src/protocol/notification/config.rs
+++ b/src/protocol/notification/config.rs
@@ -94,11 +94,12 @@ impl Config {
         let (command_tx, command_rx) = channel(DEFAULT_CHANNEL_SIZE);
         let handshake = Arc::new(RwLock::new(handshake));
         let handle = NotificationHandle::new(
+            protocol_name.clone(),
+            sync_channel_size,
             event_rx,
             notif_rx,
             command_tx,
             Arc::clone(&handshake),
-            protocol_name.clone(),
         );
 
         (
--- a/src/protocol/notification/handle.rs
+++ b/src/protocol/notification/handle.rs
@@ -170,6 +170,12 @@ impl NotificationSink {
 /// Handle allowing the user protocol to interact with the notification protocol.
 #[derive(Debug)]
 pub struct NotificationHandle {
+    /// Protocol name served by this handle.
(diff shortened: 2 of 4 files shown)


peter941221:
Quick follow-up on this PR after the newer field investigation on `polkadot-sdk#11540`.

The later runtime attribution run ended up showing that the observed `ChannelClogged -> ForceClose` events in that setup were concentrated on `grandpa/1`, which is exactly why I wanted the protocol-aware logging in this branch upstream rather than keeping it only in a local patch.

I still view this PR as a deliberately narrow first slice:

- fix one transport-state correctness hole in `TransportService` when the cached primary is stale but a live secondary still exists
- keep protocol-aware `ForceClose` attribution logs in the notification path so the next runtime collection is easier to interpret

I am not trying to claim this PR fully resolves every disconnect reported in `#11540`.

If the preferred review shape here is:

- split the transport fix from the attribution logs
- trim the instrumentation further
- or take only one of the two pieces first

I am happy to respin it that way.

PR #564, opened 2026-04-21
deps: upgrade multiaddr 0.18, multihash 0.19, libp2p 0.56 

anstylian:
Triggered by the deprecation of https://crates.io/crates/core2 which was
  a transitive dependency pulled in by multihash 0.17 via multihash-codetable 0.1.0.
  Upgrading to multihash-codetable 0.2.1 removes the core2 dependency entirely.

  - multiaddr 0.17 → 0.18.2: Protocol::P2p now carries multiaddr::PeerId
    instead of raw Multihash; updated all call sites to use peer.into()
    via a new From<PeerId> for multiaddr::PeerId bridge impl

  - multihash 0.17 → 0.19.4: Multihash is now generic (Multihash<64>);
    Code and MultihashDigest moved to multihash-codetable 0.2.1; Code::Identity
    removed from codetable so identity peer IDs now use a manual 0x00 constant

  - libp2p 0.51 → 0.56 (dev dep): migrate conformance tests to new fluent
    SwarmBuilder API, remove keep_alive::Behaviour (handled by swarm idle
    timeout in 0.56), rename kad::Kademlia/KademliaConfig/KademliaEvent,
    fix add_external_address signature, force kad::Mode::Server on loopback
    test nodes to avoid silent query discard in client mode

  - cid 0.11.1 → 0.11.2 to drop the core2 dependency

### DIFF
--- a/src/transport/manager/address.rs
+++ b/src/transport/manager/address.rs
@@ -605,8 +605,6 @@ mod tests {
         assert_eq!(store.addresses.get(&address).unwrap().score, failure_score);
     }
 
-    // --- AddressRecord::new peer ID handling ---
-
     #[test]
     fn address_record_new_appends_peer_id_when_missing() {
         let peer = PeerId::random();
--- a/src/transport/manager/address.rs
+++ b/src/transport/manager/address.rs
@@ -22,7 +22,6 @@ use crate::{error::DialError, PeerId};
 
 use ip_network::IpNetwork;
 use multiaddr::{Multiaddr, Protocol};
-use multihash::Multihash;
 
 use std::collections::{hash_map::Entry, HashMap};
 
@@ -70,9 +69,7 @@ impl AddressRecord {
     /// append the provided `PeerId` to the address.
     pub fn new(peer: &PeerId, address: Multiaddr, score: i32) -> Self {
         let address = if !std::matches!(address.iter().last(), Some(Protocol::P2p(_))) {
-            address.with(Protocol::P2p(
-                Multihash::from_bytes(&peer.to_bytes()).expect("valid peer id"),
-            ))
+            address.with(Protocol::P2p((*peer).into()))
         } else {
             address
         };
@@ -536,9 +533,7 @@ mod tests {
         let address = Multiaddr::empty()
             .with(Protocol::Ip4(Ipv4Addr::new(8, 8, 8, 8)))
             .with(Protocol::Tcp(9999))
-            .with(Protocol::P2p(
-                multihash::Multihash::from_bytes(&peer.to_bytes()).unwrap(),
-            ));
+            .with(Protocol::P2p(peer.into()));
 
         let record = AddressRecord::from_multiaddr(address.clone()).unwrap();
         assert_eq!(record.score, 0);
@@ -584,9 +579,7 @@ mod tests {
         let address = Multiaddr::empty()
             .with(Protocol::Ip4(Ipv4Addr::new(8, 8, 8, 8)))
             .with(Protocol::Tcp(9999))
-            .with(Protocol::P2p(
-                multihash::Multihash::from_bytes(&peer.to_bytes()).unwrap(),
-            ));
+            .with(Protocol::P2p(peer.into()));
 
         // First, add the address normally.
         let record = AddressRecord::from_multiaddr(address.clone()).unwrap();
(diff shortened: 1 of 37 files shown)


shawntabrizi:
@anstylian now that most clippy fixes are in, you should be able to update from master and make this PR much smaller

lexnv:
IIRC we had issues with bumping multiaddr / multihash while bringing litep2p into substrate, mostly because we use an outdated libp2p version

Would be good to have some confirmation with a git-path dependency that substrate will compile 🙏

anstylian:
> IIRC we had issues with bumping multiaddr / multihash while bringing litep2p into substrate, mostly because we use an outdated libp2p version
> 
> Would be good to have some confirmation with a git-path dependency that substrate will compile 🙏

 I expect polkadot-sdk and zombienet-sdk to need follow-up changes for this bump. I am tracking those changes at the moment.

If you have any suggestions for additional testing or specific checks that would give us confidence this did not break downstream usage, please let me know.

bkchr:
> Would be good to have some confirmation with a git-path dependency that substrate will compile 🙏

I checked it locally with this pr and it worked. Requires some changes in polkadot-sdk, but nothing breaking. I think we can go ahead.

src/transport/webrtc/opening.rs
```diff
@@ -272,9 +272,9 @@ impl OpeningWebRtcConnection {
             .clone()
             .bytes;
 
-        const MULTIHASH_SHA256_CODE: u64 = 0x12;
-        let certificate = Multihash::wrap(MULTIHASH_SHA256_CODE, &remote_fingerprint)
-            .expect("fingerprint's len to be 32 bytes");
+        let certificate =
+            multihash::Multihash::<64>::wrap(u64::from(Code::Sha2_256), &remote_fingerprint)
```
bkchr:
```suggestion
            multihash::Multihash::<64>::wrap(Code::Sha2_256, &remote_fingerprint)
```

src/transport/webrtc/opening.rs
```diff
@@ -188,8 +189,7 @@ impl OpeningWebRtcConnection {
 
     /// Convert `Fingerprint` to bytes.
     fn fingerprint_to_bytes(fingerprint: &Fingerprint) -> Vec<u8> {
-        const MULTIHASH_SHA256_CODE: u64 = 0x12;
-        Multihash::wrap(MULTIHASH_SHA256_CODE, &fingerprint.bytes)
+        multihash::Multihash::<64>::wrap(u64::from(Code::Sha2_256), &fingerprint.bytes)
```
bkchr:
```suggestion
        multihash::Multihash::<64>::wrap(Code::Sha2_256.into(), &fingerprint.bytes)
```

bkchr (approved):

src/transport/webrtc/opening.rs
anstylian: applied at commit 00abd827920cfd520aa8f3b8aa1cd80b922634dc

src/transport/webrtc/opening.rs
anstylian: applied at commit 00abd827920cfd520aa8f3b8aa1cd80b922634dc

src/transport/manager/address.rs
```diff
@@ -612,6 +605,31 @@ mod tests {
         assert_eq!(store.addresses.get(&address).unwrap().score, failure_score);
     }
 
+    // --- AddressRecord::new peer ID handling ---
+
```
lexnv:
```suggestion
```

src/transport/manager/peer_state.rs
```diff
@@ -942,4 +944,47 @@ mod tests {
             }
         );
     }
+
+    // --- ensure_peer_id (tested via ConnectionRecord::new) ---
+
```
lexnv:
```suggestion
```

lexnv (approved):
LGTM! Thanks for contributing! 🙏

src/transport/manager/address.rs
anstylian: handled at commit: eb5cfeadd3cea470eaeaf4f3f18a96226fbd4e66

src/transport/manager/peer_state.rs
anstylian: handled at commit: eb5cfeadd3cea470eaeaf4f3f18a96226fbd4e66

PR #565, opened 2026-04-22
fix clippy errors and warnings

anstylian:
The PR includes only fixes for Clippy warnings and updates to the `.gitignore` file to ignore the `target` folder created by internal crates located in the `fuzz/` directory.


shawntabrizi:
Is there a ci pipeline for clippy? If not you should probably add one?

Otherwise the problem with occur again

anstylian:
> Is there a ci pipeline for clippy? If not you should probably add one?
> 
> Otherwise the problem with occur again

There is a CI pipeline for Clippy. I didn’t intend to change the CI pipeline, but since you mentioned it, I’ll make two changes and let me know if you’d like me to revert them. 

I have added `--all-targets` for `cargo clippy` and `cargo test`.

tests/conformance/rust/kademlia.rs
```diff
@@ -449,20 +449,16 @@ async fn litep2p_add_provider_to_libp2p() {
...
-                            assert_eq!(providers.len(), 1);
-                            let record = providers.pop().unwrap();
-
-                            assert_eq!(record.key.as_ref(), key);
-                            assert_eq!(record.provider, litep2p_peer_id);
-                            assert_eq!(record.addresses, vec![litep2p_public_addr.clone()]);
-                            break
-                        }
-                    }
+                if let SwarmEvent::Behaviour(BehaviourEvent::Kad(Libp2pKademliaEvent::InboundRequest{ request: InboundRequest::AddProvider{..} })) = event {
```
lexnv: nit: we ll need to format this manually since its behind a `select!` macro

Cargo.toml
```diff
@@ -87,8 +87,8 @@ libp2p = { version = "0.51.4", features = [
     "kad",
     "websocket",
     "yamux",
-    "quic",
 ]}
+libp2p-quic = { version = "0.7.0-alpha.3", features = ["tokio"] }
```
lexnv, line 91: dq: Why are we not relying on the libp2p quic feature and importing the alpha release instead?

lexnv (approved):
LGTM! Thanks for tackling this 🙏

Cargo.toml
anstylian: Using `quic` from libp2p version 0.51.4, produces this warning: 
```
cargo check --all-features --all-targets
   Compiling litep2p v0.13.3 (/home/angelos/Documents/git/litep2p)
warning: use of deprecated module `libp2p::quic`: `quic` is only in alpha status. Please depend on `libp2p-quic` directly and don't ues the `quic` feature of `libp2p`.
  --> tests/conformance/rust/quic_ping.rs:25:21
   |
25 |     identity, ping, quic,
   |                     ^^^^
   |
   = note: `#[warn(deprecated)]` on by default

warning: `litep2p` (test "mod") generated 1 warning
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.76s
```

I used the version of libp2p-quic that was available when the deprecation warning was introduced.

tests/conformance/rust/kademlia.rs
anstylian: handled at: 52f6190fdb984d97649b59e10563378c1f5bcc5e

shawntabrizi (approved):

PR #569, opened 2026-04-23
refactor(webrtc): update str0m dep

### LINKED ISSUE #421: webrtc: Update dependencies

gab8i:
The main update is related to how `CryptoProvider` is generated, before while creating `DtlsCert` you had to specify which `CryptoProvider` to use while now the same `CryptoProvider` can only be generated from the specified feature flag.

edit after https://github.com/paritytech/litep2p/pull/569#discussion_r3131620034

'fingerprint' method over 'DtlsCertificate' was inherited by a crate under a default feature flag, which cannot be used otherwise it would take over openssl crypto backend priority. The fingerprint needs thus to be computed 'manually' with the crypto backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ rcgen = { version = "0.14.5", optional = true }
 # End of Quic related dependencies.
 
 # WebRTC related dependencies. WebRTC is an experimental feature flag. The dependencies must be updated.
-str0m = { version = "0.11.1", optional = true }
+str0m = { version = "0.18.1", default-features = false, features = ["openssl"], optional = true }
 # End of WebRTC related dependencies.
 
 # Fuzzing related dependencies.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,6 +65,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "76a2e8124351fda1ef8aaaa3bbd7ebbcb486bbcd4225aca0aa0d84bb2db8fecb"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
+
+[[package]]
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -66,7 +66,7 @@ rcgen = { version = "0.14.5", optional = true }
 # End of Quic related dependencies.
 
 # WebRTC related dependencies. WebRTC is an experimental feature flag. The dependencies must be updated.
-str0m = { version = "0.11.1", optional = true }
+str0m = { version = "0.18.1", optional = true, features = ["openssl"] }
```
lexnv: Maybe we should add a `default-features = false` here?

It looks like by default we have enabled https://github.com/algesten/str0m/blob/a07b11e7e6e304495e4233f3ce4a2e55166ee39b/Cargo.toml#L100:
- aws-lc-rs 
- examples

The `from_feature_flags` https://github.com/algesten/str0m/blob/a07b11e7e6e304495e4233f3ce4a2e55166ee39b/src/crypto/mod.rs#L29-L36 says the prioiry order is:

> /// Priority order: aws-lc-rs, rust-crypto, openssl, wincrypto (Windows only)

gab8i: Oh wow, yes, thank you! I didn't notice it, that's definitely required!

gab8i: Interesting, the method 'fingerprint' over 'DtlsCertificate' was inherited by a crate under a default feature flag, which could not be used otherwise it would take over crypto backend priority. Pushing right now a wip commit to see if a manual fingerprint computation passes CI

lexnv (approved):
Nice!

skunert (approved):

Issue #312, opened 2024-12-17
[litep2p-milestone]: Implement WebRTC transport layer


lexnv:
# The Vision

Enhance the robustness and fault tolerance of the WebRTC protocol.

WebRTC serves as an alternative to the WebSocket transport layer, enabling lightweight client connectivity.

The goal is to establish seamless connectivity between lightweight clients (e.g., https://github.com/smol-dot/smoldot/issues/1806) and substrate-based chains over the WebRTC transport layer.

# The Plan

We need to focus on the following key areas:
- Ensuring proper implementation of the specification and maintaining smoldot connectivity
- Analyzing and optimizing memory usage to eliminate any potential memory leaks and reduce unnecessary cloning
- Implementing backpressure handling within the WebRTC layer
- Exposing detailed transport layer metrics
- Expanding and improving our testing scenarios

```[tasklist]
### Tasks
- [ ] https://github.com/paritytech/litep2p/issues/45
- [ ] https://github.com/paritytech/litep2p/issues/107
- [ ] https://github.com/paritytech/litep2p/issues/75
- [ ] https://github.com/paritytech/litep2p/issues/43
- [ ] https://github.com/paritytech/litep2p/issues/42
- [ ] https://github.com/paritytech/litep2p/issues/41
```

PR #573, opened 2026-05-13
feat(webrtc): multistream-select protocol implementation

### LINKED ISSUE #75: `multistream-select` implementation used by WebRTC doesn't conform to the spec

gab8i:
The following PR just wraps the work done within https://github.com/paritytech/litep2p/pull/554

The only reason for this PR and not the original one is a matter of simplicity: WebRTC work has been continued over pending PRs to reach a stable point, and now it is easier to split the final work into multiple PRs from the main branch where WebRTC has been e2e tested.

Now include also #576:
After the multistream-select header has been exchanged, only protocol and na messages should flow between peers. propose_next_fallback was incorrectly transitioning state back to WaitingResponse (expecting another header) and wrapping the outgoing protocol message with the header. Keep the state at WaitingProtocol and emit the protocol message without the header prefix.

### DIFF
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -339,9 +339,9 @@ impl WebRtcDialerState {
         protocol: ProtocolName,
         fallback_names: Vec<ProtocolName>,
     ) -> crate::Result<(Self, Vec<u8>)> {
-        let message = webrtc_encode_multistream_message(std::iter::once(Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        )))?
+        ))?
         .freeze()
         .to_vec();
 
@@ -368,9 +368,9 @@ impl WebRtcDialerState {
         self.protocol = next;
         self.state = HandshakeState::WaitingResponse;
 
-        let message = webrtc_encode_multistream_message(std::iter::once(Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        )))?
+        ))?
         .freeze()
         .to_vec();
 
@@ -931,9 +931,9 @@ mod tests {
 
     #[test]
     fn negotiate_main_protocol() {
-        let message = webrtc_encode_multistream_message(vec![Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(&b"/13371338/proto/1"[..]).unwrap(),
-        )])
+        ))
         .unwrap()
         .freeze();
 
@@ -953,9 +953,9 @@ mod tests {
 
     #[test]
     fn negotiate_fallback_protocol() {
-        let message = webrtc_encode_multistream_message(vec![Message::Protocol(
+        let message = webrtc_encode_multistream_message(Message::Protocol(
             Protocol::try_from(&b"/sup/proto/1"[..]).unwrap(),
-        )])
+        ))
         .unwrap()
         .freeze();
 
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -339,9 +339,12 @@ impl WebRtcDialerState {
         protocol: ProtocolName,
         fallback_names: Vec<ProtocolName>,
     ) -> crate::Result<(Self, Vec<u8>)> {
-        let message = webrtc_encode_multistream_message(Message::Protocol(
-            Protocol::try_from(protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        ))?
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(
+                Protocol::try_from(protocol.as_ref()).map_err(|_| Error::InvalidData)?,
+            ),
+            true,
+        )?
         .freeze()
         .to_vec();
 
@@ -368,9 +371,12 @@ impl WebRtcDialerState {
         self.protocol = next;
         self.state = HandshakeState::WaitingResponse;
 
-        let message = webrtc_encode_multistream_message(Message::Protocol(
-            Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
-        ))?
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(
+                Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
+            ),
+            true,
+        )?
         .freeze()
         .to_vec();
 
@@ -931,9 +937,10 @@ mod tests {
 
     #[test]
     fn negotiate_main_protocol() {
-        let message = webrtc_encode_multistream_message(Message::Protocol(
-            Protocol::try_from(&b"/13371338/proto/1"[..]).unwrap(),
-        ))
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(Protocol::try_from(&b"/13371338/proto/1"[..]).unwrap()),
+            true,
+        )
         .unwrap()
         .freeze();
 
@@ -953,9 +960,10 @@ mod tests {
 
     #[test]
     fn negotiate_fallback_protocol() {
-        let message = webrtc_encode_multistream_message(Message::Protocol(
-            Protocol::try_from(&b"/sup/proto/1"[..]).unwrap(),
-        ))
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(Protocol::try_from(&b"/sup/proto/1"[..]).unwrap()),
+            true,
+        )
         .unwrap()
         .freeze();
 
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -465,13 +465,6 @@ impl WebRtcDialerState {
                         return Ok(HandshakeResult::Succeeded(self.protocol.clone()));
                     }
 
-                    for fallback in &self.fallback_names {
-                        if fallback.as_bytes() == protocol.as_ref() {
-                            check_trailing_bytes(&remaining);
-                            return Ok(HandshakeResult::Succeeded(fallback.clone()));
-                        }
-                    }
-
                     return Err(crate::error::NegotiationError::MultistreamSelectError(
                         NegotiationError::Failed,
                     ));
@@ -1024,6 +1017,30 @@ mod tests {
                 assert_eq!(negotiated, ProtocolName::from("/sup/proto/1"))
             }
             _ => panic!("invalid event"),
+        }
+    }
+
+    #[test]
+    fn reject_unproposed_fallback_confirmation() {
+        let (mut dialer_state, _message) = WebRtcDialerState::propose(
+            ProtocolName::from("/13371338/proto/1"),
+            vec![ProtocolName::from("/sup/proto/1")],
+        )
+        .unwrap();
+
+        // The dialer has only proposed the main protocol. The fallback is stored for a
+        // later round and must not be accepted until `propose_next_fallback()` sends it.
+        let mut response = BytesMut::with_capacity(64);
+        response.put_u8(MSG_MULTISTREAM_1_0.len() as u8);
+        Message::Header(HeaderLine::V1).encode(&mut response).unwrap();
+
+        let fallback = Protocol::try_from(&b"/sup/proto/1"[..]).expect("valid protocol name");
+        response.put_u8((fallback.as_ref().len() + 1) as u8);
+        Message::Protocol(fallback).encode(&mut response).unwrap();
+
+        match dialer_state.register_response(response.freeze().to_vec()) {
+            Err(error::NegotiationError::MultistreamSelectError(NegotiationError::Failed)) => {}
+            event => panic!("expected unproposed fallback to be rejected, got: {event:?}"),
         }
     }
 }
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -427,16 +427,32 @@ impl WebRtcDialerState {
                 "Decoded message while registering response",
             );
 
+            let check_trailing_bytes = |bytes: &Bytes| {
+                // remote may be optimistically pipelining the first application frame
+                if !bytes.is_empty() {
+                    tracing::warn!(
+                        bytes_len = bytes.len(),
+                        "trailing bytes after multistream-select negotiation were discarded"
+                    );
+                }
+            };
+
             match (&self.state, message) {
                 (HandshakeState::WaitingResponse, Ok(Message::Header(HeaderLine::V1))) => {
                     self.state = HandshakeState::WaitingProtocol;
                 }
-                (HandshakeState::WaitingResponse, Ok(Message::Protocol(_))) => {
+                (HandshakeState::WaitingResponse, Ok(msg)) => {
+                    tracing::trace!(
+                      target: LOG_TARGET,
+                      ?msg,
+                      "Expected header response from peer, got different message"
+                    );
                     return Err(crate::error::NegotiationError::MultistreamSelectError(
                         NegotiationError::Failed,
                     ));
                 }
-                (_, Ok(Message::NotAvailable)) => {
+                (HandshakeState::WaitingProtocol, Ok(Message::NotAvailable)) => {
+                    check_trailing_bytes(&remaining);
                     return Ok(HandshakeResult::Rejected);
                 }
                 (HandshakeState::WaitingProtocol, Ok(Message::Protocol(protocol))) => {
@@ -445,11 +461,13 @@ impl WebRtcDialerState {
                     }
 
                     if self.protocol.as_bytes() == protocol.as_ref() {
+                        check_trailing_bytes(&remaining);
                         return Ok(HandshakeResult::Succeeded(self.protocol.clone()));
                     }
 
                     for fallback in &self.fallback_names {
                         if fallback.as_bytes() == protocol.as_ref() {
+                            check_trailing_bytes(&remaining);
                             return Ok(HandshakeResult::Succeeded(fallback.clone()));
                         }
                     }
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -33,7 +33,7 @@ use crate::{
     types::protocol::ProtocolName,
 };
 
-use bytes::{Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
 use futures::prelude::*;
 use std::{
     convert::TryFrom as _,
@@ -369,12 +369,10 @@ impl WebRtcDialerState {
     /// Returns `None` if there are no more fallback protocols to try.
     /// Returns `Some(message)` with the encoded message to send, containing the protocol name.
     pub fn propose_next_fallback(&mut self) -> crate::Result<Option<Vec<u8>>> {
-        if self.fallback_names.is_empty() {
+        let Some(next) = self.fallback_names.pop() else {
             return Ok(None);
-        }
+        };
 
-        // UNWRAP: fallback_names has just been checked to not be empty.
-        let next = self.fallback_names.pop().unwrap();
         self.protocol = next;
 
         let message = webrtc_encode_multistream_message(
@@ -394,8 +392,7 @@ impl WebRtcDialerState {
         &mut self,
         payload: Vec<u8>,
     ) -> Result<HandshakeResult, crate::error::NegotiationError> {
-        let bytes = Bytes::from(payload);
-        let mut remaining = bytes.clone();
+        let mut remaining = Bytes::from(payload);
 
         while !remaining.is_empty() {
             let (len, tail) = unsigned_varint::decode::usize(&remaining).map_err(|error| {
@@ -408,8 +405,6 @@ impl WebRtcDialerState {
                 error::NegotiationError::ParseError(ParseError::InvalidData)
             })?;
 
-            let len_size = remaining.len() - tail.len();
-
             if len > tail.len() {
                 tracing::debug!(
                     target: LOG_TARGET,
@@ -421,8 +416,9 @@ impl WebRtcDialerState {
                 return Err(error::NegotiationError::ParseError(ParseError::InvalidData));
             }
 
-            let payload = remaining.slice(len_size..len_size + len);
-            remaining = remaining.slice(len_size + len..);
+            let len_size = remaining.len() - tail.len();
+            remaining.advance(len_size);
+            let payload = remaining.split_to(len);
             let message = Message::decode(payload);
 
             tracing::trace!(
@@ -462,6 +458,11 @@ impl WebRtcDialerState {
                         NegotiationError::Failed,
                     ));
                 }
+                (HandshakeState::WaitingProtocol, Ok(Message::ListProtocols)) => {
+                    return Err(crate::error::NegotiationError::MultistreamSelectError(
+                        NegotiationError::ProtocolError(ProtocolError::InvalidMessage),
+                    ));
+                }
                 _ => {
                     return Err(crate::error::NegotiationError::StateMismatch);
                 }
@@ -480,7 +481,7 @@ mod tests {
     use super::*;
     use crate::multistream_select::{listener_select_proto, protocol::MSG_MULTISTREAM_1_0};
     use bytes::BufMut;
-    use std::{sync::OnceState, time::Duration};
+    use std::time::Duration;
     #[tokio::test]
     async fn select_proto_basic() {
         async fn run(version: Version) {
--- a/src/multistream_select/dialer_select.rs
+++ b/src/multistream_select/dialer_select.rs
@@ -333,11 +333,14 @@ pub struct WebRtcDialerState {
 impl WebRtcDialerState {
     /// Propose protocol to remote peer.
     ///
+    /// `fallback_names` must be in preference order, the first element is the
+    /// next protocol to try.
+    ///
     /// Return [`WebRtcDialerState`] which is used to drive forward the negotiation and an encoded
     /// `multistream-select` message that contains the protocol proposal for the substream.
     pub fn propose(
         protocol: ProtocolName,
-        fallback_names: Vec<ProtocolName>,
+        mut fallback_names: Vec<ProtocolName>,
     ) -> crate::Result<(Self, Vec<u8>)> {
         let message = webrtc_encode_multistream_message(
             Message::Protocol(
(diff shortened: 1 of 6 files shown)


src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +358,86 @@ impl WebRtcDialerState {
...
+
+        let next = self.fallback_names.remove(0);
+        self.protocol = next;
+        self.state = HandshakeState::WaitingResponse;
+
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(
+                Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
+            ),
+            true,
```
lexnv: Shouldn't we only send the `/multistream/1.0.0` header on the first negotiation? 

It seems we are enconding the header with every proposed fallback protocol:

```
# successful negotiation of the kad protocol, with missing version support
> /multistream/1.0.0\n/ipfs/kad/1.0.0\n
< /multistream/1.0.0\nna\n
> /ipfs/kad/0.2.3\n
< /ipfs/kad/0.2.3\n
```
- extracted from https://github.com/multiformats/multistream-select/blob/master/README.md

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +358,86 @@ impl WebRtcDialerState {
...
+    /// Propose the next fallback protocol to the remote peer.
+    ///
+    /// Returns `None` if there are no more fallback protocols to try.
+    /// Returns `Some(message)` with the encoded message to send, containing the protocol name.
+    pub fn propose_next_fallback(&mut self) -> crate::Result<Option<Vec<u8>>> {
+        if self.fallback_names.is_empty() {
+            return Ok(None);
+        }
+
+        let next = self.fallback_names.remove(0);
```
lexnv: We could replace the `Vec` with `VecDequeu` for `O(1)` removal compared to `O(n)` for the vec

src/multistream_select/dialer_select.rs
lexnv: Confirmed with the go-libp2p implementation, they are sending the header exactly once which could mean:

- T0: litep2p sends `/multistream/1.0.0\n/ipfs/kad/1.0.0\n `
- T1: remote responds with `/multistream/1.0.0\nna\n `
- T2: we'll send the next fallback protocol wrongly prefixed by the header `multistream/1.0.0\n` ++ `/ipfs/kad/0.2.3\n`
- T3: remote will read up to the `\n` and try to match `multistream/1.0.0\n` to an actually installed protocol, which fails and remote will send `naa`
- T4: remote continues to read our message: detects `/ipfs/kad/0.2.3\n` and accepts it
- T5: litep2p sees `naa` thinking it coresponds to the `/ipfs/kad/0.2.3\n` when it was actually sent as a response for the header


Reference: https://github.com/multiformats/go-multistream/blob/61d3adf7e79a2dea83d54f4c1797a1e5b8ad55b5/multistream.go#L217-L232

gab8i: Yes, sorry I didn't mentioned this, to  keep the initial work mostly as is and to make the PRs 'easy' to review I split the work and the fix for what you just described should be part of: https://github.com/paritytech/litep2p/pull/576

src/multistream_select/listener_select.rs
```diff
@@ -333,52 +332,121 @@ pub enum ListenerSelectResult {
...
+    // Save for zero-copy header echo (Bytes::clone is O(1)).
+    let raw_payload = payload.clone();
+
+    let first_msg = decode_multistream_message(&mut payload)?;
 
-    // skip the multistream-select header because it's not part of user protocols but verify it's
-    // present
-    if protocol_iter.next() != Some(PROTO_MULTISTREAM_1_0) {
+    let (protocol, header_in_this_payload) = match first_msg {
+        Message::Header(HeaderLine::V1) => {
```
lexnv: If we already received the header via `header_received` then we should fail here. This is effectively accepting the header multiple times similar to the dialer state

src/multistream_select/protocol.rs
```diff
@@ -228,30 +246,43 @@ impl Message {
...
+    let capacity = {
+        let msg_varint_len = unsigned_varint::encode::usize(msg_len, &mut varint_buf).len();
+        let total = if prepend_header {
+            let header_varint_len =
+                unsigned_varint::encode::usize(header_len, &mut varint_buf).len();
+            header_varint_len + header_len + msg_varint_len + msg_len
+        } else {
+            msg_varint_len + msg_len
+        };
+        total.min(super::length_delimited::MAX_FRAME_SIZE as usize)
```
lexnv: We should return an `InvalidData` error when the total length exceeds the max frame size.

Otherwise, we are just pre-allocating capacity for the `BytesMut` capped at `MAX_FRAME_SIZE`. If the bytes overwire exceeds the capacity, then the `BytesMut` would re-allocate and we'd return an invalid length data

Would be good to also add a debug log about this missbehavior so we could debug it in the future if it becomes an issue

src/transport/webrtc/connection.rs
```diff
@@ -329,9 +345,10 @@ impl WebRtcConnection {
         let protocols = self.protocol_set.protocols_with_keep_alives();
         let protocol_names = protocols.keys().cloned().collect();
         let (response, negotiated) =
-            match webrtc_listener_negotiate(protocol_names, payload.into())? {
+            match webrtc_listener_negotiate(protocol_names, payload.into(), header_received)? {
```
lexnv, line 348: The `webrtc_listener_negotiate` can actually receive a new header inside of it. If the `header_received` is false, but `webrtc_listener_negotiate` received the header, we'd not propagate this information back accepting spec-uncomplient connections

src/transport/webrtc/connection.rs
```diff
@@ -849,6 +958,20 @@ impl WebRtcConnection {
...
+                        if !self.rtc.is_alive() || !self.rtc.is_connected() {
+                            tracing::debug!(
+                                target: LOG_TARGET,
+                                peer = ?self.peer,
+                                ?protocol,
+                                is_alive = self.rtc.is_alive(),
+                                is_connected = self.rtc.is_connected(),
+                                "rejecting substream open: connection not healthy",
+                            );
+                            continue;
```
lexnv, line 996: When will the protocol that asked for a new substream be informed that the connection is tearing down?

If rtc remains stuck in `!is_alive or !is_connected` would we be leaking the connection and stalling it indefinetely?

src/multistream_select/dialer_select.rs
lexnv: Aah perfect. Could we cherry-pick / merge that commit here? It helps to have everything related to multistream-select in a single PR for having a clear picture about what's happening here, or if any other PRs are building on top of multistream-select could mention them in the description as part of a review series for this feature in isolation 🙏

gab8i: Yes, definitely! I was trying to keep the work taken from previous pr separated from mine but from a review perspective it doesn't make much sense!

src/multistream_select/dialer_select.rs
gab8i: Given that Vec is used all around for fallback names and there are usually only a couple of them, I thought just reversing the vector and then popping from the back would be more efficient, without having to work with VecDeque's ring buffer. Lmk if this makes sense to you, or if you think VecDeque is still a better option!

src/multistream_select/listener_select.rs
gab8i: Yes, that's right, thanks for spotting this!

src/multistream_select/protocol.rs
gab8i: Oh yes!! It doesn't make sense to cap it there but then allow the grow of the bytes while encoding the message.

src/transport/webrtc/connection.rs
gab8i: The header is the first message expected, so after any successfully handled payload the header has been received, either in a previous round or in the current one. This information is propagated back because, on a successful but still-negotiating round, the caller re-inserts the InboundOpening state with header_received: true

```rust
// Header has been exchanged after any successful round.
self.channels.insert(
    channel_id,
    ChannelState::InboundOpening {
        header_received: true,
    },
);
```

src/transport/webrtc/connection.rs
gab8i: Yes, this is a real issue: leaking connections! I already started working on, making sure that connection closes and errors are properly propagated without crashing anything is not straight forward. For sick of simplicity I will address this in a new PR on top of the stack to avoid having to deal with too many rebase conflicts!

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
+    ///
+    /// Returns `None` if there are no more fallback protocols to try.
+    /// Returns `Some(message)` with the encoded message to send, containing the protocol name.
+    pub fn propose_next_fallback(&mut self) -> crate::Result<Option<Vec<u8>>> {
+        if self.fallback_names.is_empty() {
+            return Ok(None);
+        }
+
+        // UNWRAP: fallback_names has just been checked to not be empty.
+        let next = self.fallback_names.pop().unwrap();
```
lexnv: I generally prefer avoiding unwraps where possible. It's harder to maintain long-term assumptions, and we might change things without noticing:


```suggestion
        let Some(next) = self.fallback_names.pop() else {
             return Ok(None);
        };
```

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
         &mut self,
         payload: Vec<u8>,
     ) -> Result<HandshakeResult, crate::error::NegotiationError> {
-        // All multistream-select messages are length-prefixed. Since this code path is not using
-        // multistream_select::protocol::MessageIO, we need to decode and remove the length here.
-        let remaining: &[u8] = &payload;
-        let (len, tail) = unsigned_varint::decode::usize(remaining).map_err(|error| {
-            tracing::debug!(
+        let bytes = Bytes::from(payload);
+        let mut remaining = bytes.clone();
```
lexnv: nit: No need for the extra clone (although it should be cheep since its behind Arc)

```suggestion
        let mut remaining = Bytes::from(payload);
```

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
-            ?message,
-            "Decoded message while registering response",
-        );
-
-        let mut protocols = match message {
-            Ok(Message::Header(HeaderLine::V1)) => {
-                vec![PROTO_MULTISTREAM_1_0]
+            let len_size = remaining.len() - tail.len();
+
+            if len > tail.len() {
```
lexnv, line 408: nit: lets use advance and try to avoid some manual index slicing below

```suggestion
            remaining.advance(len_size);
            if len > remaining.len() {
```

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
-                    _ => Err(error::NegotiationError::StateMismatch),
-                },
-            Ok(Message::ListProtocols) => return Err(error::NegotiationError::StateMismatch),
-            Err(_) => return Err(error::NegotiationError::ParseError(ParseError::InvalidData)),
-        };
 
-        protocols.extend(drain_trailing_protocols(remaining)?);
+            let payload = remaining.slice(len_size..len_size + len);
+            remaining = remaining.slice(len_size + len..);
+            let message = Message::decode(payload);
```
lexnv:
```suggestion
            let payload = remaining.split_to(len);
            let message = Message::decode(payload);
```

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
+                    return Err(crate::error::NegotiationError::MultistreamSelectError(
+                        NegotiationError::Failed,
+                    ));
+                }
+                (_, Ok(Message::NotAvailable)) => {
+                    return Ok(HandshakeResult::Rejected);
                 }
-                (HandshakeState::WaitingProtocol, Some(protocol)) => {
+                (HandshakeState::WaitingProtocol, Ok(Message::Protocol(protocol))) => {
                     if protocol == PROTO_MULTISTREAM_1_0 {
```
lexnv, line 459: We could improve a bit the robustness of this function wrt the state machine management.

Currently, we accept only `[HEADER ++ protocol]`.

 If the buffer contains an extra message `[HEADER ++ protocol ++ first_message/handshake]` because smoldot/other impl decides to send us optimistically an extra packet to avoid excessive flushing, then we'd lose the trailing bytes. 

We should have at least an warning to easily detect if the remaining bytes are non empty

src/multistream_select/protocol.rs
```diff
@@ -228,30 +246,53 @@ impl Message {
...
+            let header_varint_len =
+                unsigned_varint::encode::usize(header_len, &mut varint_buf).len();
+            header_varint_len + header_len + msg_varint_len + msg_len
+        } else {
+            msg_varint_len + msg_len
+        }
+    };
+
+    if capacity > super::length_delimited::MAX_FRAME_SIZE as usize {
+        tracing::debug!(
```
lexnv: I would even bump this to an WARN for the MVP while we keep the webrtc experimental to easily detect if this is an issue with smoldot over long term

src/transport/webrtc/connection.rs
```diff
@@ -411,23 +437,75 @@ impl WebRtcConnection {
...
+                        peer = ?self.peer,
+                        ?channel_id,
+                        "protocol rejected, trying next fallback",
+                    );
+
+                    let message = WebRtcMessage::encode(message, None);
+                    self.rtc
+                        .channel(channel_id)
+                        .ok_or(Error::ChannelDoesntExist)
+                        .map_err(|_| {
```
lexnv: nit: Here and below lets add traces for these errors since we are hiding them under `NegotiationError::Failed` it would be easier to debug if something goes wrong

src/transport/webrtc/connection.rs
```diff
@@ -411,23 +437,75 @@ impl WebRtcConnection {
...
+                        peer = ?self.peer,
+                        ?channel_id,
+                        "all protocols rejected by remote peer",
+                    );
+
+                    return Err(SubstreamError::NegotiationError(
+                        NegotiationError::Failed.into(),
+                    ));
+                }
+                Err(_) => {
```
lexnv: nit: Same here a trace could help

src/multistream_select/listener_select.rs
```diff
@@ -333,52 +332,122 @@ pub enum ListenerSelectResult {
...
+            return Err(Error::NegotiationError(
+                error::NegotiationError::MultistreamSelectError(NegotiationError::Failed),
+            )),
+    };
 
-    // skip the multistream-select header because it's not part of user protocols but verify it's
-    // present
-    if protocol_iter.next() != Some(PROTO_MULTISTREAM_1_0) {
+    // Reject messages with unexpected trailing data.
+    if !payload.is_empty() {
```
lexnv, line 434: nit: Lets add a log here so we can capture this better, otherwise itt would be a bit more difficult to debug an `InvalidData` that's comming from all the edge-cases :D

src/multistream_select/listener_select.rs
```diff
@@ -333,52 +332,122 @@ pub enum ListenerSelectResult {
...
+            if payload.is_empty() {
+                // Header only — echo the exact received bytes back (zero alloc).
+                return Ok(ListenerSelectResult::PendingProtocol {
+                    message: raw_payload,
+                });
+            }
+            // Header + protocol in same payload.
+            match decode_multistream_message(&mut payload)? {
+                Message::Protocol(protocol) => (protocol, true),
+                _ =>
```
lexnv: nit: Same here, lets match this / add a few more trace/debug logs

src/multistream_select/dialer_select.rs
```diff
@@ -353,69 +364,86 @@ impl WebRtcDialerState {
...
+            match (&self.state, message) {
+                (HandshakeState::WaitingResponse, Ok(Message::Header(HeaderLine::V1))) => {
+                    self.state = HandshakeState::WaitingProtocol;
+                }
+                (HandshakeState::WaitingResponse, Ok(Message::Protocol(_))) => {
+                    return Err(crate::error::NegotiationError::MultistreamSelectError(
+                        NegotiationError::Failed,
+                    ));
+                }
+                (_, Ok(Message::NotAvailable)) => {
```
lexnv: This could be a protocol violation if the header was not received.

The spec says that for the first message the header should be echoed back:

```
> /multistream/1.0.0\n/ipfs/kad/1.0.0\n
< /multistream/1.0.0\nna\n
```

However, we currently support missing the header:

```
> /multistream/1.0.0\n/ipfs/kad/1.0.0\n
< na\n
```

Which will make us propose the next fallback protocol that will most definetely not be accepted since we are still in the `HandshakeState::WaitingResponse` state.

In case the our state is waiting protocol this is fine. However, if we are still waiting the header we should return a state missmatch / error


```rust
// This is not oki, we ahve not received the headr
(HandshakeState::WaitingResponse, Ok(Message::NotAvailable)) => return Ok(HandshakeResult::StateMismatch),
(HandshakeState::WaitingProtocol, Ok(Message::NotAvailable)) => return Ok(HandshakeResult::Rejected),
```

Lets also add a bit more trace logs / debugs here especially for the `HandshakeResult::Failed` cases

src/multistream_select/dialer_select.rs
```diff
@@ -434,11 +462,16 @@ impl WebRtcDialerState {
                         NegotiationError::Failed,
                     ));
                 }
-                (HandshakeState::WaitingProtocol, None) => {
-                    return Ok(HandshakeResult::NotReady);
+                _ => {
```
lexnv, line 477: nit: We could also receive an `Message::ListProtocols`

src/multistream_select/dialer_select.rs
```diff
@@ -447,7 +480,7 @@ mod tests {
     use super::*;
     use crate::multistream_select::{listener_select_proto, protocol::MSG_MULTISTREAM_1_0};
     use bytes::BufMut;
-    use std::time::Duration;
+    use std::{sync::OnceState, time::Duration};
```
lexnv: nit: Is `OnceState` used anywhere?

src/transport/webrtc/connection.rs
lexnv: Have created a followup for this: https://github.com/paritytech/litep2p/issues/582

src/multistream_select/dialer_select.rs
gab8i: Totally fair! I was used to this approach of explain their safety but from now on I will try to avoid them!

gab8i: Now that I'm rewriting it I don't see any reason for that unwrap, probably a re-work which lead me to this weird solution.

src/multistream_select/dialer_select.rs
gab8i: manual slicing is now completely substituted by: 

```rust
            remaining.advance(len_size);
            let payload = remaining.split_to(len);
```

lexnv:
We are getting closer here! Thanks again @gab8i for tackling this! 🙏 Will approve the PR as soon as we tackle https://github.com/paritytech/litep2p/pull/573#discussion_r3241271759 and above nits to improve debuggability 


Have created two followups not to bloat the scope of this PR:
- https://github.com/paritytech/litep2p/issues/580
- https://github.com/paritytech/litep2p/issues/582


Have added the issues to our webrtc milestone as well for better visibility:
- https://github.com/paritytech/litep2p/issues/312

src/multistream_select/dialer_select.rs
gab8i: Nice catch! This needs to be addressed to avoid any protocol issue! Solution should be pretty basic:

 ```rust
                 (HandshakeState::WaitingResponse, Ok(msg)) => {
                    tracing::trace!(
                      target: LOG_TARGET,
                      ?msg,
                      "Expected header response from peer, got different message"
                    );
                    return Err(crate::error::NegotiationError::MultistreamSelectError(
                        NegotiationError::Failed,
                    ));
                }
 ```

src/multistream_select/dialer_select.rs
gab8i: Making this function more reliable would imply to change quite a lot the code, but what you describe could totally be implemented by a peer which can send those 3 things out at once. Right now it is only expected to interact with smoldot which seems to not do so. 

I've pushed a change which warn if trailing bytes are found after having decoded valid multistream-select messages. Is this ok for now?

lexnv: Yep warning should be sufficient for now since the fix is quite intrusive. Could you create an issue to not forget about it 🙏 ?

src/multistream_select/dialer_select.rs
```diff
@@ -353,79 +364,110 @@ impl WebRtcDialerState {
...
+                target: LOG_TARGET,
+                ?message,
+                "Decoded message while registering response",
+            );
+
+            let check_trailing_bytes = |bytes: &Bytes| {
+                // remote may be optimistically pipelining the first application frame
+                if !bytes.is_empty() {
+                    tracing::warn!(
+                        ?bytes,
```
lexnv: nit: We could print the len instead of showing random wire messages


```suggestion
                        bytes_len=bytes.len(),
```

lexnv (approved):
Nice job here! 🚀

src/multistream_select/dialer_select.rs
gab8i: oh yes, my bad!!

src/multistream_select/dialer_select.rs
```diff
@@ -353,79 +364,110 @@ impl WebRtcDialerState {
...
                     if protocol == PROTO_MULTISTREAM_1_0 {
                         return Err(crate::error::NegotiationError::StateMismatch);
                     }
 
                     if self.protocol.as_bytes() == protocol.as_ref() {
+                        check_trailing_bytes(&remaining);
                         return Ok(HandshakeResult::Succeeded(self.protocol.clone()));
                     }
 
                     for fallback in &self.fallback_names {
```
skunert: Is it correct to accept fallback protocols here? If we offered only `self.protocol` but a peer would answer with one of our fallback protocols, we would accept. But indeed we should first offer the fallback right?

test that fails:
```rust
    #[test]
    fn reject_unproposed_fallback_confirmation() {
        let (mut dialer_state, _message) = WebRtcDialerState::propose(
            ProtocolName::from("/13371338/proto/1"),
            vec![ProtocolName::from("/sup/proto/1")],
        )
        .unwrap();

        // The dialer has only proposed the main protocol. The fallback is stored for a
        // later round and must not be accepted until `propose_next_fallback()` sends it.
        let mut response = BytesMut::with_capacity(64);
        response.put_u8(MSG_MULTISTREAM_1_0.len() as u8);
        Message::Header(HeaderLine::V1).encode(&mut response).unwrap();

        let fallback = Protocol::try_from(&b"/sup/proto/1"[..]).expect("valid protocol name");
        response.put_u8((fallback.as_ref().len() + 1) as u8);
        Message::Protocol(fallback).encode(&mut response).unwrap();

        match dialer_state.register_response(response.freeze().to_vec()) {
            Err(error::NegotiationError::MultistreamSelectError(NegotiationError::Failed)) => {}
            event => panic!("expected unproposed fallback to be rejected, got: {event:?}"),
        }
    }

```

src/transport/webrtc/connection.rs
skunert, line 311: Is it fine to just remove here? If the channel state is `ChannelState::OutboundOpening`, we would need to call ` report_substream_open_failure`.

src/multistream_select/dialer_select.rs
gab8i: You are totally right! The fix is to remember which one is the last proposed one and only succeed if the response contains it!

gab8i: After looking at the code, it is just a matter of removing fallback_protocols from being used to accept or not the registered response!

src/transport/webrtc/connection.rs
gab8i: Thanks! This was already addressed within [this](https://github.com/paritytech/litep2p/commit/1071581bfbd4979e27d97f6ef7c8184cabd31f94) commit, part of #586!

src/multistream_select/dialer_select.rs
```diff
@@ -353,92 +364,126 @@ impl WebRtcDialerState {
...
+
+        self.protocol = next;
+
+        let message = webrtc_encode_multistream_message(
+            Message::Protocol(
+                Protocol::try_from(self.protocol.as_ref()).map_err(|_| Error::InvalidData)?,
+            ),
+            false,
+        )?
+        .freeze()
```
lexnv, line 384: nit: This and above `?` operator alters the state of `fallback_names` which we have just popped. I think this is ok since we terminate the substream immediately and return negotiation-error?

src/transport/webrtc/connection.rs
```diff
@@ -263,7 +267,12 @@ impl WebRtcConnection {
                 "inbound channel opened, wait for `multistream-select` message",
             );
 
-            self.channels.insert(channel_id, ChannelState::InboundOpening);
+            self.channels.insert(
```
lexnv, line 270: We could get into a situation where the remote (smoldot) contains a bug and never sends more data after the `header` in the negotiation phase.

We could probably add a `SUBSTREAM_OPEN_TIMEOUT` similar to TCP to ensure that the negotiation phase is bounded and cannot leak channels until the connection is closed. In this case if we reach the timeout, we could simply propagate back the substream failed to installed protocols

src/multistream_select/dialer_select.rs
gab8i: Yes, it works because if `propose_next_fallback` fails then also `on_outbound_opening_channel_data` fails and the data channel is getting closed!

src/transport/webrtc/connection.rs
gab8i: Yes, that's definitely a bug where both inbound and outbound opened channels can hang forever waiting for the protocol handshake. I will open an issue to address this in a follow up PR!

skunert (approved):
Looks good

PR #574, opened 2026-05-13
fixes(webrtc): webrtc substream/ICE issues + request-response close ordering

gab8i:
Same as #573, this PR wraps work already done in https://github.com/ChainSafe/litep2p/pull/17.

Some of the features were not taken from the original PR and will be part of following PRs.

### DIFF
--- a/src/protocol/notification/negotiation.rs
+++ b/src/protocol/notification/negotiation.rs
@@ -79,6 +79,7 @@ pub enum HandshakeEvent {
 }
 
 /// Outbound substream's handshake state
+#[derive(Debug)]
 enum HandshakeState {
     /// Send handshake to remote peer.
     SendHandshake,
@@ -218,6 +219,13 @@ impl Stream for HandshakeService {
             inner.substreams.iter_mut()
         {
             if let Poll::Ready(()) = timer.poll_unpin(cx) {
+                tracing::trace!(
+                    target: LOG_TARGET,
+                    ?peer,
+                    ?direction,
+                    ?state,
+                    "handshake negotiation timed out",
+                );
                 return Poll::Ready(Some((
                     *peer,
                     HandshakeEvent::NegotiationError {
@@ -285,10 +293,36 @@ impl Stream for HandshakeService {
                     },
                     HandshakeState::ReadHandshake => match pinned.poll_next(cx) {
                         Poll::Ready(Some(Ok(handshake))) => {
+                            tracing::trace!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                handshake_len = handshake.len(),
+                                "successfully read handshake from substream",
+                            );
                             inner.ready.push_back((*peer, *direction, handshake.freeze().into()));
                             continue 'outer;
                         }
-                        Poll::Ready(Some(Err(_))) | Poll::Ready(None) => {
+                        Poll::Ready(Some(Err(error))) => {
+                            tracing::trace!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                ?error,
+                                "error reading handshake from substream",
+                            );
+                            return Poll::Ready(Some((
+                                *peer,
+                                HandshakeEvent::NegotiationError {
+                                    peer: *peer,
+                                    direction: *direction,
+                                },
+                            )));
+                        }
+                        Poll::Ready(None) => {
+                            tracing::trace!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                "substream closed while reading handshake",
+                            );
                             return Poll::Ready(Some((
                                 *peer,
                                 HandshakeEvent::NegotiationError {
--- a/src/protocol/request_response/mod.rs
+++ b/src/protocol/request_response/mod.rs
@@ -543,9 +543,12 @@ impl RequestResponseProtocol {
                             ?request_id,
                             "timed out while sending response",
                         ),
-                        Ok(Ok(_)) => feedback.take().map_or((), |feedback| {
-                            let _ = feedback.send(());
-                        }),
+                        Ok(Ok(_)) => {
+                            let _ = substream.close().await;
+                            if let Some(feedback) = feedback.take() {
+                                let _ = feedback.send(());
+                            }
+                        }
                         Ok(Err(error)) => tracing::trace!(
                         target: LOG_TARGET,
                             ?peer,
--- a/src/transport/webrtc/substream.rs
+++ b/src/transport/webrtc/substream.rs
@@ -171,6 +171,11 @@ impl SubstreamHandle {
         // This ensures that if a FIN message contains data, we deliver it before closing.
         if let Some(payload) = message.payload {
             if !payload.is_empty() {
+                tracing::trace!(
+                    target: "litep2p::webrtc::substream",
+                    payload_len = payload.len(),
+                    "forwarding payload to substream",
+                );
                 self.inbound_tx
                     .send(Event::Message {
                         payload,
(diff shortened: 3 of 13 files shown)


src/transport/webrtc/substream.rs
```diff
@@ -171,12 +182,28 @@ impl SubstreamHandle {
         // This ensures that if a FIN message contains data, we deliver it before closing.
         if let Some(payload) = message.payload {
             if !payload.is_empty() {
-                self.inbound_tx
+                tracing::trace!(
+                    target: "litep2p::webrtc::substream",
```
lexnv, line 186: nit: Let's add a LOG_TARGET constant for these. Similarly, I think we should keep the number of groups to a maximum of 2: grafana is only able to parse target ("litep2p") and subtarget ("webrtc") effectively ignoring the substream.

src/protocol/notification/negotiation.rs
```diff
@@ -285,10 +293,36 @@ impl Stream for HandshakeService {
...
                             inner.ready.push_back((*peer, *direction, handshake.freeze().into()));
                             continue 'outer;
                         }
-                        Poll::Ready(Some(Err(_))) | Poll::Ready(None) => {
+                        Poll::Ready(Some(Err(error))) => {
+                            tracing::trace!(
+                                target: LOG_TARGET,
+                                ?peer,
+                                ?error,
+                                "error reading handshake from substream",
```
lexnv, line 310: Nice ones! We should have now visibility into issues!

src/protocol/request_response/mod.rs
```diff
@@ -543,9 +543,12 @@ impl RequestResponseProtocol {
                             ?request_id,
                             "timed out while sending response",
                         ),
-                        Ok(Ok(_)) => feedback.take().map_or((), |feedback| {
-                            let _ = feedback.send(());
-                        }),
+                        Ok(Ok(_)) => {
+                            let _ = substream.close().await;
```
lexnv, line 547: This could be a bit problematic and might leak the `pending_outbound_responses` future if `WebrtcSubstream::poll_shutdown` has a bug. This code path will effectively:
- flush the underlying data
- send FIN
- wait to receive FIN_ACK

To ensure the substream is closed gracefully. If the remote peer decides to not respond, we'll be relying on the internal FIN_ACK_TIMEOUT, which is also not matching go-libp2p.

https://github.com/libp2p/go-libp2p/blob/3bfc5492c5b765e906c196633c3ac44ef7c777ae/p2p/transport/webrtc/stream.go#L38-L40

For TCP the close method is calling `shutdown` syscall which should return instantly, same for WebSocket. Quic shouldn't wait for peer ACK (but its still experimental). 

A few tiny things we can improve here:
- I think we should move the `substream.close()` after notifying sending the feedback back the user
- Lets also capture in a trace the substream close result, it could indicate a bug further down the line
- We could add a tokio timeout to be extra safe in case our poll impl has some issues
- Let's double check if we should bump FIN_ACK_TIMEOUT from 5s to 10s to match go-libp2p

src/transport/webrtc/substream.rs
```diff
@@ -194,8 +221,18 @@ impl SubstreamHandle {
...
+                    // If the Substream is already dropped, this is expected for
+                    // request-response protocols where the handler writes the response
+                    // and drops before the FIN arrives. Fall through to still send FIN_ACK.
+                    if self.inbound_tx.send(Event::RecvClosed).await.is_err() {
+                        tracing::debug!(
+                            target: "litep2p::webrtc::substream",
+                            "substream already dropped, skipping RecvClosed notification",
+                        );
+                    } else {
+                        self.fin_delivered.store(true, std::sync::atomic::Ordering::SeqCst);
```
lexnv, line 234: Shouldn't the fin delivered be set only if the `Some(Flag::FinAck)` message below doesnt fail?

src/transport/webrtc/substream.rs
```diff
@@ -157,6 +159,15 @@ pub struct SubstreamHandle {
     /// Whether we've already sent RecvClosed to the inbound channel.
     /// Prevents duplicate RecvClosed events if multiple FIN messages are received.
     read_closed: std::sync::atomic::AtomicBool,
+
+    /// Whether the remote's FIN was successfully delivered to the Substream.
```
lexnv, line 163: This flag actually has the meaning of sending `RecvClosed` successfully into the `inbound_tx` channel.

Edge cases this doesnt capture and might set the flag wrongly: 
- If `RecvClosed` failed, we won't set the flag but might still submit FIN.
- Similarly if `RecvClosed` succeeds, but then sending FIN fails, the flag would be set on true.

src/transport/webrtc/substream.rs
```diff
@@ -194,8 +221,18 @@ impl SubstreamHandle {
                         return Ok(());
                     }
 
-                    // Received FIN from remote, close our read half
-                    self.inbound_tx.send(Event::RecvClosed).await?;
+                    // Notify the Substream that the remote closed its write half.
+                    // If the Substream is already dropped, this is expected for
+                    // request-response protocols where the handler writes the response
+                    // and drops before the FIN arrives. Fall through to still send FIN_ACK.
+                    if self.inbound_tx.send(Event::RecvClosed).await.is_err() {
```
lexnv, line 228: nit: If we have a bug in the substream implemetation or smoldot is slow to read this might delay/stall the `FinAck`

src/protocol/request_response/mod.rs
gab8i: Yes, the standard is 10s, so it needs to be updated. [js-libp2p](https://github.com/libp2p/js-libp2p/blob/94f875c3fd1269e21c13cd364ca56b35fc8c72bf/packages/transport-webrtc/src/constants.ts#L76-L80) also uses 10 seconds.

You brought up really important flaws that needs to be addressed, if that's ok for you I would also work on this in a following PR instead of adding more stuff to this one!

src/transport/webrtc/substream.rs
gab8i: The fin_delivered flag records whether a FIN received from the peer was successfully dispatched to the Substream task. It is consulted at teardown to decide whether the connection should be closed
abruptly or with the normal half-close.

This handling is needed for the request-response case: after writing the response, the Substream is dropped immediately, so when the peer's FIN arrives the dispatch to the (dropped) substream and fails.

The flag stays false, and the teardown path uses that to decide its next move: if dispatch failed, emit ResetStream to close abruptly, if dispatch succeeded, let the half-close proceed normally.

A problem that I just noted is: the peer is still sent FIN_ACK because the outbound channel survives the Substream drop via a cloned sender held by SubstreamHandle. Only the inbound dispatch fails.

Thinking more about it, I will now refactor and change its name and modify some, in my opinion, confusing comments. A better name should be: `fin_dispatched`, the fin flag arrived and it was properly dispatched to the substream task.

src/transport/webrtc/substream.rs
gab8i: No, I think it works differently. Given what I said in the other comment explaining this feature, my understanding is:
- If RecvClosed dispatch fails → keep fin_dispatched = false and emit ResetStream during teardown.
- If RecvClosed dispatch succeeds → set fin_dispatched = true and let the graceful half-close proceed during teardown. 

The problem of delivering FIN_ACK in both cases remains, I will try to find which is the best way to refactor to solve any problem of sending both FIN_ACK and ResetStream.

gab8i:
Closing in favor of #586

PR #575, opened 2026-05-13
feat(webrtc): producer backpressure

gab8i:
Sctp data channels have a limit over the amount of messages that can be saved within its internal buffer before sending them. Writes fail if there is no more space, messages needs to be saved as pending and sent again as soon as there is avaiable space.

### DIFF
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -62,7 +62,7 @@ const LOG_TARGET: &str = "litep2p::webrtc::connection";
 /// Threshold under which str0m emits Event::ChannelBufferedAmountLow.
 const BACKPRESSURE_THRESHOLD: usize = 16 * (1 << 10); // 16 KB
 
-/// Maximum number of pending message supported per channel.
+/// Maximum number of pending messages supported per channel.
 const MAX_PENDING_PER_CHANNEL: usize = 16;
 
 /// Opening channel context.
@@ -95,7 +95,7 @@ struct SubstreamHandleSet {
     /// Substream handles.
     handles: IndexMap<ChannelId, SubstreamHandle>,
 
-    /// Substream which has pending messages.
+    /// Substreams that have pending messages.
     pending: HashSet<ChannelId>,
 
     /// Waker used to drive the stream when no handle can make progress.
@@ -130,12 +130,12 @@ impl SubstreamHandleSet {
         self.handles.swap_remove(key)
     }
 
-    /// Add a channel to the ones which has pending messages.
+    /// Mark channel as having pending messages.
     pub fn add_pending(&mut self, key: ChannelId) {
         self.pending.insert(key);
     }
 
-    /// Remove the channel from the ones with pending messages.
+    /// Unmark channel as having pending messages.
     pub fn clear_pending(&mut self, key: &ChannelId) {
         if self.pending.remove(key) {
             self.waker.wake();
@@ -396,7 +396,7 @@ impl WebRtcConnection {
                     peer = ?peer,
                     ?channel_id,
                     ?e,
-                     "failed to write multistream-select fallback proposal",
+                    "failed to write message to webrtc channel",
                 );
                 Err(Error::WebRtc(e))
             }
@@ -404,7 +404,7 @@ impl WebRtcConnection {
     }
 
     // Attempt to write all pending messages of the specified ChannelId.
-    // Returns either all messages has been sent or not.
+    // Returns whether all messages have been sent or not.
     fn write_pending(&mut self, channel_id: ChannelId) -> Result<bool, Error> {
         let Some(mut channel) = self.rtc.channel(channel_id) else {
             tracing::trace!(
@@ -432,12 +432,12 @@ impl WebRtcConnection {
             };
 
             let succeeded = Self::channel_write(&mut channel, channel_id, message, self.peer)?;
-
-            match self.pending_messages.get_mut(&channel_id) {
-                Some(messages) if succeeded => {
-                    messages.pop_front();
-                }
-                _ => break Ok(false),
+            if succeeded {
+                self.pending_messages
+                    .get_mut(&channel_id)
+                    .and_then(|messages| messages.pop_front());
+            } else {
+                break Ok(false);
             }
         }
     }
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -38,7 +38,7 @@ use crate::{
     PeerId,
 };
 
-use futures::{Stream, StreamExt};
+use futures::{task::AtomicWaker, Stream, StreamExt};
 use indexmap::IndexMap;
 use str0m::{
     channel::{ChannelConfig, ChannelId},
@@ -94,6 +94,9 @@ struct SubstreamHandleSet {
 
     /// Substream which has pending messages.
     pending: HashSet<ChannelId>,
+
+    /// Waker used to drive the stream when no handle can make progress.
+    waker: AtomicWaker,
 }
 
 impl SubstreamHandleSet {
@@ -103,6 +106,7 @@ impl SubstreamHandleSet {
             index: 0usize,
             handles: IndexMap::new(),
             pending: HashSet::new(),
+            waker: AtomicWaker::new(),
         }
     }
 
@@ -114,6 +118,7 @@ impl SubstreamHandleSet {
     /// Insert new handle to [`SubstreamHandleSet`].
     pub fn insert(&mut self, key: ChannelId, handle: SubstreamHandle) {
         assert!(self.handles.insert(key, handle).is_none());
+        self.waker.wake();
     }
 
     /// Remove handle from [`SubstreamHandleSet`].
@@ -129,7 +134,9 @@ impl SubstreamHandleSet {
 
     /// Remove the channel from the ones with pending messages.
     pub fn clear_pending(&mut self, key: &ChannelId) {
-        self.pending.remove(key);
+        if self.pending.remove(key) {
+            self.waker.wake();
+        }
     }
 }
 
@@ -138,7 +145,10 @@ impl Stream for SubstreamHandleSet {
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let len = match self.handles.len() {
-            0 => return Poll::Pending,
+            0 => {
+                self.waker.register(cx.waker());
+                return Poll::Pending;
+            }
             len => len,
         };
         let start_index = self.index;
@@ -159,6 +169,7 @@ impl Stream for SubstreamHandleSet {
             }
 
             if self.index == start_index + len {
+                self.waker.register(cx.waker());
                 break Poll::Pending;
             }
         }
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -242,7 +242,8 @@ pub struct WebRtcConnection {
     /// Pending outbound channels.
     pending_outbound: HashMap<ChannelId, ChannelContext>,
 
-    /// Pending outbound messages, at most one per channel.
+    /// Pending outbound messages,
+    /// at most [`MAX_PENDING_PER_CHANNEL`] per channel.
     pending_messages: HashMap<ChannelId, VecDeque<Vec<u8>>>,
 
     /// Open channels.
@@ -342,7 +343,7 @@ impl WebRtcConnection {
     // Attempt to write a message over the specified channel,
     // save the message as pending if `WebRtcConnection` didn't have
     // enough space.
-    fn write(&mut self, channel_id: ChannelId, message: Vec<u8>) -> Result<bool, Error> {
+    fn write(&mut self, channel_id: ChannelId, message: Vec<u8>) -> Result<(), Error> {
         let Some(mut channel) = self.rtc.channel(channel_id) else {
             tracing::trace!(
                 target: LOG_TARGET,
@@ -360,7 +361,7 @@ impl WebRtcConnection {
                 }
 
                 messages.push_back(message);
-                return Ok(false);
+                return Ok(());
             }
             _ => (),
         }
@@ -375,10 +376,10 @@ impl WebRtcConnection {
 
             pending_messages.push_back(message);
             self.handles.add_pending(channel_id);
-            return Ok(false);
+            return Ok(());
         }
 
-        Ok(true)
+        Ok(())
     }
 
     fn channel_write(
@@ -873,7 +874,7 @@ impl WebRtcConnection {
         );
 
         let message = WebRtcMessage::encode(data, flag);
-        self.write(channel_id, message).map(|_| ())
+        self.write(channel_id, message)
     }
 
     /// Open outbound substream.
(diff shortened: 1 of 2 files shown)


src/transport/webrtc/connection.rs
```diff
@@ -912,6 +960,14 @@ impl WebRtcConnection {
 
                         continue;
                     }
+                    Event::ChannelBufferedAmountLow(_channel_id) => {
+                        let channel_ids: Vec<_> =
+                            self.pending_messages.keys().into_iter().cloned().collect();
+                        for channel_id in channel_ids {
+                            let _ = self.write_pending(channel_id);
```
lexnv, line 1015: Shouldn't we retry just the channel ID that causd the buffer event?

src/transport/webrtc/connection.rs
```diff
@@ -276,16 +303,16 @@ impl WebRtcConnection {
             return Ok(());
         };
```
lexnv, line 324: If we return above it means we handled the inbound substream, however we are not setting the `set_buffered_amount_low_threshold` limit on the channel

src/transport/webrtc/connection.rs
```diff
@@ -129,10 +147,15 @@ impl Stream for SubstreamHandleSet {
...
+
+            if !self.pending.contains(&key) {
+                let (key, stream) = self.handles.get_index_mut(index).expect("handle to exist");
+                match stream.poll_next_unpin(cx) {
+                    Poll::Pending => {}
+                    Poll::Ready(event) => return Poll::Ready(Some((*key, event))),
+                }
             }
 
             if self.index == start_index + len {
```
lexnv, line 174: If all substreams are in the pending state, meaning we never called `stream.poll_next_unpin`, then no context waker is registered.

For that case, we should register the waker here before returning Pending. Then if we had a waker registered, during the `clear_pending` we should wake it as well.

Otherwise this `Stream` implementation would remain stuck. I think this is working atm because we rely on the `Event::ChannelBufferedAmountLow` event to fire inside the tokio::select loop which would then re call into `poll_next`

src/transport/webrtc/connection.rs
gab8i: Yes and no. It makes sense what you just described, and that's also what I did until I discovered that there is a shared write buffer (which has a 128 KiB cap) and then each channel has an amount of used bytes within that buffer. Each channel has its own counter, so what's the problem:

Channel A fills up most of the shared 128 KiB budget
Channel B calls write(), gets Ok(false)
B's buffered amount stays at 0 (or less than the specified threshold)
A gets ChannelBufferedAmountLow
B never does

Because internally ChannelBufferedAmountLow is triggered only if there is a transition of a channel's buffered amount going from above the threshold to below the threshold.

src/transport/webrtc/connection.rs
```diff
@@ -276,16 +303,16 @@ impl WebRtcConnection {
             return Ok(());
         };
 
+        if let Some(mut channel) = self.rtc.channel(channel_id) {
+            channel.set_buffered_amount_low_threshold(BACKPRESSURE_THRESHOLD);
```
lexnv: dq: If we bump this to 2 MiB, would we still be limited by anything?

src/transport/webrtc/connection.rs
```diff
@@ -205,6 +228,9 @@ pub struct WebRtcConnection {
     /// Pending outbound channels.
     pending_outbound: HashMap<ChannelId, ChannelContext>,
 
+    /// Pending outboud messages, at most one per channel.
```
lexnv:
```suggestion
    /// Pending outbound messages, at most one per channel.
```

src/transport/webrtc/connection.rs
```diff
@@ -298,6 +325,59 @@ impl WebRtcConnection {
         Ok(())
     }
 
+    // Attempt to write a message over the specified channel,
+    // save the message as pening if `WebRtcConnection` didn't have
```
lexnv:
```suggestion
    // save the message as pending if `WebRtcConnection` didn't have
```

src/transport/webrtc/connection.rs
```diff
@@ -298,6 +325,59 @@ impl WebRtcConnection {
...
+
+        Ok(true)
+    }
+
+    // Attempt to write all pending messages of the specified ChannelId.
+    // Returns either all messages has been sent or not.
+    fn write_pending(&mut self, channel_id: ChannelId) -> Result<bool, Error> {
+        let Some(message) = self.pending_messages.remove(&channel_id) else {
+            // This should never happen, `write_pending` should be called
+            // for only chennel with pending messages. Treat as a no-op
```
lexnv:
```suggestion
            // for only the channel with pending messages. Treat as a no-op
```

src/transport/webrtc/connection.rs
gab8i: This threshold is only used by str0m to emit a `ChannelBufferedAmountLow` event per channel. 
Each channel keeps its own `buffered_amount` counter tracking how many bytes it currently has queued, but all channels draw from the same shared 128 KiB write buffer in str0m. The event only fires on the transition from above the threshold to below it. So if we bump the threshold to an high value, no single channel can ever accumulate that much, so the transition never occurs, the event is never emitted.

src/transport/webrtc/connection.rs
```diff
@@ -893,6 +1006,13 @@ impl WebRtcConnection {
 
                         continue;
                     }
+                    Event::ChannelBufferedAmountLow(_channel_id) => {
+                        let channel_ids: Vec<_> = self.pending_messages.keys().cloned().collect();
+                        for channel_id in channel_ids {
```
lexnv, line 1014: Should we close the channel ID if we cant write data via `channel_write`?

src/transport/webrtc/connection.rs
```diff
@@ -959,6 +1079,9 @@ impl WebRtcConnection {
                                 ?error,
                                 "failed to send data to remote peer",
                             );
+
+                            self.channels.insert(channel_id, ChannelState::Closing);
+                            self.rtc.direct_api().close_data_channel(channel_id);
```
lexnv, line 1087: On erorrs here we rely on the `ChannelClose` to clean pending data / remove the channel id from the handles?

src/transport/webrtc/connection.rs
```diff
@@ -205,6 +242,9 @@ pub struct WebRtcConnection {
     /// Pending outbound channels.
     pending_outbound: HashMap<ChannelId, ChannelContext>,
 
+    /// Pending outbound messages, at most one per channel.
```
lexnv: Nit: at most `MAX_PENDING_PER_CHANNEL`

src/transport/webrtc/connection.rs
```diff
@@ -234,6 +274,7 @@ impl WebRtcConnection {
             endpoint,
             dgram_rx,
             pending_outbound: HashMap::new(),
+            pending_messages: HashMap::new(),
```
lexnv, line 278: nit: `HashMap::with_capacity(MAX_PENDING_PER_CHANNEL)`

src/transport/webrtc/connection.rs
lexnv: Could it happen that all channels are blow the capacity and `Event::ChannelBufferedAmountLow` never unblocks channels? 

- A has 17 KiB budget
- B has 5KiB with no pending messages to write
- A writes 1KiB and trigers the  `ChannelBufferedAmountLow` with no channel having messages to write


I think in this case we still rely on `MAX_PENDING_PER_CHANNEL` to kill the subtreams or on the keep-alive to terminate the connection

src/transport/webrtc/connection.rs
```diff
@@ -755,12 +872,8 @@ impl WebRtcConnection {
...
         );
 
-        self.rtc
-            .channel(channel_id)
-            .ok_or(Error::ChannelDoesntExist)?
-            .write(true, WebRtcMessage::encode(data, flag).as_ref())
-            .map_err(Error::WebRtc)
-            .map(|_| ())
+        let message = WebRtcMessage::encode(data, flag);
+        self.write(channel_id, message).map(|_| ())
```
lexnv: nit: This was here before, but why aren't we using the write's result?

lexnv (approved):
LGTM! 

Feel free to address pending comments as follow-ups, this is a good improvement for not silently dropping packets by introducing the pending messages queue. Together with the capacity cleanup on the backpressure threshold, we should be in better shape 🙏 

Nicely done here, we can improve things later if we discover compatibility issues/connection stalls 👍

src/transport/webrtc/connection.rs
gab8i: Why? It could happen that there is X amount of space available and most of the channels have pending messages > X then no one will succeeded beside one maybe

src/transport/webrtc/connection.rs
gab8i: Yes, str0m emits `ChannelClose` after `close_data_channel` completes! It would be worth adding a comment explaining this non obvious pattern

src/transport/webrtc/connection.rs
gab8i: Thanks!

src/transport/webrtc/connection.rs
gab8i: I'm not sure it would be really helpful, the hash map is a Key=ChannelId and Value=Queue of pending messages, only the latter, the queue, has a cap of MAX_PENDING_PER_CHANNEL.

The `field` pending messages should be empty most of the time so that iterating the keys is as close as possible to a no-op.

src/transport/webrtc/connection.rs
gab8i: Write result is something which was needed in the previous version to re-use the same write function also when pending messages were written. Now with the introduction of `write_channel` the returned boolean of the `write` function isn't used anywhere. I will remove it!

src/transport/webrtc/connection.rs
```diff
@@ -59,6 +59,12 @@ use std::{
 /// Logging target for the file.
 const LOG_TARGET: &str = "litep2p::webrtc::connection";
 
+/// Threshold under which str0m emits Event::ChannelBufferedAmountLow.
+const BACKPRESSURE_THRESHOLD: usize = 16 * (1 << 10); // 16 KB
+
+/// Maximum number of pending message supported per channel.
```
```suggestion
/// Maximum number of pending messages supported per channel.
```

src/transport/webrtc/connection.rs
```diff
@@ -88,6 +94,12 @@ struct SubstreamHandleSet {
 
     /// Substream handles.
     handles: IndexMap<ChannelId, SubstreamHandle>,
+
+    /// Substream which has pending messages.
```
```suggestion
    /// Substreams that have pending messages.
```

src/transport/webrtc/connection.rs
```diff
@@ -107,11 +121,25 @@ impl SubstreamHandleSet {
...
     }
 
     /// Remove handle from [`SubstreamHandleSet`].
     pub fn remove(&mut self, key: &ChannelId) -> Option<SubstreamHandle> {
-        self.handles.shift_remove(key)
+        self.pending.remove(key);
+        self.handles.swap_remove(key)
+    }
+
+    /// Add a channel to the ones which has pending messages.
```
```suggestion
    /// Mark channel as having pending messages.
```

src/transport/webrtc/connection.rs
```diff
@@ -107,11 +121,25 @@ impl SubstreamHandleSet {
...
+    pub fn add_pending(&mut self, key: ChannelId) {
+        self.pending.insert(key);
+    }
+
+    /// Remove the channel from the ones with pending messages.
```
```suggestion
    /// Unmark channel as having pending messages.
```

src/transport/webrtc/connection.rs
```diff
@@ -298,6 +340,108 @@ impl WebRtcConnection {
...
+    ) -> Result<bool, Error> {
+        match channel.write(true, message) {
+            Ok(succeeded) => Ok(succeeded),
+            Err(e) => {
+                tracing::trace!(
+                    target: LOG_TARGET,
+                    peer = ?peer,
+                    ?channel_id,
+                    ?e,
+                     "failed to write multistream-select fallback proposal",
```
dmitry-markin: nit: can we use a better message?

src/transport/webrtc/connection.rs
```diff
@@ -298,6 +340,108 @@ impl WebRtcConnection {
...
+                );
+                Err(Error::WebRtc(e))
+            }
+        }
+    }
+
+    // Attempt to write all pending messages of the specified ChannelId.
+    // Returns either all messages has been sent or not.
```
```suggestion
    // Returns whether all messages have been sent or not.
```

src/transport/webrtc/connection.rs
```diff
@@ -298,6 +340,108 @@ impl WebRtcConnection {
...
+            };
+
+            let succeeded = Self::channel_write(&mut channel, channel_id, message, self.peer)?;
+
+            match self.pending_messages.get_mut(&channel_id) {
+                Some(messages) if succeeded => {
+                    messages.pop_front();
+                }
+                _ => break Ok(false),
+            }
```
dmitry-markin: nit: the logic seems inverted to me. It will be more clear to write `if succeeded { ... } else { ... }`.

src/transport/webrtc/opening.rs
```diff
@@ -209,12 +209,22 @@ impl OpeningWebRtcConnection {
...
             .channel(self.noise_channel_id)
             .ok_or(Error::ChannelDoesntExist)?
             .write(true, payload.as_slice())
             .map_err(Error::WebRtc)?;
 
+        if !succeeded {
+            tracing::error!(
+                target: LOG_TARGET,
+                connection_id = ?self.connection_id,
+                "noise channel full during initial handshake write"
```
dmitry-markin: It's better to log different messages so we know the place where it failed.

```suggestion
                "sending first noise handshake message failed: channel full"
```

src/transport/webrtc/opening.rs
```diff
@@ -305,9 +315,18 @@ impl OpeningWebRtcConnection {
...
 
-        channel.write(true, payload.as_slice()).map_err(Error::WebRtc)?;
-        self.rtc.direct_api().close_data_channel(self.noise_channel_id);
+        let succeeded = channel.write(true, payload.as_slice()).map_err(Error::WebRtc)?;
 
+        if !succeeded {
+            tracing::error!(
+                target: LOG_TARGET,
+                connection_id = ?self.connection_id,
+                "noise channel full during initial handshake write"
```
dmitry-markin:
```suggestion
                "sending second noise handshake message failed: channel full"
```

dmitry-markin (approved):
Looks good, even though I still don't understand all the interactions of the changes with the pre-existing WebRTC implementation.

src/transport/webrtc/connection.rs
gab8i: You are totally right, this log message ended up being completely out of place

src/transport/webrtc/connection.rs
gab8i: You are right, something like:

```rust
            if succeeded {
                self.pending_messages
                    .get_mut(&channel_id)
                    .and_then(|messages| messages.pop_front());
            } else {
                break Ok(false);
            }
```

It's definitely clearer, I was biased by the -1 LOC

PR #576, opened 2026-05-13
fix(webrtc): don't re-send multistream header

gab8i:
After the multistream-select header has been exchanged, only protocol and na messages should flow between peers. propose_next_fallback was incorrectly transitioning state back to WaitingResponse (expecting another header) and wrapping the outgoing protocol message with the header. Keep the state at WaitingProtocol and emit the protocol message without the header prefix.


gab8i:
Closing, now part of #573

PR #579, opened 2026-05-13
feat(webrtc): introduce `WebRtcListener` to support multiple listen addresses

gab8i:
webrtc module has been slightly refactored with the introduction of `AddressPair` and the `WebRtcLisner` structs to support multiple, or zero, listen addresses


src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,174 @@
...
+
+            // Each socket that returned Pending registered its waker,
+            // Err sockets did not but will re-register on the next poll.
+            if idx == self.next_listener {
+                return Poll::Pending;
+            }
+        }
+    }
+
+    /// Extract socket address and `PeerId`, if found, from `address`.
```
lexnv: We are not extracting the peer ID in this fn

gab8i: Oh yes, that's a vestigial from a function been moved and refactored, will fix this! Thanks!

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,174 @@
...
+            listen_multi_addresses.push(
+                Multiaddr::empty()
+                    .with(Protocol::from(listen_address.ip()))
+                    .with(Protocol::Udp(listen_address.port()))
+                    .with(Protocol::WebRTCDirect)
+                    .with(Protocol::Certhash(certificate)),
+            );
+        }
+
+        Ok((
```
lexnv, line 89: nit: We could return an error if there are no listener addresses available

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,174 @@
...
+            protocol => {
+                tracing::error!(
+                    target: LOG_TARGET,
+                    ?protocol,
+                    "invalid protocol, expected `WebRTCDirect`"
+                );
+                return Err(Error::AddressError(AddressError::InvalidProtocol));
+            }
+        }
+
```
lexnv, line 210: Should we also check that the address has a certificate attached?

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,174 @@
...
+                    }));
+                }
+                // All UdpSocket errors are transient and noone
+                // of them implies a complete shutdown of the socket.
+                // Log the error but do not tear down the WebRtc instance.
+                Poll::Ready(Err(e)) => tracing::warn!(
+                    target: LOG_TARGET,
+                    ?local,
+                    ?e,
+                    "failed to receive a datagram",
```
lexnv, line 145: nit: When a `Poll::Ready(Err())` is returned the waker will not be registered. If this happens for all sockets we'll effectively be stuck in pending

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,174 @@
...
+        certificate: Multihash<64>,
+    ) -> crate::Result<(Self, Vec<Multiaddr>)> {
+        let mut listen_multi_addresses = Vec::with_capacity(multiaddr_listen_addresses.len());
+        let mut listen_addresses = Vec::with_capacity(multiaddr_listen_addresses.len());
+
+        for listen_address in multiaddr_listen_addresses {
+            let listen_address = Self::get_socket_address(&listen_address)?;
+            let socket = if listen_address.is_ipv4() {
+                let socket = Socket::new(Domain::IPV4, Type::DGRAM, Some(socket2::Protocol::UDP))?;
+                socket.set_reuse_address(true)?;
```
lexnv: `set_reuse_address` was called after the `bind()` previously, any reason for changing this? Should we deduplicate the listen addresses now that we suport multiple ones?

src/transport/webrtc/listener.rs
gab8i: No, this is a listen address, so a certhash isn't required. 
I've added a comment noting that there is a check to ensure the address is a webrtc-direct multiaddr.

The DTLS certificate is generated once in `WebRtcTransport::new()` and presented to
dialing peers during the DTLS handshake, the certhash is the fingerprint of that certificate,
append when building the advertised multiaddr.

src/transport/webrtc/listener.rs
gab8i: `set_reuse_address` needs to be called before binding the socket, it sets `SO_REUSEADDR` flag. The same applies to `set_reuse_port`, which I just noticed is currently being called after the bind. This ordering is a Linux requirement.

The real question, though: do we actually want multiple WebRTC instances bound to the same address
 and rely on the kernel's load balancer? I don't think we should allow two listeners on the same
(ip, port). So I'd keep `set_reuse_address`, multiple listeners can coexist on the same host, but drop `set_reuse_port`, ensuring each listener has a unique local address.

dmitry-markin: We shouldn't set any of these. `SO_REUSEADDR` only makes sense when binding a TCP socket to a previously bound `ip:port` pair that was not cleanly shut down by previous process to not get `EADDRINUSE`. It doesn't make sense for UDP and was likely blindly copied from TCP/WS.

Agree that we shouldn't set `SO_REUSEPORT` either.

src/transport/webrtc/mod.rs
```diff
@@ -458,33 +407,19 @@ impl TransportBuilder for WebRtcTransport {
     where
         Self: Sized,
     {
+        if config.listen_addresses.is_empty() {
+            return Err(Error::Other(
+                "WebRTC transport requires at least one listen address but none were configured"
+                    .to_string(),
+            ));
+        }
```
dmitry-markin, line 424: It should be possible to enable WebRTC transport and not set any WebRTC listen addresses. Can you double check this won't result in an error — i.e., if there are no WebRTC listen addresses, we don't instantiate `WebRtcTransport` at all?

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,206 @@
...
+
+use super::AddressPair;
+use crate::{error::AddressError, Error};
+
+const LOG_TARGET: &str = "litep2p::webrtc::listener";
+
+/// WebRtc listener.
+pub(super) struct WebRtcListener {
+    /// Bound sockets paired with their local address.
+    listen_addresses: Vec<(SocketAddr, Arc<UdpSocket>)>,
```
dmitry-markin: nit: `listen_addresses` -> `listen_sockets`

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,206 @@
...
+        multiaddr_listen_addresses: Vec<Multiaddr>,
+        certificate: Multihash<64>,
+    ) -> crate::Result<(Self, Vec<Multiaddr>)> {
+        let mut listen_multi_addresses = Vec::with_capacity(multiaddr_listen_addresses.len());
+        let mut listen_addresses = Vec::with_capacity(multiaddr_listen_addresses.len());
+
+        let handle_multiaddr = |listen_address| -> crate::Result<(UdpSocket, SocketAddr)> {
+            let listen_address = Self::get_socket_address(&listen_address)?;
+            let socket = if listen_address.is_ipv4() {
+                Socket::new(Domain::IPV4, Type::DGRAM, Some(socket2::Protocol::UDP))?
```
dmitry-markin, line 49: What happens if the user passes multiple multiaddresses that resolve to the same ip:port? May be we need to handle this case to fail with a clear error message?

src/transport/webrtc/mod.rs
```diff
@@ -293,15 +240,18 @@ impl WebRtcTransport {
                 opening::WebRtcEvent::Transmit {
                     destination,
                     datagram,
-                } =>
-                    if let Err(error) = self.socket.try_send_to(&datagram, destination) {
+                } => {
+                    // UNWRAP: local addr originated from self.listener, must be present.
+                    let socket = self.listener.socket(&addrs.local).unwrap();
```
dmitry-markin: This is not obvious from this function code (i.e., `addrs` is passed to it and we also need to check the caller), so better to handle error or refactor so it's not possible.

src/transport/webrtc/mod.rs
```diff
@@ -493,29 +428,23 @@ impl TransportBuilder for WebRtcTransport {
...
-        let listen_multi_addresses = {
-            let fingerprint = crypto_provider.sha256_provider.sha256(&dtls_cert.certificate);
+        let fingerprint = crypto_provider.sha256_provider.sha256(&dtls_cert.certificate);
 
-            const MULTIHASH_SHA256_CODE: u64 = 0x12;
-            let certificate = Multihash::wrap(MULTIHASH_SHA256_CODE, &fingerprint)
-                .expect("fingerprint's len to be 32 bytes");
+        const MULTIHASH_SHA256_CODE: u64 = 0x12;
+        let certificate = Multihash::wrap(MULTIHASH_SHA256_CODE, &fingerprint)
+            .expect("fingerprint's len to be 32 bytes");
```
dmitry-markin: We can simplify this to:
```
let cert_hash = multihash_codetable::Code::Sha2_256::digest(&dtls_cert.certificate);
```
Also, `certificate` is strictly speaking either `hash` of `fingerprint`.

src/transport/webrtc/mod.rs
```diff
@@ -601,18 +529,18 @@ impl Transport for WebRtcTransport {
...
+            addrs,
             ConnectionContext {
                 tx,
                 peer,
                 connection_id,
             },
         );
 
+        // UNWRAP: local addr originated from this.listener, must be present.
+        let socket = self.listener.socket(&addrs.local).unwrap();
```
dmitry-markin: It's hardly traceable in this func, so better either handle errors or refactor.

src/transport/webrtc/mod.rs
```diff
@@ -697,106 +625,96 @@ impl Stream for WebRtcTransport {
...
-            .filter_map(|(source, mut delay)| match Pin::new(&mut delay).poll(cx) {
-                Poll::Pending => None,
-                Poll::Ready(_) => Some(*source),
-            })
-            .collect::<Vec<_>>()
-            .into_iter()
-            .filter_map(|source| {
-                let mut pending_event = None;
+            let nread = read_buf.filled().len();
+            buf.truncate(nread);
```
dmitry-markin, line 659: We should avoid allocating/truncating a buffer on every loop iteration. Instead we should reuse the same buffer and pass a slice/cheap reference to `on_socket_input`. This should be enough, because `DatagramRecv::try_from` in `on_socket_input` accepts a borrowed slice anyway.

src/transport/webrtc/mod.rs
```diff
@@ -697,106 +625,96 @@ impl Stream for WebRtcTransport {
...
 
-                loop {
-                    match this.poll_connection(&source) {
+            match this.on_socket_input(addrs, buf) {
+                Ok(false) => {}
+                Ok(true) => loop {
+                    match this.poll_connection(&addrs) {
```
dmitry-markin, line 664: I have a nagging feeling that regular/opening connection polling can be refactored. Currently, we handle opening connection polling as a special case by `on_socket_input` indicating we should manually poll it, and manually keeping track and polling the opening connections' timeouts. Instead, it should be possible to encapsulate this logic in the opening connection object, poll just it, and as a result of polling it can upgrade to a fully established connection. Definitely not for this PR, but worth considering as a follow up.

The name `poll_connection` also caused some confusion during the review, so may be let's at least rename it into `poll_opening_connection`, and also `this.timeouts` -> `this.opening_timeouts`. Actually, even a better name would be `opening_timers`, because these timeouts mostly indicate work to do, not that something has timed out (i.e., connection was unsuccessful).

dmitry-markin: One more thing: if we put opening connection objects into `FuturesUnordered`, it will automatically skip polling objects that didn't generate wakeup events, contrary to polling all timeouts in a loop as we do now.

dmitry-markin (approved):

dmitry-markin (approved):

dmitry-markin (approved):

src/transport/webrtc/mod.rs
gab8i: I've also talked with @lexnv about this. There is one main thing, which is the difference between the transport layers, TCP or WS can be created with 0 listen addresses because they still need to be able to dial another peer. WebRTC is different. Dialing in WebRTC is a no-op, thus it doesn't make sense to create a WebRTC instance without any listen addresses, not only because of the dialing, but also because it will be inserted in Substrate as 'experimental', and so if there is no listen address there shouldn't be any code path that touches WebRTC code, to avoid any problems.

This is to say that the same pattern of accepting 0 listen addresses cannot be used in WebRTC as it is in the other transports, and thus another approach should be taken. Something basic I can think of:
1. Instead of exposing `webrtc::Config` as public, we force the usage of a constructor that fails on empty listen_addresses.
2. We keep the failure within the WebRTC initialization and handle it gracefully within Substrate.

I think the first one is the easiest, also taking into consideration a future change, the implementation of the dialing side of WebRTC.

src/transport/webrtc/listener.rs
gab8i: Yes, thanks this absolutely needs to be handled! 

The same approach of a binding error could be used, thus just skipping the address which resolved to an error or to an already used socket, log + continue

gab8i: Oh, I just realized that this is already covered with the removal of `reuse_addr` and `reuse_port`, if two specified listen address resolve to the same ip:port the kernel will return an error while binding the address.

src/transport/webrtc/mod.rs
gab8i: Yes, a little refactor to avoid the panic is definitely the way to go! Thanks!

src/transport/webrtc/mod.rs
```diff
@@ -121,102 +123,47 @@ enum ConnectionEvent {
...
 }
 
+/// Endpoints of a received UDP datagram: which local socket received it,
+/// and the remote socket that sent it.
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+struct AddressPair {
+    /// Address of the local listening socket that received the datagram.
+    local: SocketAddr,
+    /// Address of the remote peer that sent the datagram.
+    source: SocketAddr,
```
lexnv, line 134: dq: Could we replacethis by connection ID?

src/transport/webrtc/mod.rs
gab8i: This has been addressed along with the other panic coming from non-existent addresses within the listener

src/transport/webrtc/mod.rs
gab8i: This is already been noticed by @lexnv and there is a live issue #588. As you said the solution is pretty easy, just a matter of sharing a single buffer within the `WebRtcTransport` struct, I already have a branch with the fix, I will create a pr soon!

src/transport/webrtc/mod.rs
gab8i: oh, nice, yes this simplify it by a lot and remove the needs for that ugly constant!

I'm not sure though how this relates with the usage of `str0m::crypto::from_feature_flags` which uses the specified feature flag within cargo to construct the correct crypto provider which should be aligned with the `multihash_codetable::Code` used

src/transport/webrtc/mod.rs
lexnv: Forcing a listen address different from 0.0.0.0 would guarantee that the sending UDP packets arrive on the same IP:port that it was received on.

For TCP this is not an issue, we'll bind to 0.0.0.0 then during the `accept` we ll receive a per connection socket. In UDP this becomes problematic on binding 0.0.0.0:
- peer A sent out a message on A-ip:port
- we provide the response, but since we listen on 0.0.0.0 the kernel can choose any interface to facilitate the transport, and might pick B-ip:port
- this will then cause issues with ICE / STUN:


>  The ICE agent MUST check that the source and destination transport
   addresses in the Binding request and response are symmetric.  That
   is, the source IP address and port of the response MUST be equal to
   the destination IP address and port to which the Binding request was
   sent
   
   https://datatracker.ietf.org/doc/html/rfc8445#section-7.2.5.2.1
   
   
I think we have a few options to make this work:
- https://man7.org/linux/man-pages/man2/IP_PKTINFO.2const.html
- If the address is unspecified, before binding the UDP socket translate it similar to TCP:

https://github.com/paritytech/litep2p/blob/4cb002ea99a2d8db76f8b47607caa3b5685ada2a/src/transport/common/listener.rs#L273-L282

Probably the most straight forward one since we are still experimental would be to: 
- return error if no addresses are provided
- return error/warn if an address tries to bind to 0.0.0.0


@dmitry-markin @gab8i would love to get your thoughts on this 🙏

src/transport/webrtc/listener.rs
```diff
@@ -0,0 +1,206 @@
...
+            let listen_address = socket.local_addr()?;
+            Ok((socket, listen_address))
+        };
+
+        for listen_address in multiaddr_listen_addresses {
+            let (socket, listen_address) = match handle_multiaddr(listen_address) {
+                Ok(res) => res,
+                Err(err) => {
+                    tracing::warn!(target: LOG_TARGET, ?err, "failed to bind listen address, skipping");
+                    continue;
```
lexnv: nit: Since the user provided these addresses explicitely we could also fial here

src/transport/webrtc/mod.rs
gab8i: I think that something like the following is not as clean as what you proposed but definitely better than before (no constant used and panic handled properly): 

```rust
        let fingerprint = crypto_provider.sha256_provider.sha256(&dtls_cert.certificate);
        let cert_hash = multihash_codetable::Code::Sha2_256.wrap(&fingerprint).map_err(|err| {
            tracing::warn!(target: LOG_TARGET, ?err, "failed to wrap WebRTC certificate");
            Error::Other("could not compute WebRTC certificate".to_string())
        })?;
```

But still re-using the same crypto provider

src/transport/webrtc/mod.rs
```diff
@@ -697,106 +625,96 @@ impl Stream for WebRtcTransport {
...
-                                Some(TransportEvent::ConnectionEstablished { peer, endpoint });
+                            this.pending_events.push_back(TransportEvent::ConnectionEstablished {
+                                peer,
+                                endpoint,
+                            });
                         }
                         ConnectionEvent::ConnectionClosed => {
-                            this.opening.remove(&source);
-                            return None;
+                            this.opening.remove(&addrs);
```
lexnv: Should we also remove at this point all entries from connections that reference this addr?

```rust
this.connections.retain(a != addr)
```

lexnv (approved):
LGTM!

There's an open question about tackling the listen address mode (https://github.com/paritytech/litep2p/pull/579#discussion_r3355498718), we could either integrate the changes here of followu 🙏

src/transport/webrtc/mod.rs
dmitry-markin: > I've also talked with @lexnv about this. There is one main thing, which is the difference between the transport layers, TCP or WS can be created with 0 listen addresses because they still need to be able to dial another peer.

You are right, I was mostly thinking about the case when we enable the WebRTC feature, but don't pass any addresses. I guess in this case no WebRtcTransport should be initialized at all, so there is no issue.

dmitry-markin: > * return error/warn if an address tries to bind to 0.0.0.0

It should be possible to bind to 0.0.0.0, and it typically is not an issue, because the source IP address of the outgoing packet is selected according to the routing rules. If the machine is part of multiple networks, the interface will be selected correctly if routing is set up correctly. The only exception is if the machine has multiple public interfaces / interfaces connected to the same network, but this situation is so niche that the administrators are expected to understand the consequences and configure the networking accordingly.

gab8i: Wow, this adds a lot more information and constraints into the picture. Yes I think that given the current phase and the uncertainty of how the substrate integration will look like (given also the work that needs to be done to address persistent certificates #605) it would be easier to do what you described, keeping thus litep2p kind of clean and then just filter things on the higher layer within substrate!

src/transport/webrtc/mod.rs
gab8i: What you described would make the implementation a lot cleaner and makes perfectly sense, I opened an issue to make sure this will be addressed in a follow-up 

https://github.com/paritytech/litep2p/issues/609

src/transport/webrtc/mod.rs
gab8i: I don't think this is possible, `ConnectionID` are used as keys within the `connections` map which has `AddressPair` part of the value.

src/transport/webrtc/listener.rs
gab8i: This would be fair yes, I will update the code to return an error in this case!

src/transport/webrtc/mod.rs
gab8i: Great, nice catch! `connections` clean up was also missing from another place!!

src/transport/webrtc/mod.rs
dmitry-markin: There is no need to reuse a crypto provider for sha256, `multihash_codetable` uses Rust sha2 implementation.

gab8i: Sorry, I completely miss this comment before merging, I will address this within the persistent dtls certificate PR!

PR #581, opened 2026-05-13
  deps: quinn 0.11 + rustls 0.23 + hickory-resolver 0.26 (closes #308)

### LINKED ISSUE #308: quic: Update quinn dependency to 0.11 from 0.9
The latest quinn crate includes several performance improvements and fixes.

Offhand it looks like the API changed quite a bit.

For a full changelog see:
- https://github.com/quinn-rs/quinn/releases

hitchho:
Closes #308. Bumps the three deps that gate four open RustSec/GHSA
advisories surfacing in every downstream consumer of litep2p:

  - quinn-proto GHSA-6xvm-j4wr-6v98 / CVE-2026-31812 (high)
  - ring GHSA-4p46-pwfr-66x6 / CVE-2025-4432 (medium)
  - hickory-proto GHSA-3v94-mw7p-v465 (high)
  - hickory-proto GHSA-q2qq-hmj6-3wpp (medium)

Split into two commits along the dep boundary; each compiles
independently:

  1. `deps: hickory-resolver 0.25 → 0.26`
  2. `deps: quinn 0.9 → 0.11, rustls 0.20 → 0.23`

quinn and rustls had to land together — quinn 0.11 wraps
`rustls::ClientConfig` / `ServerConfig` via `QuicClientConfig::try_from` /
`QuicServerConfig::try_from` which require rustls 0.23's pluggable
`CryptoProvider`.

The hickory bump deserves extra eyeballs because hickory 0.26 ships two
**silent breaking default-value changes** that compile cleanly but
break at runtime:

  - `ResolverConfig::default()` lost its `Default` impl (was Google DNS,
    now empty `name_servers`); fixed by explicit Cloudflare fallback.
  - `LookupIpStrategy::default()` flipped from `Ipv4thenIpv6` (sequential
    v4-first) to `Ipv6AndIpv4` (parallel v6-preferred). Left at hickory's
    new default; documented in the commit so consumers in v6-broken
    environments know what to expect.

The rustls bump touches the libp2p TLS cert verifier substantially.
Notable behavioural decisions in the commit:

  - `AlwaysResolvesCert` resolver replaces `with_single_cert` (rustls
    0.23 rejects the libp2p X.509 extension as `UnsupportedCriticalExtension`
    on the old path). Same pattern as rust-libp2p's `libp2p-tls` 0.6.
  - Peer-id mismatch returns
    `CertificateError::ApplicationVerificationFailure` (ships a
    `bad_certificate`/`access_denied` TLS alert) rather than the generic
    `Error::General`.
  - `kx_groups` pinned explicitly so a future ring release can't widen
    the curve set.

## Testing

  - `cargo check --features quic,websocket` — clean (0 errors, 13 warnings,
    all cosmetic in pre-existing code).
  - `cargo check --no-default-features` — clean.
  - `cargo test --lib --features quic,websocket` — **361 passed, 0 failed,
    1 ignored**.
  - `cargo test --test mod conformance::rust::quic_ping` — **4/4 passed**,
    including the existing `libp2p_dials` interop test that performs a real
    rust-libp2p ↔ litep2p QUIC handshake + ping exchange. This is the
    best wire-level equivalence proof in the suite.
  - Added `transport::quic::listener::tests::handshake_extracts_peer_id_both_sides`
    asserting both peers extract each other's PeerId from the libp2p X.509
    extension after a real handshake.
  - End-to-end TCP/WSS sweep against polkadot/kusama/paseo bootnodes via
    [a downstream consumer](https://github.com/rotkonetworks/bootyspector) with `[patch.crates-io]` — behaviour-equivalent
    for A-record hosts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -294,13 +294,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "chacha20"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6f8d983286843e49675a4b7a2d174efe136dc93a18d69130dd18198a6c167601"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chacha20poly1305"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "10cd79432192d1c0f4e1a0fef9527696cc039165d729fb41b3f4f4f354c2dc35"
 dependencies = [
  "aead",
- "chacha20",
+ "chacha20 0.9.1",
  "cipher",
  "poly1305",
  "zeroize",
@@ -545,7 +556,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ab67060fc6b8ef687992d439ca0fa36e7ed17e9a0b16b25b601e8757df720de"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


kayabaNerve:
While I dislike the undisclosed use of an LLM and don't want to review this PR in general, I do want to say IMO, LiteP2P should not fall back to Cloudflare and if there is no system config, should fail or accept a user-provided fallback. Else, there's privacy implications buried in the LiteP2P stack via a pseudo-dial home.

hitchho:
> While I dislike the undisclosed use of an LLM and don't want to review this PR in general, I do want to say IMO, LibP2P should not fall back to Cloudflare and if there is no system config, should fail or accept a user-provided fallback. Else, there's privacy implications buried in the LibP2P stack via a pseudo-dial home.

there already is tho. its already fallbacking on googles recursive dns so dropping that would be breaking change.  

dropping the default now would be breaking change so even I do agree with you here, that didnt feel right either. out of public recursive dns, my pick would be to use quad9(9.9.9.9) which is nonprofit anycast with ~200 locations ww if we do not want to break anything downstream. 

anyways ill just run my own patch and closing the pr here.

kayabaNerve:
Then IMO, LiteP2p should make the same breaking change. This PR is already a breaking change which will mandate a new release because these dependencies are part of its type contract.

PR #594, opened 2026-05-25
webrtc: Decode webrtc message from multiple SCTP messages

lexnv:
This PR extends WebRTC's compatibility to support interaction with libp2p-go clients.

The libp2p-go implementation makes two separate write calls on the substream:
- one for variant length
- one for protobuf actual message

This then translates over the wire into two SCTP messages. The current implementation only accepts `WebRTC::Messages` if it can be decoded from a single SCTP message.

This breaks the WebRTC spec compatibility at two levels: 
- during the noise handshake logic
- during the message communication

To align the WebRTC impl to be spec-compliant, this PR adds a buffer for both places, ensuring the decoding acts immediately on:
 - a single message (legacy smoldot)
 - two SCTP messages that form one single WebRTC message (libp2p-go)
 

cc @gab8i @dmitry-markin @skunert


gab8i (approved):
LGTM!

src/transport/webrtc/connection.rs
```diff
@@ -210,6 +211,18 @@ pub struct WebRtcConnection {
 
     /// Substream handles.
     handles: SubstreamHandleSet,
+
+    /// Inbound noise-channel byte buffer for reassembling pbio-delimited frames.
```
```suggestion
    /// Inbound data channel byte buffer for reassembling full protobuf frames.
```
nit: pbio is too cryptic :)

src/transport/webrtc/opening.rs
```diff
@@ -111,6 +115,18 @@ pub struct OpeningWebRtcConnection {
 
     /// Local address.
     local_address: SocketAddr,
+
+    /// Inbound noise-channel byte buffer for reassembling pbio-delimited frames.
```
dmitry-markin:
```suggestion
    /// Inbound noise-channel byte buffer for reassembling protobuf frames.
```

src/transport/webrtc/opening.rs
```diff
@@ -244,7 +261,28 @@ impl OpeningWebRtcConnection {
...
+
+        self.noise_recv_buffer.extend_from_slice(&data);
+
+        let body = match extract_framed_message(&mut self.noise_recv_buffer)? {
+            Some(body) => body,
+            None => {
+                tracing::trace!(
+                    target: LOG_TARGET,
+                    buffered = self.noise_recv_buffer.len(),
+                    "incomplete noise frame, waiting for more bytes",
```
dmitry-markin, line 301: It would be good to also log peerid/remote endpoint here and in other logs.

src/transport/webrtc/util.rs
```diff
@@ -114,14 +110,86 @@ impl WebRtcMessage {
...
+    let _ = frame.split_to(varint_len);
+
+    tracing::trace!(
+        target: LOG_TARGET,
+        message_len = len,
+        varint_len,
+        "Successfully extracted SCTP framed message"
+    );
+
+    Ok(Some(frame.freeze()))
```
dmitry-markin, line 171: This results in a copy we ideally want to avoid. Better to return a slice in the original buffer and let the caller advance the pointer/index/deque/etc.. This will also correctly handle the case when there is a next protobuf frame in the same SCTP datagram.

dmitry-markin (approved):

src/transport/webrtc/util.rs
dmitry-markin: Aah, sorry, my bad, no copy here :see_no_evil:

Issue #523, opened 2026-01-23
webrtc: Replace Mutex<State> with lock-free AtomicU8 and consolidate wakers for WebRTC substream


timwu20:
## Summary

Replace the current `Arc<Mutex<State>>` with a lock-free `AtomicU8` for managing WebRTC substream state, as suggested in PR #513.

This change also proposes merging the two separate wakers (`shutdown_waker` and `write_waker`) into a single `state_waker`. Before implementing, we need to verify this waker consolidation is safe and won't cause issues like infinite poll loops or unacceptable spurious wakeups.

## Current Implementation

```rust
state: Arc<Mutex<State>>,
```

With separate wakers:
- `shutdown_waker: Arc<AtomicWaker>` - notifies when FIN_ACK received
- `write_waker: Arc<AtomicWaker>` - notifies when STOP_SENDING/RESET_STREAM received

## Proposed Implementation

```rust
struct SharedState {
    // lock-free / more efficient
    // BITS 0 to 3
    // 0: open 1: sendClosed 2: closing 3: FinSent (our states) 4: FinAck

    // BIT 4: remote recv closed via FIN recv
    // BIT 5: STOP_SENDING remote send closed
    state: AtomicU8,

    // Waker for tasks waiting on state changes (ie poll write blocked on state, but poll shutdown has
    state_waker: AtomicWaker,
}
```

## Investigation Required

Before implementing, we need to verify that we can **merge all wakers into a single `state_waker`**:

1. **`shutdown_waker`** - currently woken when:
   - FIN_ACK received (`on_message` → `State::FinAcked`)
   - Registered in `poll_shutdown` when waiting for FIN_ACK

2. **`write_waker`** - currently woken when:
   - STOP_SENDING received (`on_message` → `State::SendClosed`)
   - RESET_STREAM received (`on_message` → `State::SendClosed`)
   - Registered in `poll_write` when blocked due to backpressure

The key question: Can we safely wake a single `state_waker` on any state transition without causing spurious wakeups that degrade performance?

Since `Substream` is owned and polled by exactly one task, merging should be safe - the same task handles both `poll_write` and `poll_shutdown`. However, we should verify:
- No scenario where waking causes infinite poll loops
- Performance impact of spurious wakeups is acceptable (state transitions are rare)

## Benefits

- Lock-free state access (no mutex contention)
- More efficient for concurrent access patterns
- Reduced memory overhead (single AtomicU8 vs Mutex)

## Trade-offs

[...]

Issue #593, opened 2026-05-21
webrtc: Simplify Substream / SubstreamHandle state machine


lexnv:
### Problem

The state management between Substream (object received by higher-level protocols) and SubstreamHandle polled on webrtc task is getting a bit too convoluted: 

1. We share an `state: Arc<Mutex<State>>` between both objects 
- This is altered by `SubstreamHandle::on_message` to transition the state on `FinAck` and wake up the shutdown wakers. Similarly, transition into `SendClosed` on receiving `Flag::StopSending` and wake up the poll_write waker. And similar upon receiving `ResetStream`
- Then the `Substream::poll_write` rejects any state thats not in `State::Open` with a brokek pipe error.
- `Substream::poll_shutdown` returns pending until the state transitions into `FinAck / Reset`

2. We have a `write_waker` atomic waker registered by the `poll_write`
 - In combination with state mutex, it is effectively signaling a write error upon `Flag::StopSending` or `Flag::ResetStream` (broken pipe)
 
 3. We have a `shutdown_waker` atomic waker registered inside `Substream::poll_shutdown`
 - This is woken up on `FinAck` / `ResetStream`
 - Also from `fn half_close` onm timeout if we didn't receive the `FinAck` in due time
 
 4. We also got a `substream_shutdown` atomic bool with `self.substream_shutdown_waker` atomic waker
 - `substream_shutdown_waker` registered by `SubstreamHandle::poll_next`
 - `Substream::poll_shutdown` / on drop sets the bool to true and wakes the waker
 
### Solution
We can significantly simplfy this architecture by:

 - Eliminatring the Mutex entirely and keeping the state private to the SubstreamHandle 
 - Replace Waker + AtomicBool patterns with tokio channels
   
Ideally, the `Substream` should be simplified to hold only the following:
- `read_buffer` and `rx: Receiver<Event>` unchanged
- `tx: Option<PollSender<Event>>` signal when the poll_write should fail
- `done_signal: Option<oneshot sender<()>>` signals the substreamHandle that shutdown was called / or substream dropped
- `shutdown_complete: Option<oneshot receiver<()>>` awaited by `shutdown` to resolve when the fin ack/reset/timeout reached its state 
 

https://github.com/paritytech/litep2p/pull/586#issuecomment-4508577274

gab8i:
Is this being addressed by #600? Or do you think it deserves even more simplification? @lexnv

lexnv:
[...]

PR #600, opened 2026-05-27
refactor(webrtc): rework Substream <-> SubstreamHandle communication

gab8i:
This PR attempt the refactor of the `substream` module explained here: https://github.com/paritytech/litep2p/issues/593#issue-4495914810

Contrary to what's described in the issue, I ended up using the Mutex + AtomicWaker pattern, because it turned out to be the cleanest and most effective option.

One thing worth knowing: there are roughly 3 states, with multiple tasks modifying and reading them, so there can be multiple producers and consumers. Every attempt to use a data structure that handles the polling internally, so that no task is left behind, stuck waiting on some other parallel event, ended up making the code messy: it added many fields and, most importantly, required remembering to notify/update a watcher (or push to several different places) at every relevant point in the codebase.

What the PR does is:

Rework the communication mechanism between `Substream` and its `SubstreamHandle`. State is shared through `Mutex` and `AtomicWaker` abstracted behind a small helper, which makes them easy to use and ensures the relevant tasks are woken whenever the shared state changes.

This also decouples the reading half from the writing half: a graceful close of either half no longer implies closing the other. An abrupt RESET_STREAM still tears down both, as required by the spec.

### DIFF
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -29,7 +29,7 @@ use crate::{
     transport::{
         webrtc::{
             schema::webrtc::message::Flag,
-            substream::{Event as SubstreamEvent, Substream as WebRtcSubstream, SubstreamHandle},
+            substream::{Message, Substream as WebRtcSubstream, SubstreamHandle},
             util::WebRtcMessage,
         },
         Endpoint,
@@ -116,7 +116,7 @@ impl SubstreamHandleSet {
 }
 
 impl Stream for SubstreamHandleSet {
-    type Item = (ChannelId, Option<SubstreamEvent>);
+    type Item = (ChannelId, Option<Message>);
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let len = match self.handles.len() {
@@ -1068,7 +1068,7 @@ impl WebRtcConnection {
                         self.channels.insert(channel_id, ChannelState::Closing);
                         self.handles.remove(&channel_id);
                     }
-                    Some((channel_id, Some(SubstreamEvent::Message { payload, flag }))) => {
+                    Some((channel_id, Some(Message { payload, flag }))) => {
                         if let Err(error) = self.on_outbound_data(channel_id, payload, flag) {
                             tracing::debug!(
                                 target: LOG_TARGET,
@@ -1083,7 +1083,6 @@ impl WebRtcConnection {
                             self.handles.remove(&channel_id);
                         }
                     }
-                    Some((_, Some(SubstreamEvent::RecvClosed))) => {}
                 },
                 command = self.protocol_set.next() => match command {
                     None | Some(ProtocolCommand::ForceClose) => {
--- a/src/transport/webrtc/substream.rs
+++ b/src/transport/webrtc/substream.rs
@@ -31,10 +31,7 @@ use tokio_util::sync::PollSender;
 
 use std::{
     pin::Pin,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
+    sync::Arc,
     task::{Context, Poll},
     time::Duration,
 };
@@ -46,43 +43,74 @@ const MAX_FRAME_SIZE: usize = 16384;
 /// Matches go-libp2p and js-libp2p's 10-second stream close timeout.
 const FIN_ACK_TIMEOUT: Duration = Duration::from_secs(10);
 
-/// Substream event.
-#[derive(Debug, PartialEq, Eq)]
-pub enum Event {
-    /// Receiver closed.
-    RecvClosed,
+/// Substream Message.
+#[derive(PartialEq, Eq, Debug)]
+pub struct Message {
+    pub payload: Vec<u8>,
+    pub flag: Option<Flag>,
+}
 
-    /// Send/receive message with optional flag.
-    Message {
-        payload: Vec<u8>,
-        flag: Option<Flag>,
-    },
+/// Shared state used to sync `Substream` and `SubstreamHandle`.
+struct SharedState<T: Clone> {
+    state: Arc<Mutex<T>>,
+    wakers: Vec<Arc<AtomicWaker>>,
+}
+
+struct SharedStateWatcher<T: Clone> {
+    state: Arc<Mutex<T>>,
+    waker: Arc<AtomicWaker>,
+}
+
+impl<T: Clone> SharedState<T> {
+    fn new(val: T) -> Self {
+        Self {
+            state: Arc::new(Mutex::new(val)),
+            wakers: vec![],
+        }
+    }
+
+    fn set(&self, new_value: T) {
+        *self.state.lock() = new_value;
+        for waker in self.wakers.iter() {
+            waker.wake()
+        }
+    }
+
+    fn get(&self) -> T {
+        self.state.lock().clone()
+    }
+
+    fn register_watcher(&mut self) -> SharedStateWatcher<T> {
+        let waker = Arc::new(AtomicWaker::new());
+        self.wakers.push(waker.clone());
+        SharedStateWatcher {
+            state: self.state.clone(),
+            waker,
+        }
+    }
+}
+
+impl<T: Clone> SharedStateWatcher<T> {
+    fn get(&self, cx: &mut Context<'_>) -> T {
+        self.waker.register(cx.waker());
+        self.state.lock().clone()
+    }
 }
 
 /// Substream stream.
 #[derive(Debug, Clone, Copy)]
-enum State {
+enum ChannelState {
     /// Substream is fully open.
     Open,
-
-    /// Remote is no longer interested in receiving anything.
-    SendClosed,
-
     /// ResetStream has been received.
+    ///
+    /// This preempts both writer and reader state.
     Reset,
-
-    /// We sent FIN, waiting for FIN_ACK.
-    FinSent,
-
-    /// We received FIN_ACK, write half is closed.
-    FinAcked,
-
-    /// Substream was dropped before a graceful close completed.
-    ForceShutdown,
 }
 
 /// State of the reading side of the stream.
-enum ReadingState {
+#[derive(Clone)]
+enum ReaderState {
     /// The reading stream is open.
     Open,
     /// A Fin flag was received.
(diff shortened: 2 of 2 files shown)


src/transport/webrtc/substream.rs
```diff
@@ -46,43 +43,74 @@ const MAX_FRAME_SIZE: usize = 16384;
...
+
+    fn set(&self, new_value: T) {
+        *self.state.lock() = new_value;
+        for waker in self.wakers.iter() {
+            waker.wake()
+        }
+    }
+
+    fn get(&self) -> T {
+        self.state.lock().clone()
```
lexnv: nit: Could we get into TOCTOU issues? We lock here, clone, on drop the lock goes away, we use the cloned state, and by the time we read the clone, the actual state behind mutex changes?

src/transport/webrtc/substream.rs
```diff
@@ -46,43 +43,74 @@ const MAX_FRAME_SIZE: usize = 16384;
...
+    }
+
+    fn register_watcher(&mut self) -> SharedStateWatcher<T> {
+        let waker = Arc::new(AtomicWaker::new());
+        self.wakers.push(waker.clone());
```
lexnv: Would this grow unboundedly overtime?

gab8i: The idea behind a vec is just to simplify the struct itself and its usage. Right now the code doesn't use more than 2 wakers per state, this would allow the usage of just a `[Option<AtomicWaker>; 2]` and panic if we register more then 2 wakers. That could be fine but not really maintainable, if for example at some point the webrtc protocl changes or we need for some other reason a new waker.

src/transport/webrtc/connection.rs
```diff
@@ -1083,7 +1083,6 @@ impl WebRtcConnection {
                             self.handles.remove(&channel_id);
                         }
                     }
-                    Some((_, Some(SubstreamEvent::RecvClosed))) => {}
```
lexnv, line 1086: Nice!

src/transport/webrtc/substream.rs
```diff
@@ -202,33 +197,30 @@ impl SubstreamHandle {
...
-                    .inbound_tx
-                    .send(Event::Message {
+        match (self.message_tx.as_ref(), message.payload) {
+            (Some(message_tx), Some(payload)) if !payload.is_empty() => {
+                // TODO: awaiting here makes the entire connection
+                // rely on the readers to be fast enough, a slow reader
+                // could cause this method to wait and thus stall the entire webrtc
+                // connection. Solution would be to implement reading
+                // backpressure, keeping track of pending incoming messages.
+                let _ = message_tx
```
lexnv: nit: Lets add a debug log here on errors

src/transport/webrtc/substream.rs
```diff
@@ -163,36 +175,19 @@ impl Substream {
...
+    /// State of the channel.
+    channel_state: SharedState<ChannelState>,
+    /// State of the writing half.
+    writer_state: SharedState<WriterState>,
+    /// State of the reading half.
+    reader_state: SharedState<ReaderState>,
     /// TX channel for sending inbound messages from `peer` to the associated `Substream`.
-    inbound_tx: Sender<Event>,
-
+    message_tx: Option<Sender<Message>>,
```
lexnv, line 189: I would extend the comment to document that we are dropping the channel upon receiving the `Fin` packet, after which no more messages are sent.

src/transport/webrtc/substream.rs
```diff
@@ -202,33 +197,30 @@ impl SubstreamHandle {
...
         }
 
         // Process payload first, before handling flags.
-        // This ensures that if a FIN message contains data, we deliver it before closing.
-        if let Some(payload) = message.payload {
-            if !payload.is_empty()
-                && self
-                    .inbound_tx
-                    .send(Event::Message {
+        match (self.message_tx.as_ref(), message.payload) {
```
lexnv, line 213: `message_tx` is dropped upon receiving the `Fin` packet.

Practically, we could still receive a message payload here even tho we received the `fin` and dropped the tx channel.
IIRC, spec states that no further messages are received, maybe we should emit an warning (that will become debug once we stabilize)?

src/transport/webrtc/substream.rs
```diff
@@ -237,67 +229,50 @@ impl SubstreamHandle {
...
                 }
                 Flag::FinAck => {
                     // Received FIN_ACK, we can now fully close our write half
-                    let mut state = self.state.lock();
-                    if matches!(*state, State::FinSent) {
-                        *state = State::FinAcked;
-                        // Wake up any task waiting on shutdown
-                        self.substream_shutdown_waker.wake();
-                        self.shutdown_waker.wake();
+                    if matches!(self.writer_state.get(), WriterState::Fin) {
```
lexnv: nit: Let's cache the writer_state in a variable, by the time we read it again under the mutex it might have already mutated so the debug log will make investigations a bit more difficult

src/transport/webrtc/substream.rs
```diff
@@ -237,67 +229,50 @@ impl SubstreamHandle {
...
+                        self.writer_state.set(WriterState::FinAck);
                     } else {
-                        tracing::warn!(
+                        tracing::debug!(
                             target: LOG_TARGET,
-                            ?state,
+                            state = ?self.writer_state.get(),
                             "received FIN_ACK in unexpected state, ignoring"
```
lexnv: We expect the state to be exactly `Fin` to transition to `FinAck`. How could we make this a bit more robust here just in case we introduced a bug somewhere or the other peer is misbehaving?

- maybe we could clean up the states regardless or send a reset or return ConnectionClosed

src/transport/webrtc/substream.rs
```diff
@@ -237,67 +229,50 @@ impl SubstreamHandle {
...
-                    let recv_closed_sent = self.inbound_tx.send(Event::RecvClosed).await.is_ok();
-                    if !recv_closed_sent {
-                        tracing::debug!(target: LOG_TARGET, "Substream dropped, skipping RecvClosed");
-                    }
-
-                    // TODO: after this reading_state change there is nothing which wakes the
-                    // Stream thus FinAck will only be sent once something else awake it.
-                    *self.reading_state.lock() = ReadingState::Fin;
+                    self.reader_state.set(ReaderState::Fin);
+                    let _ = self.message_tx.take();
```
lexnv: nit: 

```suggestion
if self.message_tx.take().is_none() {
    warn!("Unexpected to have channel already dropped / similar")
}

File: src/transport/webrtc/substream.rs [hunk 10]
```diff
@@ -306,120 +281,140 @@ impl SubstreamHandle {
...
-                            flag: Some(Flag::ResetStream),
-                        }));
-                    }
-                }
-
-                Poll::Pending
-            }
+    // This function carries forward the writer half close process.
+    //
+    // It is expected to:
```
lexnv: nit or similar: The following behaviors are expected on:

- `WriterState::Open` state:
  - etc

- `WriteState::Fin` state:
  - etc

src/transport/webrtc/substream.rs
```diff
@@ -306,120 +281,140 @@ impl SubstreamHandle {
...
+        match self.writer_state.get() {
             // First call to shutdown, if peer sent StopSending we are still
             // free to send Fin to make sure this half closes properly.
-            State::Open | State::SendClosed => {
+            WriterState::Open | WriterState::StopSending => {
                 // Initialize the timeout for FIN_ACK
                 let mut timeout = Box::pin(tokio::time::sleep(FIN_ACK_TIMEOUT));
                 // Poll the timeout once to register it with tokio's timer
                 // This ensures we'll be woken when it expires
                 let _ = timeout.as_mut().poll(cx);
```
lexnv:
```nit:

if timer.poll().is_ready() {
    error! ("Misconfigured timer is not supposed to be ready")
}

File: src/transport/webrtc/substream.rs [hunk 12]
```diff
@@ -306,120 +281,140 @@ impl SubstreamHandle {
...
+            // If something went wrong, RESET_STREAM should have been sent,
+            // in that case the channel is treated as closed.
             return Poll::Ready(None);
         }
 
+        // - reader_state: this is mainly driven by the `on_message` function which reacts to
+        //   incoming messages, there are 2 side effects which connects the two streams:
+        //   1. If FIN arrived then FIN_ACK is expected to be sent back.
+        //   2. If FIN_ACK arrived the writer_state is updated.
         {
```
lexnv: I think we can safely drop the `{ }` block, we are hiding the locks under `set / register_and_get`

src/transport/webrtc/substream.rs
```diff
@@ -202,33 +197,30 @@ impl SubstreamHandle {
...
         // Process payload first, before handling flags.
-        // This ensures that if a FIN message contains data, we deliver it before closing.
-        if let Some(payload) = message.payload {
-            if !payload.is_empty()
-                && self
-                    .inbound_tx
-                    .send(Event::Message {
+        match (self.message_tx.as_ref(), message.payload) {
+            (Some(message_tx), Some(payload)) if !payload.is_empty() => {
+                // TODO: awaiting here makes the entire connection
```
lexnv, line 222: Lets create an issue for this one if not already and place it in the comment here

src/transport/webrtc/substream.rs
```diff
@@ -46,43 +43,66 @@ const MAX_FRAME_SIZE: usize = 16384;
...
+}
+
+/// Shared state used to sync `Substream` and `SubstreamHandle`.
+#[derive(Clone)]
+struct SharedState<T: Clone> {
+    inner: Arc<Inner<T>>,
+}
+
+struct Inner<T> {
+    state: Mutex<T>,
```
lexnv, line 60: I would take this oen step further and replace the `Mutex` with an `AtomicU8`

Accidently, this is what I've proposed in the past to the original shutdown impl from https://github.com/paritytech/litep2p/pull/513#discussion_r2716332541. 

Since we follow the behavior:
- set: mutex.lock (ie `.store(Release)`)
- get: mutex.lock (ie `.load(Acquire)`)

It should be straight forward (and cause no side-effects) to replace it with a lock free atomic:

```enum
enum WriterState {
    /// The writing stream is open.
    Open = 0,
    /// A Fin flag was sent.
    Fin = 1,
    /// FinAck was received.
    FinAck = 2,
    /// StopSending was received.
    StopSending = 3,
}

```

Then we could also close: https://github.com/paritytech/litep2p/issues/523

lexnv:
Nice job here @gab8i! This simplifies the state machines quite a lot!

While at it,  `AtomicU8` should be a drop-in replacement of `Arc<Mutex<>` without side-effects. Since everything is placed behind a nice `set/get` API, I would tackle that replacement here 🙏  Other than that left some tiny nits about comments and debuggability (feel free to ingnore where it doesnt make sense)

src/transport/webrtc/substream.rs
gab8i: Yes, this makes sense!! str0m should guarantee the order of things and thus after FIN no other message should arrive, beside just flags!

src/transport/webrtc/substream.rs
gab8i: We should reset the connection if we received FIN_ACK from an unexpected state, this seems to make perfectly sense! Within spec FIN_ACK should be received only in one scenario so I think it is right to expect it only when we have a WriterState::FIN

src/transport/webrtc/substream.rs
gab8i: Yup, that's a vestigial from when a lock was acquired and dropped at the end of the sub-scope! :+1:

src/transport/webrtc/substream.rs
gab8i: That's a great idea!! And given the current `SharedState` abstraction is seems pretty straight forward to implement!!!

PR #605, opened 2026-05-29
[WIP] feat(webrtc): persistent dtls certificate

gab8i:


src/transport/webrtc/mod.rs
```diff
@@ -422,11 +424,43 @@ impl TransportBuilder for WebRtcTransport {
...
-        )?;
+        let generate_cert = |crypto: &str0m::crypto::CryptoProvider| {
+            crypto.dtls_provider.generate_certificate().ok_or(crate::error::Error::WebRtc(
+                RtcError::Dtls(DtlsError::CryptoError(CryptoError::Other(
+                    "OpenSsl failed to generate certificate".to_string(),
+                ))),
+            ))
+        };
+
+        let dtls_cert = if let Some(dtls_cert_path) = config.dtdl_cert_persistent_path {
```
lexnv: This is a good step in the right direction.

I would refactor the code a bit to improve maintenance and simplify where possible:
- We could introduce a new `struct CerthashCache` in a new module `webrtc/certhash.rs`
- `CerthashCache` is constructed with an optional `certhash_persistent_path`
- The encode / decode fns are methods on the certhahs as private
- `webrtc_certhash` appended is a const on the module and reflected in the documentation on the config
- then we can test in isolation different scenarios:
  - encode / decode w real file 
  - encode + broken file / decode etc

Then the `webrtc/mod.rs` would only contain (or similar):

```rust
let cert = CerthashCache::new(config.path).build()?;
```

src/transport/webrtc/mod.rs
```diff
@@ -746,3 +780,91 @@ fn is_stun_packet(bytes: &[u8]) -> bool {
...
+    file.write_all(&(dtls_cert.private_key.len() as u64).to_be_bytes()[..])?;
+    file.write_all(&dtls_cert.private_key[..])?;
+
+    Ok(())
+}
+
+/// Decodes a [`DtlsCert`] from the `webrtc_certhash` file inside `path`.
+///
+/// Layout, lengths as big-endian `u64`:
+/// `cert len ++ cert bytes ++ key len ++ key bytes`.
```
lexnv: Looks good no strong preferences here

src/transport/webrtc/dtlscert.rs
```diff
@@ -0,0 +1,194 @@
...
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                "dtls cert folder doesn't exist",
+            ));
+        }
+
+        let path = path.join(DTLS_CERT_FILE);
+
+        let mut file =
+            std::fs::OpenOptions::new().create(true).write(true).truncate(true).open(path)?;
```
dmitry-markin, line 117: We shouldn't do any file IO from a networking library. Instead we need to expose an API for substrate to load/save the certificate.

dmitry-markin: Similarly to how we handle the network key.

src/transport/webrtc/config.rs
```diff
@@ -32,6 +33,17 @@ pub struct Config {
...
+    /// Folder used to persist the DTLS certificate across restarts.
+    ///
+    /// If specified and a certificate file already exists in the folder, the
+    /// certificate is loaded from it, otherwise a new certificate is generated
+    /// and saved into the folder. Persisting it keeps the node's `certhash`
+    /// (and therefore its advertised addresses) stable across restarts.
+    ///
+    /// NOTE: a folder must be used by a single node only, different nodes on the
+    /// same machine must use different folders.
+    pub dtdl_cert_persistent_path: Option<PathBuf>,
```
lexnv, line 46:
```suggestion
    pub dtls_cert_persistent_path: Option<PathBuf>,
```

src/transport/webrtc/dtlscert.rs
lexnv: Don't have a strong opinion on this, but at some point we might want to persist some internal structs (maybe peer addresses / dht but remember that turned out to not work as expected in the past)

Then the alternative would be to expose a `DtlsCert` that has `build(wihtout str0m exposure)` and `load(cert_bytes, key_bytes)`. The decoding / encoding then happens in substrate and this is the structure we pass on the config as optional

lexnv, line 117: I believe write / truncate will create the file with broader access permissions (644), for secrets we should use (600). Worth double checking the private keys under the substrate DB

src/transport/webrtc/dtlscert.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        file.write_all(&(dtls_cert.certificate.len() as u64).to_be_bytes()[..])?;
```
lexnv, line 119: This could partially fail: we write the len, but then fail to write the cert bytes. This will permanently break the file format. 

Instead, if moving to substrate from Dmitry's suggestion we can have a single write/read:
```rust
let bytes = (cert, key).encode();
write(bytes)

let bytes = read(file)
let (cert, key) = bytes.decode()
```

src/transport/webrtc/dtlscert.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    /// Decodes a [`DtlsCert`] from the `webrtc_certhash` file inside `path`.
+    ///
+    /// Layout, lengths as big-endian `u64`:
+    /// `cert len ++ cert bytes ++ key len ++ key bytes`.
+    ///
+    /// The `bool` in the error says whether the caller may persist a fresh cert:
+    /// - true: file absent, generate and save a new one.
+    /// - false: file present but unreadable/corrupt, do NOT overwrite it
+    fn decode(path: &Path) -> Result<DtlsCert, bool> {
```
lexnv, line 135: The `decode / load` could double check the integrity of the bytes: confirm the cert + key form a valid certificate. 

Currently, we are checking the bytes match the length without other guarantees. If one byte is modified / file is invalid but has the same length header, we'd say the certificate is valid. Then, fallback to generate a new ephemeral certificate.

We should double check here and fall back to the network key behavior from substrate

lexnv: Probably we should regenerate the cert and overwrite the file with an error / warn to tackle these cases, but worth a double check with substrate behavior

src/transport/webrtc/dtlscert.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /// Encodes `dtls_cert` into the `webrtc_certhash` file inside `path`.
+    ///
+    /// Layout, lengths as big-endian `u64`:
+    /// `cert len ++ cert bytes ++ key len ++ key bytes`.
+    ///
+    /// Errors if `path` does not exist.
+    fn encode(path: &Path, dtls_cert: &DtlsCert) -> std::io::Result<()> {
+        if !path.exists() {
```
lexnv, line 107: Don't think we need this, probably can propagate the open path errors which provides standard os errors instead of crafting a new one

src/transport/webrtc/dtlscert.rs
```diff
@@ -0,0 +1,194 @@
...
+    Invalid(PathBuf),
+    /// A valid certificate was loaded from disk and will be reused as-is.
+    Valid(DtlsCert),
+}
+
+/// Loads and persists the node's WebRTC DTLS certificate.
+///
+/// Construct with [`DtlsCertCache::new`], then call [`DtlsCertCache::build`]
+/// to obtain the [`DtlsCert`] the transport should use.
+pub struct DtlsCertCache {
```
lexnv, line 39: Since we have this structure in palce, we could add some dedicated testing

src/transport/webrtc/dtlscert.rs
dmitry-markin: > at some point we might want to persist some internal structs

This must also go explicitly through the client code. Writing files from a library, specifically a networking library (e.g., as opposed to DB library) can cause a lot of issues and is strong no-go. To count a few — you already noticed the file permissions, that should be managed by the user/host process; the process might not have a read-write access to the filesystem at all due to missing home dir/confinement/containerization/etc..; the host app might choose to store data not in a file, but in an external database. We must under no circumstances drop/read files from a "default" location (at a very minimum, the location will be shared by all processes using the library). I.e., the things will subtly break and will be hard to debug.

lexnv: Oki doki makes sense to me 🙏 Thanks for the info, yep, let's move this in substrate

gab8i:
The approach to handling persistent certificates has changed, and the responsibility for keeping them persistent is now moved to the user. Litep2p only offers a way to generate the certificate and then expects it during the initialization phase.

PR #610, opened 2026-06-05
feat(webrtc): update how webrtc certificate is handled

gab8i:
add generate, encode, decode and validate functionalities for webrtc certificate, this to allow creating a persistent version which needs to be saved by the user and specified on every start-up to create a consistent multiaddr.

NOTE: while developing the validation, I discovered that str0m panics if the specified DTLS cert is not valid, it doesn't return an error.

To implement the `validate_certificate` function I used `catch_unwind`, but there are several problems with this approach. Two of them are:

1. It isn't pretty, and doesn't feel safe to rely on. `catch_unwind` is not foolproof, and there will still be traces of the panic that happened within str0m.
2. The main construction of the WebRTC instance can technically panic, but this should not happen: when the certificate is specified by the user, it is validated before being used, and when it is generated on the fly, it is not expected to cause any issue, at least according to what I gathered from the str0m docs.

EDIT: after multiple comments and discussion this PR has been simplified quite a lot. Now litep2p has only the bare minimum objective of wrapping `str0m` dtls certificate and provide and easy way to the caller to generate and to save it somewhere to make it persistent across run.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -531,6 +531,31 @@ impl Litep2p {
             }
         }
     }
+
+    /// Generates a fresh WebRTC DTLS certificate, encoded for persistence.
+    ///
+    /// The returned bytes contain the certificate **and its private key**.
+    /// Feed the value back via [`webrtc::Config::raw_dtls_certificate`]
+    /// to keep a stable WebRTC identity (`certhash`) across restarts.
+    #[cfg(feature = "webrtc")]
+    pub fn generate_webrtc_certificate() -> crate::Result<Vec<u8>> {
+        let certificate = transport::webrtc::generate_certificate()?;
+        Ok(transport::webrtc::encode(&certificate))
+    }
+
+    /// Validates an encoded WebRTC DTLS certificate.
+    ///
+    /// Accepts the bytes produced by [`Litep2p::generate_webrtc_certificate`] and
+    /// checks that they decode into a certificate and private key the DTLS stack
+    /// can actually use.
+    ///
+    /// Returns an error if the bytes are truncated, the encoded lengths are
+    /// inconsistent, or the certificate/key fails to load.
+    #[cfg(feature = "webrtc")]
+    pub fn validate_webrtc_certificate(encoded_certificate: Vec<u8>) -> crate::Result<()> {
+        let certificate = transport::webrtc::decode(encoded_certificate)?;
+        transport::webrtc::validate_certificate(&certificate)
+    }
 }
 
 #[cfg(test)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -361,9 +361,11 @@ impl Litep2p {
             transport_manager.register_transport(SupportedTransport::Quic, Box::new(transport));
         }
 
-        // enable webrtc transport if the config exists
+        // Enable webrtc transport if the config exists and is valid.
+        // A config without any listen address is skipped gracefully instead of
+        // failing startup.
         #[cfg(feature = "webrtc")]
-        if let Some(config) = litep2p_config.webrtc.take() {
+        if let Some(config) = litep2p_config.webrtc.take().filter(|config| config.is_valid()) {
             let handle = transport_manager.transport_handle(Arc::clone(&litep2p_config.executor));
             let (transport, transport_listen_addresses) =
                 <WebRtcTransport as TransportBuilder>::new(handle, config, resolver.clone())?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -365,17 +365,23 @@ impl Litep2p {
         // A config without any listen address is skipped gracefully instead of
         // failing startup.
         #[cfg(feature = "webrtc")]
-        if let Some(config) = litep2p_config.webrtc.take().filter(|config| config.is_valid()) {
-            let handle = transport_manager.transport_handle(Arc::clone(&litep2p_config.executor));
-            let (transport, transport_listen_addresses) =
-                <WebRtcTransport as TransportBuilder>::new(handle, config, resolver.clone())?;
+        if let Some(config) = litep2p_config.webrtc.take() {
+            if !config.is_valid() {
+                tracing::warn!(target: LOG_TARGET, "Invalid WebRtc Config, no listen address specified");
+            } else {
+                let handle =
+                    transport_manager.transport_handle(Arc::clone(&litep2p_config.executor));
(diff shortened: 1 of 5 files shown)


src/transport/webrtc/certificate.rs
```diff
@@ -0,0 +1,168 @@
...
+    str0m::crypto::from_feature_flags().dtls_provider.generate_certificate().ok_or(
+        crate::error::Error::WebRtc(RtcError::Dtls(DtlsError::CryptoError(CryptoError::Other(
+            "OpenSsl failed to generate certificate".to_string(),
+        )))),
+    )
+}
+
+/// Validates that a reconstructed certificate is usable by the DTLS stack.
+pub fn validate_certificate(certificate: &DtlsCert) -> crate::Result<()> {
+    catch_unwind(|| {
```
lexnv: Do we panic in str0m on invalid certificates?

gab8i: If the certificate is validated upfront it should not panic. There is no real guarantee over this though, str0m simply have some 'expect's over dtls certificate usage.

Do you think it is worth to have `catch_unwind` also within the main `LiteP2p` construction? It would not cover panic=abort cases though

lexnv: Was wondering if we need the catch unwind and if we could rely instead on str0m error code from:

```rust
if rtc.direct_api().start_dtls(false).is_err() {
return false
}
```

gab8i: The problem is berfore within the function `make_rtc_client`:

```rust
        let mut rtc = Rtc::builder()
            .set_ice_lite(true)
            .set_dtls_cert(self.dtls_cert.clone())
            .set_fingerprint_verification(false)
            .build(std::time::Instant::now()); // <- panic occurs here
```

Where build literally at some point does the following: 
```rust
            dtls: Dtls::new(
                &dtls_cert,
                crypto_provider.dtls_provider,
                crypto_provider.sha256_provider,
                start,
                config.dtls_version,
            )
            .expect("DTLS to init without problem"),

```

The `build` function does not return any error, for each one if encountered it just panics...

lexnv:
We could simplify this a bit by loading the DTLS certificate as 2 dedicated byte vectors: cert and private key.
This allows users to keep the DTSL certificate in any format they choose on disk (ie scale encoded tuple).


The API would look like: 

```rust

/// We expose a dedicated litep2p to not lock into str0m without a breaking change later
#[derive Clone]
pub struct DtlsCertificate {
   certificate: Vec<u8>,
   private_key: Vec<u8>,
}

impl DtlsCertificate {
   pub fn load(certificate: Vec<u8>, private_key: Vec<u8>) -> Result<Self, Error> 

   pub fn new() -> Result<Self, Error>
   
   pub fn into_parts(self) -> (Vec<u8>, Vec<u8>)
}
```


Then the `DtlsCertificate` is exported from the webrtc module:
- It is utilized inside the config 
- No method wrappers are added on the `litep2p` object itself (api remains isolated and purpose-driven)
 
 
 The loading function could validate the DTLS certificate; calling `str0m` is a bit of a hack.
 Since we expose only the OpenSSL feature flag, we can assume for the time being that.
 We could import the str0m-openssl and perform the check similarly to str0m:
 - https://github.com/algesten/str0m/blob/cdca17feafd6df514d23fd0859584f03a4785821/crypto/openssl/src/dtls_ossl.rs#L393-L395
 
 ```rust
use openssl::x509::X509;
use openssl::pkey::PKey;

fn validate_cert_and_key(cert_bytes: &[u8], key_bytes: &[u8]) -> Result<bool, ErrorStack> {
    let cert = X509::from_der(cert_bytes)?; 
    let pkey = PKey::private_key_from_der(key_bytes)?;

    // Verify the private key matches the public key inside the cert
    let is_match = cert.verify(&pkey)?;

    // Check expiration dates
    let now = openssl::asn1::Asn1Time::days_from_now(0)?;
    let is_not_expired = cert.not_after() > &now && cert.not_before() <= &now;
    Ok(is_match && is_not_expired)
}
```
 
Prev PR:
- https://github.com/paritytech/litep2p/pull/605

File: src/transport/webrtc/certificate.rs [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+            private_key,
+        })
+    }
+
+    /// Generates a fresh WebRTC DTLS certificate.
+    pub fn new() -> crate::Result<Self> {
+        let dtls_cert =
+            str0m::crypto::from_feature_flags().dtls_provider.generate_certificate().ok_or(
+                crate::error::Error::WebRtc(RtcError::Dtls(DtlsError::CryptoError(
+                    CryptoError::Other("OpenSsl failed to generate certificate".to_string()),
```
lexnv:
```suggestion
                    CryptoError::Other("Crypto provider failed to generate DTLS certificate"".to_string()),
```

src/transport/webrtc/config.rs
```diff
@@ -22,7 +22,12 @@
 
 use multiaddr::Multiaddr;
 
+use crate::transport::webrtc::certificate::DtlsCertificate;
+
 /// WebRTC transport configuration.
+///
+/// To be valid, the configuration must contain at least one listen address,
+/// otherwise the WebRTC transport is skipped at startup.
```
lexnv:
```suggestion
/// otherwise, the WebRTC transport is not initialized.
```

src/lib.rs
```diff
@@ -361,9 +361,11 @@ impl Litep2p {
             transport_manager.register_transport(SupportedTransport::Quic, Box::new(transport));
         }
 
-        // enable webrtc transport if the config exists
+        // Enable webrtc transport if the config exists and is valid.
+        // A config without any listen address is skipped gracefully instead of
+        // failing startup.
         #[cfg(feature = "webrtc")]
-        if let Some(config) = litep2p_config.webrtc.take() {
+        if let Some(config) = litep2p_config.webrtc.take().filter(|config| config.is_valid()) {
```
lexnv: Let's add a warning here

Ideally, users shouldn't provide a WebRTC config if they don't have any listen addresses.

One option to enforce this would be via a `ConfigBuilder` which requires `build(self, listen_addr) -> result<>`, that could be a followup since we are still experimental 🙏

lexnv (approved):
Nicely done here! 🚀

skunert (approved):
Nice!

dmitry-markin (approved):

PR #611, opened 2026-06-08
kad: Implement client-server mode for Kademlia

lexnv:
This PR implements the client-server mode of operation for Kademlia.

Full nodes and substrate-nodes utilize the server mode by default:
- nodes respond to queries
- They are added to the routing table in automatic mode

Light clients or DHT crawlers are encouraged to use the client mode:
- To preserve resources, they are not responding to queries
- Client nodes are not added to the routing table, therefore not polluting the DHT (or even worse, intertwining multiple DHTs)

To prepare for supporting the dynamic addition and removal of multiple DHTs at runtime, this PR introduces dynamic protocol configuration for the identify protocol.
By default, Kademlia implementations running in client mode are not advertised via the identify protocol.

cc @dmitry-markin @skunert

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,12 @@ use crate::{
     config::Litep2pConfig,
     error::DialError,
     protocol::{
-        libp2p::{bitswap::Bitswap, identify::Identify, kademlia::Kademlia, ping::Ping},
+        libp2p::{
+            bitswap::Bitswap,
+            identify::Identify,
+            kademlia::{Kademlia, KademliaMode},
+            ping::Ping,
+        },
         mdns::Mdns,
         notification::NotificationProtocol,
         request_response::RequestResponseProtocol,
@@ -265,12 +270,20 @@ impl Litep2p {
         }
 
         // start kademlia protocol event loops
+        //
+        // protocol names of client-mode instances are not advertised via identify
+        let mut unadvertised_protocols = Vec::new();
         for kademlia_config in litep2p_config.kademlia.into_iter() {
             tracing::debug!(
                 target: LOG_TARGET,
                 protocol_names = ?kademlia_config.protocol_names,
+                mode = ?kademlia_config.mode,
                 "enable ipfs kademlia protocol",
             );
+
+            if kademlia_config.mode == KademliaMode::Client {
+                unadvertised_protocols.extend(kademlia_config.protocol_names.iter().cloned());
+            }
 
             let main_protocol =
                 kademlia_config.protocol_names.first().expect("protocol name to exist");
@@ -402,9 +415,14 @@ impl Litep2p {
             }));
         }
 
-        // if identify was enabled, give it the enabled protocols and listen addresses and start it
-        if let Some((service, mut identify_config)) = identify_info.take() {
-            identify_config.protocols = transport_manager.protocols().cloned().collect();
+        // if identify was enabled, give it the advertised protocols and listen addresses and
+        // start it
+        if let Some((service, identify_config)) = identify_info.take() {
+            *identify_config.protocols.write() = transport_manager
+                .protocols()
+                .filter(|protocol| !unadvertised_protocols.contains(protocol))
+                .cloned()
+                .collect();
             let identify = Identify::new(service, identify_config);
 
             litep2p_config.executor.run(Box::pin(async move {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -417,12 +417,8 @@ impl Litep2p {
 
         // if identify was enabled, give it the advertised protocols and listen addresses and
         // start it
-        if let Some((service, identify_config)) = identify_info.take() {
-            *identify_config.protocols.write() = transport_manager
-                .protocols()
-                .filter(|protocol| !unadvertised_protocols.contains(protocol))
-                .cloned()
-                .collect();
+        if let Some((service, mut identify_config)) = identify_info.take() {
+            identify_config.protocols = transport_manager.protocols().cloned().collect();
             let identify = Identify::new(service, identify_config);
 
             litep2p_config.executor.run(Box::pin(async move {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,9 +46,7 @@ use crate::{
         SubstreamKeepAlive,
     },
     transport::{
-        manager::{
-            AdvertiseProtocol, SupportedTransport, TransportManager, TransportManagerBuilder,
-        },
+        manager::{InboundProtocol, SupportedTransport, TransportManager, TransportManagerBuilder},
         tcp::TcpTransport,
         TransportBuilder, TransportEvent,
     },
@@ -207,7 +205,7 @@ impl Litep2p {
                 config.codec,
                 litep2p_config.keep_alive_timeout,
                 SubstreamKeepAlive::Yes,
-                AdvertiseProtocol::Yes,
+                InboundProtocol::Accept,
             );
             let executor = Arc::clone(&litep2p_config.executor);
             litep2p_config.executor.run(Box::pin(async move {
@@ -229,7 +227,7 @@ impl Litep2p {
                 config.codec,
                 litep2p_config.keep_alive_timeout,
                 SubstreamKeepAlive::Yes,
-                AdvertiseProtocol::Yes,
+                InboundProtocol::Accept,
             );
             litep2p_config.executor.run(Box::pin(async move {
                 RequestResponseProtocol::new(service, config).run().await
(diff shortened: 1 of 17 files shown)


skunert:
In general looking good. From looking into this I was wondering about the impact on the routing table. So basically after this change the routing tables will be filles more conservatively which peers which have answered to some of our queries. Peers we discovered on the way but did not query will not be added, even though they might be in server mode (we just don't know).  It sounds good to me, I assume we expect no impact? Or even positive impact, because these peers have already answered us.

src/protocol/libp2p/kademlia/handle.rs
```diff
@@ -76,6 +76,17 @@ pub enum IncomingRecordValidationMode {
     Automatic,
 }
 
+/// Kademlia operating mode.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum KademliaMode {
```
dmitry-markin: nit: move to `config`, not `handle`?

gab8i (approved):
Nice! So few lines of code for quite a feature!

src/lib.rs
```diff
@@ -402,9 +415,14 @@ impl Litep2p {
             }));
         }
 
-        // if identify was enabled, give it the enabled protocols and listen addresses and start it
-        if let Some((service, mut identify_config)) = identify_info.take() {
-            identify_config.protocols = transport_manager.protocols().cloned().collect();
+        // if identify was enabled, give it the advertised protocols and listen addresses and
+        // start it
+        if let Some((service, identify_config)) = identify_info.take() {
+            *identify_config.protocols.write() = transport_manager
```
dmitry-markin: There is currently no need to share `RwLock`, because the list of identify protocols only gets updated on the litep2p startup.

In general, sharing access to a private struct should be avoided and replaced by higher-level API for either adding/removing identify protocols, or even better enabling/disabling the advertisement of individual installed protocols.

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -232,6 +242,19 @@ impl ConfigBuilder {
...
+    /// By default, Substrate nodes operate in server mode.
+    ///
+    /// Light clients, or nodes that need to query DHT information without
+    /// participating in DHT routing, should use client mode.
+    ///
+    /// Defaults to [`KademliaMode::Server`].
+    pub fn with_mode(mut self, mode: KademliaMode) -> Self {
+        self.mode = mode;
+        self
+    }
```
dmitry-markin, line 267: We should also make it possible to switch from client to server mode at runtime — e.g., when we discover a reachable external address. Can be a follow-up to make the PR smaller.

src/lib.rs
```diff
@@ -265,13 +270,21 @@ impl Litep2p {
         }
 
         // start kademlia protocol event loops
+        //
+        // protocol names of client-mode instances are not advertised via identify
+        let mut unadvertised_protocols = Vec::new();
```
dmitry-markin: Keeping a separate list of ignored protocols seems to be error-prone. Can we update a `transport_manager.register_protocol()` to accept a mandatory parameter of whether we advertise the protocol instead?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -415,18 +419,24 @@ impl Kademlia {
     async fn on_inbound_substream(&mut self, peer: PeerId, substream: Substream) {
         tracing::trace!(target: LOG_TARGET, ?peer, "inbound substream opened");
 
+        if let KademliaMode::Client = self.mode {
+            tracing::trace!(target: LOG_TARGET, ?peer, "ignoring inbound substream in client mode");
+            let _ = substream.close().await;
```
dmitry-markin, line 424: We need to make sure this won't cause the nodes without client-server understanding to busy-loop on trying to open a substream. Reading and silently discarding all the input might be better if it is the case.

src/protocol/libp2p/kademlia/mod.rs
dmitry-markin, line 454: It is not anymore the `RoutingTableUpdate` event, because we don't update the routing table. What about moving to `insert_proven_server` to not break the API by renaming the event?

src/protocol/libp2p/kademlia/mod.rs
```diff
@@ -515,7 +533,7 @@ impl Kademlia {
                         "handle `PUT_VALUE` response",
                     );
 
-                    self.engine.register_response(
+                    let _ = self.engine.register_response(
```
dmitry-markin: We also learn the node is reachable through `PUT_VALUE/ADD_PROVIDER/etc.` success, so makes sense updating the routing table here as well.

`ADD_PROVIDER` is trickier, because the spec doesn't define the response message, so we should track the substream opening / message sending success instead (already done for `ADD_PROVIDER` queries), but at least worth adding a TODO comment.

dmitry-markin (commented):
The logic seems good with some minor nits. I am going to analyze some general client-server operation consequences before the final approval, but should be good to go.

src/protocol/libp2p/kademlia/mod.rs
lexnv: That makes sense 👍

src/protocol/libp2p/kademlia/config.rs
lexnv: Have left this as a follow-up work 🙏

src/protocol/libp2p/kademlia/mod.rs
lexnv: Have added on the `register_protocol`:
- `InboundProtocol::Accept` - negotiates inbound substreams over multistream-select + advertise over identify
- `InboundProtocol::Deny` - rejects inbound substreams over multistream-select + doesnt advertise over identify

This way a client mode Kademlia won't accept any substreams at all coming from server mode Kademlia (ie maybe an old litep2p version).

### Litep2p Side

Even we matching different versions: old litep2p instance (server mode without proven peer response) <-> litep2p client mode:
- KadMode::client dials and requests a query from old-litep2p 
- old-litep2p doens't add the client mode instance to the routing table unless it is provided as a response peer to a FIND_NODE query
https://github.com/paritytech/litep2p/blob/6e21525cb62a4b7ffeb482c9281783d687dc9d59/src/protocol/libp2p/kademlia/mod.rs#L482-L483


### Substrate Side

Substrate does call into the routing table to add a peer as response to the identify protocol:

https://github.com/paritytech/polkadot-sdk/blob/dda8013292008c5a5d105622cd423d47df342bfe/substrate/client/network/src/litep2p/mod.rs#L1180-L1182

This is then gated by the disjoint list of kad protocols:

https://github.com/paritytech/polkadot-sdk/blob/1407d3feba9d3fc95e20823cd0b74acbe1ca24ee/substrate/client/network/src/litep2p/discovery.rs#L371



I. Case old-litep2p + new litep2p client on the same `/0x1/kad` protocol:
- old-litep2p has `/0x1/kad` advertised
- new client mode doesnt advertise `/0x1/kad`  (client mode deliberately omits this)

II. Case old-litep2p + multi-genesis (speculative messaging)
- old-litep2p has `/0x1/kad` advertised (foreign DHT)
- new client mode for `/0x1/kad` and normal server mode for its own genesis `/0x2/kad` => the advertised list over identify is `/0x2/kad` which doesnt match `/0x1/kad` from old-litep2p

We should be safe here as well

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -232,6 +253,19 @@ impl ConfigBuilder {
         self
     }
 
+    /// Sets the operating mode.
+    ///
+    /// By default, Substrate nodes operate in server mode.
+    ///
```
dmitry-markin:
```suggestion
    /// By default, litep2p operates in server mode.
    ///
```

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -207,6 +227,7 @@ impl ConfigBuilder {
             known_peers: HashMap::new(),
             protocol_names: Vec::new(),
             update_mode: RoutingTableUpdateMode::Automatic,
+            mode: KademliaMode::default(),
```
dmitry-markin:
```suggestion
            mode: KademliaMode::Server,
```
For readability.

src/protocol/libp2p/kademlia/config.rs
```diff
@@ -112,12 +115,24 @@ pub struct Config {
...
 }
 
+/// Kademlia operating mode.
+#[derive(Debug, Copy, Clone, Default, PartialEq, Eq)]
+pub enum KademliaMode {
+    /// Don't answer inbound requests and don't advertise the protocol via identify.
+    Client,
+
+    /// Answer inbound requests and participate fully in the DHT.
+    #[default]
```
dmitry-markin: I would drop the default completely. It should be a deliberate choice of the host process: either force server mode always on, or set it to client first and then upgrade.

```suggestion
```

dmitry-markin (approved):
Good job here!

We should state in the description/changelog that only responding peers are added. As a consequence, I would add `KademliaEvent::PeersDiscovered` event that substrate and other consumers can subscribe to in order to recover the original volume of peers learned. A single Claude Code request will one-shot this in a few minutes.

PR #616, opened 2026-06-11
fixes(webrtc): avoid panics and properly close opening substreams

### LINKED ISSUE #580: webrtc: Panics in `run_event_loop` can leak the connection to upper-layer
The `run_event_loop` of the webrtc connection panics due to extensive usage of unwraps:

https://github.com/paritytech/litep2p/blob/43035be0f78240ed6289a575b1e68202c9f09ff2/src/transport/webrtc/connection.rs#L699-L700

There are a few usages of `unwraps()` which do not propagate the connection-closed event to the higher-level. When this happens, the str0m connection is dropped, but the higher-level protocols will never be informed about it, effectively leaking the connection / causing long term missbehaviors.

1. For example, if the remote peer provided a malformed datagram, the connection panics without informing higher-level protocols that the connection has been closed:

https://github.com/paritytech/litep2p/blob/43035be0f78240ed6289a575b1e68202c9f09ff2/src/transport/webrtc/connection.rs#L784-L794

2. The socket can fail to send data due to numerous (tokio udp could return would block / network unreachable on interface down etc):

https://github.com/paritytech/litep2p/blob/43035be0f78240ed6289a575b1e68202c9f09ff2/src/transport/webrtc/connection.rs#L713

3. Str0m rtc could fail on DTLS datagrams / data that causes errors (like invalid rtp / ice validation):

https://github.com/paritytech/litep2p/blob/43035be0f78240ed6289a575b1e68202c9f09ff2/src/transport/webrtc/connection.rs#L703

4. Handle input:

https://github.com/paritytech/litep2p/blob/43035be0f78240ed6289a575b1e68202c9f09ff2/src/transport/webrtc/connection.rs#L796

Ideally, unwrap / expects should be logged for debugging purposes, the connection should be properly shutdown where possible and the protocols should be informed 

Discovered during reviewing: https://github.com/paritytech/litep2p/pull/573

cc @gab8i @skunert

### LINKED ISSUE #582: webrtc: Opening substreams could leak the connection if RTC is unhealthy
If the RTC is unhealthy or RTC is detected to no longer be connected, the request for `OpenSubstream` by some higher level protocol would not be satisfied. Currently, it is silently ignored.

This should be tackled before releasing the MVP as it could leak the connection at protocol levels and wait for a never-arriving substream.
At the very least we should inform the protocols about the substream failure before ignoring the command.

This may cause protocols to hang indefinitely if a `ConnectionClosed` is never delivered/delivered too late.

```rust
                        // Check if the connection is still healthy before opening new substreams.
                        // This prevents panics when trying to open channels on a shutting-down
                        // SCTP association.
                        if !self.rtc.is_alive() || !self.rtc.is_connected() {
                            tracing::debug!(
                                target: LOG_TARGET,
                                peer = ?self.peer,
                                ?protocol,
                                is_alive = self.rtc.is_alive(),
                                is_connected = self.rtc.is_connected(),
                                "rejecting substream open: connection not healthy",
                            );
```

https://github.com/paritytech/litep2p/pull/573/changes#r3235433126

Followup from reviewing: https://github.com/paritytech/litep2p/pull/573/

cc @gab8i @skunert

gab8i:
Closes #580 
Closes #582

### DIFF
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -162,11 +162,23 @@ impl Stream for SubstreamHandleSet {
             self.index += 1;
 
             let Some((key, _)) = self.handles.get_index(index) else {
+                tracing::debug!(
+                    target: LOG_TARGET,
+                    index,
+                    num_handles = self.handles.len(),
+                    "substream handles index out of bounds",
+                );
                 return Poll::Ready(None);
             };
 
             if !self.pending.contains(key) {
                 let Some((key, stream)) = self.handles.get_index_mut(index) else {
+                    tracing::debug!(
+                        target: LOG_TARGET,
+                        index,
+                        num_handles = self.handles.len(),
+                        "substream handles index out of bounds",
+                    );
                     return Poll::Ready(None);
                 };
 
@@ -1218,7 +1230,7 @@ impl WebRtcConnection {
                 event = self.handles.next() => match event {
                     None => {
                         tracing::warn!(
-                            target: LOG_TARGET, peer = ?self.peer, "substream handle set terminated"
+                            target: LOG_TARGET, peer = ?self.peer, "substream handle set unexpectedly terminated"
                         );
                         return self.on_connection_closed().await;
                     },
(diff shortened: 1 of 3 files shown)


lexnv (approved):
Nicely done here!

dmitry-markin (approved):

Issue #597, opened 2026-05-25
webrtc: Memory leak from connection context


lexnv:
Reproduction path with:
- https://github.com/lexnv/litep2p-perf/pull/4

Webrtc is leaking the connection context:

https://github.com/paritytech/litep2p/blob/9c58647aa59fb920cdc00944c5f96104e760b279/src/transport/webrtc/mod.rs#L144-L145

Git patch applied to detect growth over time:

```
  diff --git a/src/transport/webrtc/mod.rs b/src/transport/webrtc/mod.rs
  index f3a9bcf..52c95a5 100644
  --- a/src/transport/webrtc/mod.rs
  +++ b/src/transport/webrtc/mod.rs
  @@ -152,6 +152,8 @@ pub(crate) struct WebRtcTransport {

       /// Pending events.
       pending_events: VecDeque<TransportEvent>,
  +
  +    debug_timer: tokio::time::Interval,
   }

  @@ -518,6 +520,7 @@ impl TransportBuilder for WebRtcTransport {
                   timeouts: HashMap::new(),
                   pending_events: VecDeque::new(),
                   datagram_buffer_size: config.datagram_buffer_size,
  +                debug_timer: tokio::time::interval(Duration::from_secs(60)),
               },
               listen_multi_addresses,
           ))
  @@ -688,6 +691,18 @@ impl Stream for WebRtcTransport {
       fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
           let this = Pin::into_inner(self);

  +        if this.debug_timer.poll_tick(cx).is_ready() {
  +            tracing::info!(
  +                target: LOG_TARGET,
  +                open_connections = this.open.len(),
  +                opening_connections = this.opening.len(),
  +                connections = this.connections.len(),
  +                timeouts = this.timeouts.len(),
  +                pending_events = this.pending_events.len(),
  +                "[MEMORY STATUS] webrtc transport status",
  +            );
  +        }
  +
```

Detected over long running testing:

```
2026-05-21T15:33:21.781157Z  INFO litep2p::webrtc: [MEMORY STATUS] webrtc transport status open_connections=0 opening_connections=0 connections=0 timeouts=0 pending_events=0
2026-05-21T15:34:21.780563Z  INFO litep2p::webrtc: [MEMORY STATUS] webrtc transport status open_connections=14 opening_connections=26 connections=0 timeouts=26 pending_events=0
[...]

PR #617, opened 2026-06-11
 fix(webrtc): time out inbound and outbound opening data channels

### LINKED ISSUE #587: webrtc: inbound and outbound opening data channels can leak memory
A malicious or broken peer may never complete the handshake protocol and leave either an inbound or outbound channel waiting forever, parked with no mechanism to remove it. As explained [here](https://github.com/paritytech/litep2p/pull/573#discussion_r3261074673), an approach similar to other transport layers, using SUBSTREAM_OPEN_TIMEOUT, should be used.

gab8i:
Closes #587

### DIFF
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -32,7 +32,7 @@ use crate::{
             substream::{Message, Substream as WebRtcSubstream, SubstreamHandle},
             util::{extract_framed_message, WebRtcMessage},
         },
-        Endpoint,
+        Endpoint, SUBSTREAM_OPEN_TIMEOUT,
     },
     types::{protocol::ProtocolName, SubstreamId},
     PeerId,
@@ -247,6 +247,15 @@ pub struct WebRtcConnection {
     /// at most [`MAX_PENDING_PER_CHANNEL`] per channel.
     pending_messages: HashMap<ChannelId, VecDeque<Vec<u8>>>,
 
+    /// Deadlines of the opening phase of channels.
+    deadlines: VecDeque<(ChannelId, Instant)>,
+
+    /// Channels closed by time out.
+    ///
+    /// Need by [`Self::on_channel_closed`], so that
+    /// `NegotiationError::Timeout` can be reported as error.
+    timed_out: HashSet<ChannelId>,
+
     /// Open channels.
     channels: HashMap<ChannelId, ChannelState>,
 
@@ -289,6 +298,9 @@ impl WebRtcConnection {
             dgram_rx,
             pending_outbound: HashMap::new(),
             pending_messages: HashMap::new(),
+            deadlines: VecDeque::new(),
+
+            timed_out: HashSet::new(),
             channels: HashMap::new(),
             handles: SubstreamHandleSet::new(),
             recv_buffers: HashMap::new(),
@@ -327,6 +339,7 @@ impl WebRtcConnection {
                 "inbound channel opened, wait for `multistream-select` message",
             );
 
+            self.add_deadline(channel_id);
             self.channels.insert(
                 channel_id,
                 ChannelState::InboundOpening {
@@ -465,6 +478,15 @@ impl WebRtcConnection {
             "channel closed",
         );
 
+        let timed_out = self.timed_out.remove(&channel_id);
+        let substream_error = || {
+            if timed_out {
+                SubstreamError::NegotiationError(crate::error::NegotiationError::Timeout)
+            } else {
+                SubstreamError::ConnectionClosed
+            }
+        };
+
         // If this was a pending outbound channel (waiting for DCEP ACK from remote),
         // report the failure so the protocol handler can retry.
         if let Some(context) = self.pending_outbound.remove(&channel_id) {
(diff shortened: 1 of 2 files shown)


src/transport/webrtc/connection.rs
```diff
@@ -465,6 +478,15 @@ impl WebRtcConnection {
             "channel closed",
         );
 
+        let timed_out = self.timed_out.remove(&channel_id);
+        let substream_error = || {
```
lexnv, line 497: dq: SubstreamError is not clonable? Could we remove the || block?

gab8i: Nope, it doesn't implement Clone, tbf I'm not sure which is nicer, if this weird closure or just to propagate clone on a bunch of errors + having to deal with `yamux::error::ConnectionError`.

This closure was definitely faster to implement rather than wrapping an enum tag with `#[non_exhaustive]` :sweat_smile:

src/transport/webrtc/connection.rs
```diff
@@ -1292,8 +1321,72 @@ impl WebRtcConnection {
...
                 }
             }
         }
     }
+
+    /// Register an opening-phase deadline for `channel_id`.
+    fn add_deadline(&mut self, channel_id: ChannelId) {
+        self.deadlines.push_back((
+            channel_id,
+            std::time::Instant::now() + SUBSTREAM_OPEN_TIMEOUT,
```
lexnv: instant is already imported


```suggestion
            Instant::now() + SUBSTREAM_OPEN_TIMEOUT,
```

src/transport/webrtc/connection.rs
```diff
@@ -1292,8 +1321,72 @@ impl WebRtcConnection {
...
+        ));
+    }
+
+    /// Close channels whose opening phase exceeded [`SUBSTREAM_OPEN_TIMEOUT`],
+    /// lazily dropping entries for channels that already opened or closed.
+    ///
+    /// If the channel has an SCTP stream, its state is deliberately left untouched so
+    /// the next `on_channel_closed` call will properly handle it.
+    async fn drain_deadlines(&mut self) {
```
lexnv: Could use a test: open a channel, never answer multistream-select, assert the open failure arrives  `NegotiationError::Timeout` within ~5s and its not hanging

src/transport/webrtc/connection.rs
```diff
@@ -1292,8 +1321,72 @@ impl WebRtcConnection {
...
+
+            let channel_to_close = if self.pending_outbound.contains_key(channel_id)
+                || matches!(
+                    self.channels.get(channel_id),
+                    Some(ChannelState::InboundOpening { .. })
+                        | Some(ChannelState::OutboundOpening { .. })
+                ) {
+                *channel_id
+            } else {
+                self.deadlines.pop_front();
```
lexnv: Let's move the `self.deadlines.pop_front();` above this if so we don't have to repeat it in the code

src/transport/webrtc/opening.rs
```diff
@@ -500,6 +500,9 @@ impl OpeningWebRtcConnection {
                                 ?channel_id,
                                 "ignoring opened channel",
                             );
+                            // Do not silently ignore channels opened during the noise handshake.
+                            // Close them instead.
+                            self.rtc.direct_api().close_data_channel(channel_id);
```
dmitry-markin, line 511: nit: May be add the spec reference that we are free do this according to https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md#connection-security?

> Implementations MAY open streams before completion of the Noise handshake. Applications MUST take special care what application data they send, since at this point the peer is not yet authenticated. Similarly, the receiving side MAY accept streams before completion of the handshake.

src/transport/webrtc/connection.rs
```diff
@@ -247,6 +247,15 @@ pub struct WebRtcConnection {
     /// at most [`MAX_PENDING_PER_CHANNEL`] per channel.
     pending_messages: HashMap<ChannelId, VecDeque<Vec<u8>>>,
 
+    /// Deadlines of the opening phase of channels.
+    deadlines: VecDeque<(ChannelId, Instant)>,
```
dmitry-markin: Also, `opening_timed_out`, `add_opening_deadline`, etc.

```suggestion
    opening_deadlines: VecDeque<(ChannelId, Instant)>,
```

dmitry-markin (approved):
This is needed and the implementation seems correct. However, this builds on the already quite intertwined logic of the channel opening with manual bookkeeping & polling, and adds more bookkeeping. The logic is already quite difficult to follow. Ideally, this should be simplified to something as close to `tokio::time::timeout(OPENING_TIMEOUT, opening_channel.drive_opening())` as possible. The latter can be put into `FuturesUnordered`, as we discussed in some of the previous issues.

EDIT: not asking for the refactoring in this PR, we should handle it separately.

gab8i:
I completely agree about reaching a point where it can be `simplified to something as close to tokio::time::timeout(OPENING_TIMEOUT, opening_channel.drive_opening()) as possible`.

I started addressing the comments from oldest to newest, so I found this one and also found it pretty straightforward to implement, that's why I didn't wait to apply this to the refactored code. But yes, what you (@dmitry-markin) describe is the goal of #597! I will move on it soon!

src/transport/webrtc/opening.rs
gab8i: Fair, yes a reference to the spec is definitely worth it!!

lexnv (approved):

PR #618, opened 2026-06-12
Bitswap protocol robustness

skunert:
## Problematic flow

Substrate has a response to send to peer P. No outbound substream to P exists yet.

1. [`on_bitswap_response`](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/bitswap/mod.rs#L444-L450) pushes the response onto `pending_outbound[P]` and calls `open_substream_or_dial(P)`. The open is only triggered on the empty→non-empty transition.

2. [`open_substream_or_dial`](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/bitswap/mod.rs#L362-L395) calls `service.open_substream(P)` which returns `Ok(SubstreamId)`. The `SubstreamId` is discarded.

3. The transport's open times out after [5s](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/transport/mod.rs#L51) and reports [`TransportEvent::SubstreamOpenFailure`](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/protocol_set.rs#L386-L390) carrying only `SubstreamId` (no `PeerId`).

4. The bitswap event loop has no arm for `SubstreamOpenFailure`. It falls into [the catch-all `trace!` at mod.rs:474](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/bitswap/mod.rs#L458-L475) and is dropped.

5. `pending_outbound[P]` is only drained on `SubstreamOpened` ([on_outbound_substream](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/bitswap/mod.rs#L302-L335)). That event never comes.

6. The next response for P hits the same code as step 1. `pending_actions.is_empty()` is now false → **`open_substream_or_dial` is not called again**. Responses just keep stacking onto a dead queue.

Result: P stops getting served. Other peers are unaffected (their `outbound` substreams are healthy). `PeerId` is stable across pod restarts so restarting the remote doesn't help — only restarting the local node clears `pending_outbound`.

For comparison, [kademlia](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/kademlia/mod.rs#L1064-L1068), [request_response](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/request_response/mod.rs#L922-L933) and [ping](https://github.com/paritytech/litep2p/blob/9cca92c078327cc9ea63d47d21a070cd3b0d7e3d/src/protocol/libp2p/ping/mod.rs#L217) all handle `SubstreamOpenFailure`.

### Fix

Two independent changes, either alone would have prevented this:

- Handle `SubstreamOpenFailure` (also `DialFailure`, `ConnectionClosed`) in the bitswap event loop, dropping `pending_outbound[P]` on failure. Requires a `SubstreamId → PeerId` map since the failure event carries no peer.
- Or: retry `open_substream_or_dial(P)` after a timeout if `pending_outbound[P]` hasn't drained.

### DIFF
--- a/src/protocol/libp2p/bitswap/mod.rs
+++ b/src/protocol/libp2p/bitswap/mod.rs
@@ -365,11 +365,7 @@ impl Bitswap {
         if !self.pending_dials.remove(&peer) {
             return;
         }
-        let dropped = self
-            .pending_outbound
-            .remove(&peer)
-            .map(|actions| actions.len())
-            .unwrap_or(0);
+        let dropped = self.pending_outbound.remove(&peer).map(|actions| actions.len()).unwrap_or(0);
         tracing::debug!(
             target: LOG_TARGET,
             ?peer,
@@ -401,11 +397,7 @@ impl Bitswap {
             return;
         };
 
-        let dropped = self
-            .pending_outbound
-            .remove(&peer)
-            .map(|actions| actions.len())
-            .unwrap_or(0);
+        let dropped = self.pending_outbound.remove(&peer).map(|actions| actions.len()).unwrap_or(0);
 
         tracing::debug!(
             target: LOG_TARGET,
--- a/src/protocol/libp2p/bitswap/mod.rs
+++ b/src/protocol/libp2p/bitswap/mod.rs
@@ -344,23 +344,19 @@ impl Bitswap {
 
     /// Handle connection closed event.
     ///
-    /// The outbound substream cached in [`Self::outbound`] now references a dead transport;
-    /// drop it so the next request/response triggers a fresh open. We deliberately do NOT touch
-    /// [`Self::pending_outbound`] or [`Self::pending_substreams`] here: any in-flight open on
-    /// the closed connection will surface as [`TransportEvent::SubstreamOpenFailure`], which
-    /// cleans both maps via [`Self::on_substream_open_failure`].
+    /// Drop all peer-scoped outbound state so the next request/response can open a fresh
+    /// substream instead of waiting behind state tied to a closed connection.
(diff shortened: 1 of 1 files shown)


src/protocol/libp2p/bitswap/mod.rs
```diff
@@ -334,6 +342,81 @@ impl Bitswap {
...
+            tracing::debug!(target: LOG_TARGET, ?peer, "dropping outbound substream on connection close");
+        }
+        self.inbound.remove(&peer);
+    }
+
+    /// Dial to remote peer failed.
+    ///
+    /// We only enter `pending_dials` from [`Self::open_substream_or_dial`] when we already had
+    /// queued actions in [`Self::pending_outbound`]. Without this handler, both would stay
+    /// populated forever — same shape as the [`Self::on_substream_open_failure`] bug.
```
dmitry-markin:
```suggestion
    /// populated forever.
```
nit: there will be no bug to refer the reader to after the PR is merged.

dmitry-markin (approved):
Good catch!

PR #620, opened 2026-06-17
fix(webrtc): build vendored OpenSSL for str0m

gab8i:
Enable str0m's `vendored` feature so OpenSSL is compiled from source instead of linking against system OpenSSL, which is unavailable on some targets.


dmitry-markin (approved):
Do we have other indirect dependencies on OpenSSL with other features enabled like `websocket` (we can ignore `quic` for now)?

gab8i:
I'll dig deeper now, but since no issue was found when compiling for other targets before this one, I imagine that every other dependency also used a 'vendored' openssl instead of fetching the system one by default, as str0m does.

For example `tokio-tungstenite` uses the feature flag `rustls-tls-native-roots` which uses rustls, no openssl

lexnv (approved):

PR #621, opened 2026-06-17
feat(webrtc): WebRtcTransport Stream now uses shared 16KiB buffer

### LINKED ISSUE #588: webrtc: Stream implementation of `WebRtcTransport` allocates 16KiB per packet
https://github.com/paritytech/litep2p/blob/4b5487c465eedfb0aed012b897a49ce9af43a9dc/src/transport/webrtc/mod.rs#L695-L696

Optimize the memory allocation by reusing the vector if necessary 

Discovered during a review of webrtc:
- https://github.com/paritytech/litep2p/pull/586

cc @gab8i @skunert

gab8i:
Closes #588

### DIFF
--- a/src/transport/webrtc/mod.rs
+++ b/src/transport/webrtc/mod.rs
@@ -34,6 +34,7 @@ use crate::{
     PeerId,
 };
 
+use bytes::{Bytes, BytesMut};
 use futures::{future::BoxFuture, stream::FuturesUnordered, Future, Stream, StreamExt};
 use futures_timer::Delay;
 use hickory_resolver::TokioResolver;
@@ -99,7 +100,7 @@ struct ConnectionContext {
     connection_id: ConnectionId,
 
     /// TX channel for sending datagrams to the connection event loop.
-    tx: Sender<Vec<u8>>,
+    tx: Sender<Bytes>,
 }
 
 /// Events received from opening connections that are handled
@@ -165,6 +166,9 @@ pub(crate) struct WebRtcTransport {
 
     /// Pending events.
     pending_events: VecDeque<TransportEvent>,
+
+    /// Read buffer.
+    read_buffer: BytesMut,
 }
 
 impl WebRtcTransport {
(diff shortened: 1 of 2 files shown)


lexnv (approved):

src/transport/webrtc/mod.rs
```diff
@@ -461,6 +464,7 @@ impl TransportBuilder for WebRtcTransport {
                 timeouts: HashMap::new(),
                 pending_events: VecDeque::new(),
                 datagram_buffer_size: config.datagram_buffer_size,
+                read_buffer: BytesMut::new(),
```
dmitry-markin, line 467:
```suggestion
                read_buffer: BytesMut::with_capacity(WEBRTC_BUFFER_SIZE),
```

src/transport/webrtc/mod.rs
```diff
@@ -659,8 +663,9 @@ impl Stream for WebRtcTransport {
         }
 
         loop {
-            let mut buf = vec![0u8; 16384];
-            let mut read_buf = ReadBuf::new(&mut buf);
+            // Prepare the buffer to accept up to 16KiB.
+            this.read_buffer.resize(16 * 1024, 0);
+            let mut read_buf = ReadBuf::new(&mut this.read_buffer);
```
dmitry-markin, line 668: We don't need to initialize the buffer with zeros. Given all the read bytes are consumed every time with `split_to`, we can assert the size is zero (and it will reclaim the original buffer if the event loop has already processed the datagram and dropped the reference). Also, no need to wrap into tokio buffer, as we can read into uninitialized `BytesMut` of sufficient capacity directly (see below).
```suggestion
            assert!(this.read_buffer.len(), 0)
            this.read_buffer.reserve(WEBRTC_BUFFER_SIZE);
```

src/transport/webrtc/mod.rs
```diff
@@ -659,8 +663,9 @@ impl Stream for WebRtcTransport {
...
 
             let addrs = match this.listener.poll_recv_from(cx, &mut read_buf) {
```
dmitry-markin, line 670:
```suggestion
            let addrs = match this.listener.recv_buf_from(cx, &mut this.read_buffer) {
```

dmitry-markin (commented):
I have left some suggestions for improvements, but realized they don't make much sense because the original buffer will only be reused if the event loop happens to process the datagram and drop the reference before the next one is read.

We can't expect this to happen. Instead, what we should do — process the read buffer straight after reading from a socket (i.e., decode the protobuf), release the buffer for the next datagram to come and be written in place, and send over the channel not the reference to the original buffer but the already decoded data. We need to allocate space for the decoded data anyway, so this won't incur any new overhead.

gab8i:
Closing in favor of #623 which is just cleaner and a more effective solution

PR #622, opened 2026-06-17
fix(webrtc): decode errors during opening phase are not recoverable

### LINKED ISSUE #589: webrtc: Transport doesn't accept a flag-only frame
https://github.com/paritytech/litep2p/blob/4b5487c465eedfb0aed012b897a49ce9af43a9dc/src/transport/webrtc/connection.rs#L387

https://github.com/paritytech/litep2p/blob/4b5487c465eedfb0aed012b897a49ce9af43a9dc/src/transport/webrtc/util.rs#L94-L113

We could receive no payload and flags as `Fin` `ResetStream` / `StopSending`.

For `Fin`/`ResetStream` we should propagate the `ConnectionClsoed` to higher levels. While `StopSending` or no flag without frames should be treated as no-ops.

Discovered during a review of webrtc:
- https://github.com/paritytech/litep2p/pull/586

cc @gab8i @skunert

gab8i:
Explanation of why being so strict over decoding errors within the opening phases can be found [here](https://github.com/paritytech/litep2p/issues/589#issuecomment-4733498232).

Closes #589

### DIFF
--- a/src/transport/webrtc/connection.rs
+++ b/src/transport/webrtc/connection.rs
@@ -581,7 +581,21 @@ impl WebRtcConnection {
         );
 
         // Decode errors are not recoverable.
-        let payload = WebRtcMessage::decode(&data)?.payload.ok_or(Error::InvalidData)?;
+        let WebRtcMessage {
+            payload: Some(payload),
+            flag: None,
+        } = WebRtcMessage::decode(&data)
+            .map_err(|err| SubstreamError::NegotiationError(err.into()))?
+        else {
+            tracing::debug!(
+                target: LOG_TARGET,
+                peer = ?self.peer,
+                ?channel_id,
+                "non-payload frame during inbound opening, closing channel"
+            );
+            return Err(Error::ConnectionClosed);
+        };
+
         let protocols = self.protocol_set.protocols_with_keep_alives();
         let protocol_names = protocols.keys().cloned().collect();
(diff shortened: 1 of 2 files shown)


lexnv (approved):

src/transport/webrtc/connection.rs
```diff
@@ -581,7 +581,21 @@ impl WebRtcConnection {
...
+            payload: Some(payload),
+            flag: None,
+        } = WebRtcMessage::decode(&data)
+            .map_err(|err| SubstreamError::NegotiationError(err.into()))?
+        else {
+            tracing::debug!(
+                target: LOG_TARGET,
+                peer = ?self.peer,
+                ?channel_id,
+                "non-payload frame during inbound opening, closing channel"
```
dmitry-markin, line 594: nit: it is a good idea to add some context to the message, to know which log exactly was triggered by looking at the message. Currently all three log messages are the same.

dmitry-markin (approved):

src/transport/webrtc/connection.rs
gab8i: They are not the same, 'during inbound/outbount opening/noise handshake', the structure is definitely the same though

dmitry-markin: Aaah, sorry, missed this.

PR #626, opened 2026-06-23
fix(webrtc): input backpressure

### LINKED ISSUE #604: webrtc: input backpressure
Between a `SubstreamHandle` and its corresponding `Substream`, there is a bounded channel used to forward messages. The bound is set to 256, and currently the `SubstreamHandle` waits until space becomes available within the channel, this implies that a slow reader can cause the handle to wait indefinitely. The problem is that `on_message` is called directly from the main RTC event loop, which means that a stall in a single substream, due to a slow reader, causes the entire connection to stall.

The solution would be to implement read backpressure by keeping track of pending incoming messages.

gab8i:
The problem can be summarized in: inbound messages are forwarded from each `SubstreamHandle` to its `Substream` over a bounded channel using a blocking `send().await`, this imply that a slow `Substream` reader can stall the entire connection. 

Real backpressure though isn't possible, reads come off one shared SCTP association, not per stream. You can't slow down a single reader, you would end up slowing down all substreams (original issue).

So at the end the solution that made more sense to me was to enforce a resource ceiling instead of chasing backpressure.

Earlier attempts added explicit pending-message accounting and pause/resume machinery to emulate some sort of per stream backpressure but at the end they all ended up with a limited buffer waiting to be exceeded. This solution reaches the same point but just with the least change as possible. 

Closes #604

### DIFF
--- a/src/transport/webrtc/substream.rs
+++ b/src/transport/webrtc/substream.rs
@@ -1771,4 +1771,77 @@ mod tests {
             other => panic!("Expected BrokenPipe error, got: {:?}", other),
         }
     }
+
+    #[tokio::test]
+    async fn graceful_drop_does_not_reset_substream() {
+        let (mut substream, mut handle) = Substream::new();
+
+        // Drive the write half to FinAck (StopSending -> FIN -> FIN_ACK).
+        handle
+            .on_message(WebRtcMessage {
+                payload: None,
+                flag: Some(Flag::StopSending),
+            })
+            .await
+            .unwrap();
+        assert_eq!(
+            handle.next().await,
+            Some(Message {
+                payload: vec![],
+                flag: Some(Flag::Fin)
+            })
+        );
+        handle
+            .on_message(WebRtcMessage {
+                payload: None,
+                flag: Some(Flag::FinAck),
+            })
+            .await
+            .unwrap();
+        assert!(matches!(handle.writer_state.get(), WriterState::FinAck));
+
+        // Drive the read half to FinAck (peer FIN -> we emit FIN_ACK).
+        handle
+            .on_message(WebRtcMessage {
+                payload: None,
+                flag: Some(Flag::Fin),
+            })
+            .await
+            .unwrap();
+        assert_eq!(
+            handle.next().await,
+            Some(Message {
+                payload: vec![],
+                flag: Some(Flag::FinAck)
+            })
+        );
+        assert!(matches!(handle.reader_state.get(), ReaderState::FinAck));
+
+        // Both halves gracefully closed -> stream is done.
+        assert_eq!(handle.next().await, None);
+
+        // Dropping the gracefully-closed handle MUST NOT reset the channel.
+        drop(handle);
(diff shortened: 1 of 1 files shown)


src/transport/webrtc/substream.rs
```diff
@@ -287,25 +300,31 @@ impl SubstreamHandle {
...
+                    Err(TrySendError::Full(_)) => {
+                        // If exceeded, the peer is sending faster than the reader can consume
+                        // and thus this can be considered a DoS attack by forcing us to buffer
+                        // those messages.
+                        tracing::debug!(
+                            target: LOG_TARGET,
+                            max_pending_inbound_messages_per_channel = MAX_INFLIGHT_MESSAGES,
+                            "max number of pending inbound messages exceeded"
+                        );
+                        self.channel_state.set(ChannelState::Reset);
```
lexnv: In this case we are not emitting a `RESET_STREAM` flag message. Would it be possible to send one before dropping the `SubstreamHandle`?

src/transport/webrtc/substream.rs
```diff
@@ -528,6 +547,8 @@ impl tokio::io::AsyncRead for Substream {
         }
 
         match futures::ready!(self.rx.poll_recv(cx)) {
+            None if matches!(self.channel_state.get(), ChannelState::Reset) =>
```
lexnv, line 574: nit: We are also setting the state to `Reset` while dropping the `SubstreamHandle`. It could happen that on the `on_message(FIN)` + FinAck handling we return `Err(ConnectionReset)` instead of `Ok()` as before

src/transport/webrtc/substream.rs
gab8i: Yes, definitely it should be done. It will not be straight forward to send because this happens within the 'on_message' and thus not in the stream where output messages are created!

src/transport/webrtc/substream.rs
gab8i: Yes, you are definitely right, I added this tangential change within this commit without having carefully in mind all the implications. 

I think that we can keep this behavior but fixing the issue you described by changing the `drop` into something like:

```rust
           let graceful = matches!(self.writer_state.get(), WriterState::FinAck)
             && matches!(self.reader_state.get(), ReaderState::FinAck);
           if !graceful {
               self.channel_state.set(ChannelState::Reset);
      }
```

I will add also some unit test to make sure no regression is inserted!

gab8i:
@lexnv last 3 commits should have addressed your comments!

src/transport/webrtc/substream.rs
```diff
@@ -287,25 +311,33 @@ impl SubstreamHandle {
...
-                    })
-                    .await;
-
-                if let Err(err) = send_result {
-                    tracing::debug!(
-                        target: LOG_TARGET,
-                        ?err,
-                        "failed to propagate message to Substream"
-                    );
+                let send_result = message_tx.try_send(Message {
```
lexnv, line 314: nit: Could be good adding a unit test for the full path, not sure how complicated that would turn out

lexnv (approved):

PR #628, opened 2026-06-30
ci: Add CI step for webrtc compatibility with libp2p-go

lexnv:
This PR adds a new CI step to check the compatibility of webrtc implementation with libp2p-go client from litep2p-perf: https://github.com/lexnv/litep2p-perf.

The CI step:
- clones the repo and replaces litep2p dependency to github PR one
- spawn the webrtc litep2p server
- extract the certhash from the running logs
- tests connectivity, substream code paths, and compatibility with the libp2p-go client

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,6 +222,12 @@ jobs:
       - name: Install Rust stable toolchain
         uses: dtolnay/rust-toolchain@stable
 
+      - name: Install protoc
+        # litep2p's build.rs runs prost-build, which needs the protobuf
+        # compiler. The repo's other jobs get it from the ci-unified container;
+        # this job runs on bare ubuntu-latest, which does not ship protoc.
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
+
       - name: Rust Cache
         uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,10 +206,10 @@ jobs:
       # Address the litep2p WebRTC server listens on. litep2p advertises a
       # server-listenable transport as `/webrtc-direct` (multiaddr code 280).
       SERVER_LISTEN_ADDR: /ip4/127.0.0.1/udp/33333/webrtc-direct
-      # Peer id deterministically derived from `--node-key secret` (see the perf
-      # repo's README / run_bandwidth.sh). Used as a fallback to build the client
-      # address if the server's logged listen address omits the /p2p/<peer-id>.
+      # Starting the server with `--node-key secret` makes its identity fully
+      # deterministic. Both vaulues are therefore stable across runs.
       SERVER_PEER_ID: 12D3KooWBpZHDZu7YSbvPaPXKhkRNJvR7MkTJMQQAVBKx9mCqz3q
+      SERVER_CERTHASH: uEiDQNMVmnSW3qoB31dz_DgLjHsktkWIotqfYTBeKqNO5IQ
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -190,3 +190,162 @@ jobs:
...
+          cleanup() {
+            echo "===== client.log (go-libp2p) ====="
+            cat "$CLIENT_LOG" 2>/dev/null || echo "(no client log)"
+            echo "===== server.log (litep2p) ====="
+            cat "$SERVER_LOG" 2>/dev/null || echo "(no server log)"
+            kill "$SERVER_PID" 2>/dev/null || true
+          }
+          trap cleanup EXIT
+
+          # Wait for the server to advertise its /webrtc-direct/certhash address
```
skunert: I wonder if this is necessary, litep2p takes the cert from the outside right? Could we generate a known certificate and have it being known? Then we would not need to parse the hash here. But only a nit, can be follow up or accept status-quo

skunert (approved):

gab8i (approved):

.github/workflows/ci.yml
lexnv: Yep makes sense, the certhash and peerID are already statically generated from the `--node-secret` 🙏 Have updated thanks!

.github/workflows/ci.yml
```diff
@@ -190,3 +190,156 @@ jobs:
...
+    # its WebRTC server, and drive it with the go-libp2p perf client (see
+    # lexnv/litep2p-perf#4). A failed dial/handshake/transfer fails the job, so
+    # this guards litep2p's WebRTC server against go-libp2p interop regressions.
+    name: WebRTC interop (go-libp2p client)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      # Performance harness that contains the litep2p server and go-libp2p client.
+      PERF_REPO: https://github.com/lexnv/litep2p-perf
+      PERF_REF: master
```
lrubasze, line 205: Should we pin some revision here to not break this CI in case `master` breaks?

.github/workflows/ci.yml
```diff
@@ -190,3 +190,156 @@ jobs:
...
+          done
+          if ! grep -q "Server listening on address" "$SERVER_LOG"; then
+            echo "ERROR: perf server did not start listening in time"
+            exit 1
+          fi
+
+          # Drive the server with the go-libp2p client. A non-zero exit (failed
+          # dial, handshake, or transfer) propagates through `tee` via pipefail
+          # and fails this step. The client log is streamed live and saved.
+          ( cd libp2p-go && go mod download && timeout 120 go run . \
```
lrubasze, line 324: Minor.
Not sure how long does it take to build  `libp2p-go`.
But maybe we could build it outside this 120s timeout and use the timeout purely for running the client?

.github/workflows/ci.yml
```diff
@@ -190,3 +190,156 @@ jobs:
...
+    timeout-minutes: 30
+    env:
+      # Performance harness that contains the litep2p server and go-libp2p client.
+      PERF_REPO: https://github.com/lexnv/litep2p-perf
+      PERF_REF: master
+      # Address the litep2p WebRTC server listens on. litep2p advertises a
+      # server-listenable transport as `/webrtc-direct` (multiaddr code 280).
+      SERVER_LISTEN_ADDR: /ip4/127.0.0.1/udp/33333/webrtc-direct
+      # Starting the server with `--node-key secret` makes its identity fully
+      # deterministic. Both vaulues are therefore stable across runs.
```
lrubasze, line 210: nit. vaulues -> values

lrubasze (commented):
PR already merged, but left some minor comments anyway.

PR #631, opened 2026-07-03
fix(noise): size handshake read buffer with correct overhead

### LINKED ISSUE #332: crypto/noise: Preallocate handshake memory with correct overhead
Allocate the appropriate amount of memory instead of hardcoding a random value:

https://github.com/paritytech/litep2p/blob/2a49fd0acfa767183f5a6f3b5c116e0c653e6252/src/crypto/noise/mod.rs#L252-L262

dimartiro:
## Description

`read_handshake_message` allocated its decryption output buffer as `message.len() + 200` — an arbitrary overhead flagged with a `TODO`.

Closes #332

### DIFF
--- a/src/crypto/noise/mod.rs
+++ b/src/crypto/noise/mod.rs
@@ -257,9 +257,10 @@ impl NoiseContext {
         let mut message = BytesMut::zeroed(size as usize);
         io.read_exact(&mut message).await?;
 
-        // TODO: https://github.com/paritytech/litep2p/issues/332 use correct overhead.
-        let mut out = BytesMut::new();
-        out.resize(message.len() + 200, 0u8);
+        // Noise decryption strips the 16-byte AEAD (Poly1305) tag, so the decrypted
+        // plaintext is never larger than the ciphertext. `message.len()` is therefore
+        // a guaranteed-sufficient upper bound for the output buffer.
+        let mut out = BytesMut::zeroed(message.len());
 
         let NoiseState::Handshake(ref mut noise) = self.noise else {
             tracing::error!(target: LOG_TARGET, "invalid state to read handshake message");


src/crypto/noise/mod.rs
```diff
@@ -257,9 +257,10 @@ impl NoiseContext {
         let mut message = BytesMut::zeroed(size as usize);
         io.read_exact(&mut message).await?;
 
-        // TODO: https://github.com/paritytech/litep2p/issues/332 use correct overhead.
-        let mut out = BytesMut::new();
-        out.resize(message.len() + 200, 0u8);
+        // Noise decryption strips the 16-byte AEAD (Poly1305) tag, so the decrypted
+        // plaintext is never larger than the ciphertext. `message.len()` is therefore
+        // a guaranteed-sufficient upper bound for the output buffer.
+        let mut out = BytesMut::zeroed(message.len());
```
lexnv, line 263: Could you please double check for running nodes in Kusama and Polkadot? 🙏 I would simply place info logs for before and after with the actual length.

IIRC the actual message we receive is bigger than the `message.len()` hence the added `200`, which would otherwise trigger a reallocation and affect performance

dimartiro: The `out` buffer holds the **decrypted plaintext**, not the received frame — and in Noise the plaintext is always smaller than the ciphertext, so `message.len()` can't be exceeded. I traced this in snow 0.9.6:

- `_read_message` writes to `out` only in the final `decrypt_and_mix_hash(ptr, payload)` (`handshakestate.rs:445`); the ephemeral/static DH keys go to internal buffers (`self.re`/`self.rs`), never to `out`. `ptr` is a suffix of `message`, so `ptr.len() <= message.len()`.
- The size check is `out.len() >= ciphertext.len() - TAGLEN` with `TAGLEN = 16` (`cipherstate.rs:53`), so `out.len() = message.len()` is sufficient with margin.
- `read_message` takes a fixed `&mut [u8]`; an undersized buffer returns `Error::Decrypt`, never a reallocation. So the old `+200` was pure over-allocation, not realloc-avoidance.

Happy to add temporary before/after info logs on a Kusama/Polkadot node if you'd still like empirical confirmation, but the bound is provable from the snow source.

PR #637, opened 2026-08-05
webrtc: strict client network identity verification

dmitry-markin:
Ensure a client's Noise static key is signed with the node key when accepting WebRTC connections.


lexnv (approved):
LGTM! Nice done! 🙏

PR #640, opened 2026-08-11
webrtc: support wildcard listen addresses

### LINKED ISSUE #638: webrtc: support listening on wildcard addresses
In order to allow operators of substrate-based nodes to easily enable WebRTC protocol on default `--port 30333`, we must allow listening for incoming UDP packets on wildcard addresses like `0.0.0.0:30333` & `:::30333`.

CC @gab8i

dmitry-markin:
Allow wildcard IPv4 & IPv6 WebRTC listen addresses, taking care of local addresses of inbound/outbound UDP packets.

Resolves https://github.com/paritytech/litep2p/issues/638.

### DIFF
--- a/src/transport/webrtc/listener.rs
+++ b/src/transport/webrtc/listener.rs
@@ -73,7 +73,7 @@ impl WebRtcListener {
             };
 
             listen_sockets.push((sockaddr, Arc::new(socket)));
-            listen_multi_addresses.extend(Self::build_listen_addresses(sockaddr, certhash)?);
+            listen_multi_addresses.extend(Self::build_listen_addresses(sockaddr, certhash));
         }
 
         if listen_sockets.is_empty() {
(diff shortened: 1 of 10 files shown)


src/transport/webrtc/listener.rs
```diff
@@ -99,18 +114,79 @@ impl WebRtcListener {
         ))
     }
 
-    pub(super) fn socket(&self, local: &SocketAddr) -> Option<Arc<UdpSocket>> {
+    /// Build the multiaddresses to advertise for a socket bound to `sockaddr`.
+    ///
+    /// A wildcard listen address is advertised as all matching interface addresses (link-local IPv6
+    /// excluded).
+    fn build_listen_addresses(sockaddr: SocketAddr, certhash: Multihash<64>) -> Vec<Multiaddr> {
```
gab8i, line 121: loopback addresses are not filtered out here, is this made on purpose?

gab8i (commented):
It took me a while to understand effectively how does this work but if I got this right this is a great way of using the fact that the ICE candidates IP doesn't really need to be the same as the reachable address

Correct me if I'm wrong:
- listener addresses for wildcards are expanded into multiple multiaddress, one per effectively reachable interface (ip)
- using `quinn_udp::UdpSocketState` allow and auto set up of `IP_PKTINFO` There is a difference between the reachable ip and what needs to be used within str0m, the latter just needs to stay stable over the connection.  This makes possible to keep a node under a proxy, the reachable addr is A but internally the first message arrives at ip B (known thanks to `IP_PKTINFO`),
- An `AddrsPair` can be build knowing B, the first ip a message arrive at, and the remote addr. This pair is now used to force always the usage of B as ip source for sent messages and to internally dispatch messages to the right webrtc instances


One thing I didn't understand though is: what if the destination address changes mid-session?

Behind a wildcard, peers reach the node at A and the internally A → B.
With more than one internal address that can resolve to B1 at handshake time and B2 later,
how is this addressed? It this not possible?

src/transport/webrtc/listener.rs
dmitry-markin: This matches our approach with TCP/WebSocket. Loopback addresses deprioritized later in DHT, but we advertise them over `Identify` protocol.

dmitry-markin:
Yes, everything works as you described.

> With more than one internal address that can resolve to B1 at handshake time and B2 later

This is not possible if the routing is stable. Even if the routing changes mid-session, this requires at least two local addresses/interfaces B1 B2 both reachable from the public IP A, which is a very weird setup.

gab8i (approved):

PR #646, opened 2026-08-21
webrtc: drop non-STUN/DTLS datagrams at ingress and str0m bump

gab8i:
Add a filter at the top of `on_socket_input`, the single entry point for every inbound datagram. STUN and DTLS are the only kinds accepted, so anything else is dropped before it can reach str0m's parsers.

This also includes a `str0m` bump

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3445,9 +3445,9 @@ dependencies = [
 
 [[package]]
 name = "sctp-proto"
-version = "0.10.3"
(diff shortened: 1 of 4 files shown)


lexnv (approved):

dmitry-markin (approved):

dmitry-markin:
Cherry-picked a clippy fix that was originally made for another branch.

PR #648, opened 2026-08-24
deps: bump `str0m` v0.22.0 -> v0.23.1

dmitry-markin:


dmitry-markin:
Missed this is already done in https://github.com/paritytech/litep2p/pull/646#top, closing.

PR #652, opened 2026-08-24
fix(webrtc): always drive substream teardown (backpressure + drop leaks)

gab8i:


dmitry-markin (approved):

src/transport/webrtc/substream.rs
```diff
@@ -549,10 +587,37 @@ impl Stream for SubstreamHandle {
...
+        // The writer has concluded but the peer never sent its FIN.
+        // Bound the wait and escalate to a reset instead of staying
+        // half-open until the connection dies.
+        //
+        // NOTE: closing the write half and continuing to read is a legitimate pattern,
+        // thus this is only applied if `Substream` is not alive anymore.
+        if self.message_tx.as_ref().is_none_or(|tx| tx.is_closed()) {
+            let timeout = self
+                .read_close_timeout
+                .get_or_insert_with(|| Box::pin(tokio::time::sleep(FIN_ACK_TIMEOUT)));
```
lexnv, line 599: Are the timeouts ever diverging? Fin vs read ones?

src/transport/webrtc/connection.rs
```diff
@@ -1207,7 +1209,20 @@ impl WebRtcConnection {
...
+                                tracing::debug!(
+                                    target: LOG_TARGET,
+                                    peer = ?self.peer,
+                                    ?channel_id,
+                                    ?error,
+                                    "failed to flush pending messages, closing channel",
+                                );
+
+                                self.rtc.direct_api().close_data_channel(channel_id);
+                                self.channels.insert(channel_id, ChannelState::Closing);
```
lexnv, line 1263: Shouldnt we announce the substream open failure here if:

```rust
if let Some(ChannelState::OutboundOpening { context, .. }) = self.channels.insert(channel_id, ChannelState::Closing)

```

src/transport/webrtc/substream.rs
```diff
@@ -529,10 +558,19 @@ impl Stream for SubstreamHandle {
         //   simply be forwarded.
         let writer_state_stream_result = match self.writer_state.get() {
             WriterState::Open => {
-                match self.message_rx.poll_recv(cx) {
-                    // Writes are finished, start half close procedure.
-                    Poll::Ready(None) => self.poll_half_close(cx),
-                    res => res,
+                if pull_payload {
+                    match self.message_rx.poll_recv(cx) {
```
lexnv, line 562: Could we now get into a situation where:
- higher levels drop `Substream` while we have `message_rx` pending
- channel remains pending and now one can trigger `ChannelBufferedAmountLow`?

A peer that accepts SCTP traffic but keeps the send window full could the channel + handle + 4 MiB of buffered payloads until the connection dies

Could be a followup, maybe easiest would be to arm the `read_close_timeout` outside of the writter state

src/transport/webrtc/substream.rs
gab8i: They are used in different context, your comments made me think that I used the same constant (which makes this less clear) and effetively they can be squished into a single timer but the differences are
- the fin timer starts after having sent a FIN and the timers run while we expect a FIN_ACK back
- this read_close_timer stars if the substream has been dropped and thus we closed our read half (fin->fin_ack handshake) and also we completely closed our read side thus a STOP_SENDING has been sent to the user and now this timer starts and we expect a FIN by the other peer. If the substream is not alive anymore it doesn't make sense to keep alive a connection, this timer just sends a RESET to the other peer and close completely the connection.

src/transport/webrtc/substream.rs
```diff
@@ -512,6 +528,19 @@ impl Stream for SubstreamHandle {
             }
         }
 
+        // Substream has been dropped while the peer's write half is
+        // still open, tell the peer we will no longer read.
+        if !self.stop_sending_sent
+            && matches!(self.reader_state.get(), ReaderState::Open)
+            && self.message_tx.as_ref().is_some_and(|tx| tx.is_closed())
```
lexnv, line 535: nit: We guard the `StopSending` via `tx.is_closed()` . Usually, `Substream` drop wakes the connection task via `message_rx.poll_recv`. But now a channel can enter the pending state via `pull_payload`and therefore we never poll `message_rx.poll_recv` to give it a chance to save the context waker.

Then we drop the `Substream` and no waker is woken. I think this is fine since str0m timeouts will bound this to seconds instead of stalling

Maybe we can improve it somehow via polling a boxed `Sender::closed` fn?

src/transport/webrtc/connection.rs
```diff
@@ -1207,7 +1209,20 @@ impl WebRtcConnection {
...
+                                self.handles.remove(&channel_id);
+                                self.pending_messages.remove(&channel_id);
```
lexnv, line 1265: What would happen if as pending frames we got `FIN_ACK` or `RESET_STREAM`? Would that mean that on backpressure the other side/peer will default to 10s timeouts to close the connection gracefully since we never handled those to str0m?

src/transport/webrtc/connection.rs
gab8i: Sorry I'm not following this. What do you mean?

I think  `ChannelBufferedAmountLow` are emitted only for already established channels thus I don't think the previous state of the channel can be `OuboundOpening` here. I will dig deeper into this, maybe I'm misunderstanding something!

lexnv (approved):
LGTM! Probably the only one worth investigating for now is https://github.com/paritytech/litep2p/pull/652#discussion_r3844364108, since the others are present also before this PR

If it turns out too complicated to tackle feel free to create some issues out of them if they still make sense so we never lose track of it (some could be good onboarding issues for new joiners in the future)

Nicely done here 🙏

src/transport/webrtc/connection.rs
lexnv: Should be a consistency thingy, in other places like `drain_opening_deadlines` we call:

```
                let _ = self
                    .protocol_set
                    .report_substream_open_failure(
                        context.protocol,
                        context.substream_id,
                        SubstreamError::NegotiationError(crate::error::NegotiationError::Timeout),
                    )
                    .await;
            }
            
```

if we are in outboundOpening, but if that state cant happen here i think we should be fine

gab8i: ah, I see now! thanks!

Yes, this is easy to address, nothing structural, just a better safe than sorry approach

PR #657, opened 2026-08-24
fix(webrtc): bound teardown of dropped substreams under backpressure

gab8i:
address: https://github.com/paritytech/litep2p/pull/652#discussion_r3844364108

### DIFF
--- a/src/transport/webrtc/substream.rs
+++ b/src/transport/webrtc/substream.rs
@@ -50,7 +50,7 @@ const FIN_ACK_TIMEOUT: Duration = Duration::from_secs(10);
 /// Timeout bounding the entire remaining teardown once the `Substream`
 /// has been dropped: payload flush, FIN/FIN_ACK and the peer's
 /// FIN must all complete within this window, otherwise the substream is reset.
-/// 
+///
 /// Mirrors [`FIN_ACK_TIMEOUT`] and go-libp2p's stream close timeout.
 const TEARDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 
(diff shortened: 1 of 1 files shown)


src/transport/webrtc/substream.rs
```diff
@@ -492,6 +499,18 @@ impl SubstreamHandle {
...
+    /// Poll the teardown deadline, arming it on first use.
+    ///
+    /// Called once the `Substream` has been dropped, or
+    /// its buffered payload can no longer be delivered.
+    /// Teardown is bounded instead of depending on the peer to make progress.
+    fn poll_teardown_deadline(&mut self, cx: &mut Context<'_>) -> Poll<()> {
+        let deadline = self
+            .teardown_deadline
+            .get_or_insert_with(|| Box::pin(tokio::time::sleep(TEARDOWN_TIMEOUT)));
+        deadline.as_mut().poll(cx)
```
dmitry-markin, line 511: Shouldn't we set `self.teardown_deadline = None` on `poll()` yielding `Poll::Ready` here?

dmitry-markin (approved):

src/transport/webrtc/substream.rs
gab8i: The expiry of the deadline always transitions the channel to Reset, clearing it would re-arm using `get_or_insert_with`

gab8i: ok, I've misread the comment, you are not suggesting to always set the deadline to None at the end of the function call but to do so if the deadline has expired. This isn't isn't necessary because if the deadline is expired the Reset state would be set alongside with the `ResetStream` flag being sent, this imply that no more polling will happen on that substream

PR #661, opened 2026-09-04
webrtc: mitigate panic in `str0m` when handling oversized datagrams

dmitry-markin:
A temporary mitigation until the issue is fixed upstream.

### DIFF
--- a/src/transport/webrtc/mod.rs
+++ b/src/transport/webrtc/mod.rs
@@ -87,6 +87,9 @@ const LOG_TARGET: &str = "litep2p::webrtc";
 const REMOTE_FINGERPRINT: &str =
     "sha-256 FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF";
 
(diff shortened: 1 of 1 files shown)


src/transport/webrtc/mod.rs
```diff
@@ -87,6 +87,9 @@ const LOG_TARGET: &str = "litep2p::webrtc";
 const REMOTE_FINGERPRINT: &str =
     "sha-256 FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF:FF";
 
+/// A maximum single UDP payload size that we forward to `str0m`.
+const MAX_UDP_PAYLOAD_SIZE: usize = 2000;
```
lexnv, line 91: Is this the exact constant after which str0m panics, or is picked to be smaller? Can it be resized to a power of 2?

lexnv (approved):

src/transport/webrtc/mod.rs
dmitry-markin: Picked to be way smaller. There is no practical benefit of having it to be a power of 2.

gab8i (approved):

PR #663, opened 2026-09-11
Ensure we report connection closed to the upper layers

bkchr:
Right now it can happen that we do not report a closed connection to the upper layers. Making it look like the connection is still open and is not never removed, rendering the peer unconnectabble.

This pull request solves this by preventing any kind of errors being propagated from the inner connection calls. All errors are getting handled, close the connection and one unified path reports the connection close.


dmitry-markin (approved):
One minor nit, feel free to adopt or leave as a separate PR into master: https://github.com/paritytech/litep2p/pull/664

PR #664, opened 2026-09-11
Report `SubstreamOpenFailure` when switching to secondary connection

dmitry-markin:
This ensures in-flight requests that can't succeed anymore are terminated.


bkchr (commented):
Do we actually need this? Because they should be informed about the closed connection and then, drop all substreams?

dmitry-markin:
> Do we actually need this? Because they should be informed about the closed connection and then, drop all substreams?

They are not while the secondary connection exists.

PR #666, opened 2026-09-18
kad: Fix local provider eviction and removal panics

### LINKED ISSUE #665: kad: MemoryStore can evict its own provider record, then panics when removing it
## What happens

When a key already has `max_providers_per_key` providers, `put_provider` drops the farthest one.
That can be our own record, and `local_providers` is not updated:

```rust
// store.rs:255
if providers.len() == self.config.max_providers_per_key {
    providers.pop();                 // this can be our own record
}
providers.insert(i, provider_record.clone());
true
```

So the two maps disagree: we are gone from `provider_keys` but still listed in
`local_providers`.

Later `remove_local_provider` looks for us in `provider_keys`, does not find us, and asserts:

```rust
// store.rs:320
Err(_) => {
    tracing::error!(?key, "local provider not found during removal",);
    debug_assert!(false);
    return;
}
```

`get_providers` can cause the same mismatch when it drops expired records (store.rs:164).

## Why it matters

With debug assertions on, the Kademlia task dies. The process keeps running, so the node stays
up but loses networking and RPC. From outside it just looks like a refused connection.

With assertions off there is no panic, but the node still thinks it provides a key it was
evicted from. Bootnode discovery silently stops working for those nodes, and nothing is logged.

## How to reproduce

Run a parachain with more than 20 nodes. They all advertise as bootnodes under one per-epoch key
on the relay DHT, so the provider list fills up and starts evicting. Each epoch rollover
withdraws the advertisement and calls `remove_local_provider`.

Below 21 providers on a key this cannot happen, which is why it only showed up on a 100-node
network. We saw it on 4 of 4 runs, 1 to 8 nodes down per run, and the count grows with every
rollover (8 of 100 after 58 epochs).

Full log from one of the nodes:

```
2026-09-17 15:11:42.060  INFO tokio-runtime-worker babe: [Relaychain] 👶 New epoch 4 launching at block 0xfdd2…73ca (block slot 298276317 >= start slot 298276317).
2026-09-17 15:11:42.114 ERROR tokio-runtime-worker litep2p::protocol::libp2p::kademlia::store: [Relaychain] local provider not found during removal key=Key(b"\xb1\x0f\xd1\xc4Z\xa1\xc8x\xafcK\xda\xb0E\x7f\xd6_\xb5;Wt\xf5\xc0in\xd2\x11\xa6wO\x85\x14\x92\x86")

====================

Version: 1.24.2-dccf478651c

   0: sp_panic_handler::set::{{closure}}
   1: std::panicking::panic_with_hook
   2: std::panicking::panic_handler::{{closure}}
   3: std::sys::backtrace::__rust_end_short_backtrace
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::panicking::panic
   7: litep2p::protocol::libp2p::kademlia::store::MemoryStore::remove_local_provider
   8: litep2p::protocol::libp2p::kademlia::Kademlia::run::{{closure}}
   9: litep2p::Litep2p::new::{{closure}}
  10: <sc_service::task_manager::prometheus_future::PrometheusFuture<T> as core::future::future::Future>::poll
  11: <futures_util::future::select::Select<A,B> as core::future::future::Future>::poll
  12: <tracing_futures::Instrumented<T> as core::future::future::Future>::poll
  13: tokio::runtime::task::core::Core<T,S>::poll
  14: tokio::runtime::task::harness::Harness<T,S>::poll
  15: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
  16: tokio::runtime::scheduler::multi_thread::worker::Context::run
  17: tokio::runtime::context::scoped::Scoped<T>::set
  18: tokio::runtime::context::runtime::enter_runtime
  19: tokio::runtime::scheduler::multi_thread::worker::run
  20: tokio::runtime::task::core::Core<T,S>::poll
  21: tokio::runtime::task::harness::Harness<T,S>::poll
  22: tokio::runtime::blocking::pool::Inner::run
  23: std::sys::backtrace::__rust_begin_short_backtrace
  24: core::ops::function::FnOnce::call_once{{vtable.shim}}
  25: std::sys::thread::unix::Thread::new::thread_start
  26: start_thread
  27: clone


Thread 'tokio-runtime-worker' panicked at 'assertion failed: false', /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/litep2p-0.15.1/src/protocol/libp2p/kademlia/store.rs:320

This is a bug. Please report it at:
```

The pod stays `Running` with 0 restarts after this.

## Environment

litep2p 0.15.1, same code in 0.15.2 and master. Polkadot `1.24.2-dccf478651c` built with
`-Cdebug-assertions=y`. `max_providers_per_key` at the default 20.

DenzelPenzel:
Fixes #665.

- Keep local providers separate from remote records so eviction or expiry cannot remove an active local registration or cause a panic on removal.
- Enforce separate local/remote provider-key limits and report `QueryFailed` without publishing rejected registrations.
- Add regression coverage for eviction, expiry, refresh at capacity, and registration rejection.


src/protocol/libp2p/kademlia/store.rs
```diff
@@ -62,7 +62,10 @@ pub struct MemoryStore {
     config: MemoryStoreConfig,
     /// Records.
     records: HashMap<Key, Record>,
-    /// Provider records.
+    /// Provider records of remote peers. Never contains the local provider: local providers
+    /// are tracked in `local_providers` and merged into `get_providers()` results on the fly,
+    /// so that they never compete with remote providers for the `max_providers_per_key`
+    /// capacity.
```
gab8i: I think this comment isn't strictly related to `provider_keys`, it should be either removed or simplified because `get_providers` already mention this.

src/protocol/libp2p/kademlia/store.rs
```diff
@@ -265,68 +294,66 @@ impl MemoryStore {
...
+    /// `max_providers_per_key` eviction or expiration: we provide `key` until
+    /// [`MemoryStore::remove_local_provider`] is called. The number of keys we provide is
+    /// still bounded by `max_provider_keys`, accounted separately from the provider records
+    /// of remote peers, so that remote records can never crowd out local registrations and
+    /// vice versa.
+    ///
+    /// Returns `true` if we provide `key` from now on, `false` if the registration was
+    /// rejected because the local provider key limit was reached. Renewing an existing
+    /// registration always succeeds. On success a refresh of the provider record is
+    /// scheduled via [`MemoryStore::next_action`].
```
gab8i: A less verbose doc could be

```
/// Register ourself as a provider for `key` and schedule its refresh.
///
/// Local providers are kept apart from remote records: they are never evicted
/// or expired, only removed via [`MemoryStore::remove_local_provider`]. The number
/// of provided keys is bounded by `max_provider_keys`, counted separately from
/// remote provider keys.
///
/// Returns `false` if `key` is new and the limit is reached. Renewals always succeed.	
```

gab8i (commented):
With this change `max_provider_keys` bounds remote keys and local registrations independently, so a node can hold up to 2 × `max_provider_keys` in total. This should be documented in the docs of `MemoryStoreConfig::max_provider_keys`, `with_max_provider_keys`, and in the changelog. Alternatively, would a dedicated limit for local keys be clearer than reusing the same value?

DenzelPenzel:
> With this change `max_provider_keys` bounds remote keys and local registrations independently, so a node can hold up to 2 × `max_provider_keys` in total. This should be documented in the docs of `MemoryStoreConfig::max_provider_keys`, `with_max_provider_keys`, and in the changelog. Alternatively, would a dedicated limit for local keys be clearer than reusing the same value?

Thanks for the review! Went with documenting the current behaviour

